PR #67, opened 2014-02-20
autoconf: test for Perl and Perl's JSON module

ktdreyer:
It can be very hard to diagnose when `make-protos.pl` fails, because the build will go all the way through until it complains about missing headers (`der-protos.h` for example).

Add some checks to configure to verify that Perl and Perl's JSON module are available.

(See also the heimdal-discuss mailing list thread about this from November 2013, "building heimdal from git tar ball, missing der-protos.h")


jaltman:
298ee93ac25467c8ebfcf4d1a88caaa371755fa2 was accepted to detecting Perl modules.

PR #103, opened 2014-06-04
krb5tgs: let TGS_REQ with NULL caddr returns an addessless ticket

hautreux:
When processing a request, current tgs_make_reply uses the requested
set of addrs of the request to establish the set of addresses to
associate with the ticket in reply.

However, when the request input set of addrs is NULL, it reverts to
using the TGT set of addresses instead. As a result, it is not
possible to acquire an addressless TGS (or forwarded TGT) using a
TGT that is addressed.

This patch remove the fallback ensuring that a TGS_REQ with a set
of addrs set to NULL enables to acquire an addressless ticket.


hautreux:
Thanks !

PR #115, opened 2014-12-16
Allow LDAP modifications for entry creation

klondi:
Heimdal will refuse to create new entries when an entry already exists even
if said entry has no kerberos info and is a new entry.

This patch fixes this issue by allowing object modifications even if the
flags disallow them when we are inserting a new principal on the database.

Signed-off-by: Francisco Blas Izquierdo Riera (klondike) klondike@gentoo.org


vdukhovni:
If this patch is still needed, please rebase so we can run Travis against the latest versions of the tests, which should have fewer race conditions.

nicowilliams:
Can you rebase and investigate the `check-iprop` failure in https://travis-ci.org/heimdal/heimdal/jobs/44176699 ?

nicowilliams:
Thanks!

PR #116, opened 2014-12-18
Heimdal kdc preauth

abartlet:
While researching a very strange (at first) test failure in Samba when trying to rebase lorikeet-heimdal up to the modern world, I noticed that Heimdal no longer returned KRB5KDC_ERR_PREAUTH_FAILED errors. 

This gave the client library KRB5_GET_IN_TKT_LOOP internally when it got KRB5KDC_ERR_PREAUTH_REQUIRED twice in a row. 

This adds a potential fix, which should be carefully examined by someone who understands the intended behaviour in the presence of multiple possible PA requests on the AS-REQ.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ install:
     - make
 
 script:
-    - if [ ${COVERITY_SCAN_BRANCH} != 1 ]; then make check; fi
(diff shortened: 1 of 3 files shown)


abartlet:
Can I get a comment or thoughts on this

PR #118, opened 2015-01-15
kadm5: fix race in Makefile with kadm5_err.h

ktdreyer:
When running `make` with `-j4`, occasionally kadm5 fails due to a missing header file `kadm5_err.h`. Fix the race condition.

Thanks @jcajka for the original patch.

This is a cherry-pick of commit 6affa4cceceaa1369dd895f8acdd7a883ee65674 which was merged to master in https://github.com/heimdal/heimdal/pull/108 . This pull request is for the 1.6 branch.


lha:
Thanks!

PR #119, opened 2015-01-15
ipv6 loopbacks fix

ktdreyer:
I'm not sure if this is still needed? @net9rok had added this to the Fedora packages in July 2012, and I want to send this upstream to see if we still need it or if I should drop it from the Fedora/EPEL packages.

"Fix IPv6 and multiple interfaces bug in krb5_parse_address." See discussion at https://bugzilla.redhat.com/808147

### DIFF
--- a/lib/krb5/addr_families.c
+++ b/lib/krb5/addr_families.c
@@ -1161,6 +1161,7 @@ krb5_parse_address(krb5_context context,
 {
     int i, n;
     struct addrinfo *ai, *a;
+    struct addrinfo hint;
     int error;
     int save_errno;
 
@@ -1180,7 +1181,10 @@ krb5_parse_address(krb5_context context,
 	}
     }
 
-    error = getaddrinfo (string, NULL, NULL, &ai);
+    /* if not parsed as numeric address, do a name lookup */
+    memset(&hint, 0, sizeof(hint));
+    hint.ai_family = AF_UNSPEC;
+    error = getaddrinfo (string, NULL, &hint, &ai);
     if (error) {
 	krb5_error_code ret2;
 	save_errno = errno;


lib/krb5/addr_families.c
```diff
@@ -1180,7 +1181,10 @@ krb5_parse_address(krb5_context context,
 	}
     }
 
-    error = getaddrinfo (string, NULL, NULL, &ai);
+    // if not parsed as numeric address, do a name lookup
```
lha: please use /* */ comment

lib/krb5/addr_families.c
```diff
@@ -1180,7 +1181,10 @@ krb5_parse_address(krb5_context context,
...
+    memset(&hint, 0x00, sizeof(hint));
```
lha: please use 0 instead of 0x00 since that's what the rest of the code is doing

lha:
other the comments, I think I'm ok with the change

net9rok:
GNU glibc doesn't follow standard behaviour. Please see: http://man7.org/linux/man-pages/man3/getaddrinfo.3.html.

I quote: "Specifying hints as NULL is equivalent to setting ai_socktype and ai_protocol to 0; ai_family to AF_UNSPEC; and ai_flags to (AI_V4MAPPED | AI_ADDRCONFIG).  (POSIX specifies different defaults for ai_flags; see NOTES.)"

Notes say:
"According to POSIX.1-2001, specifying hints as NULL should cause ai_flags to be assumed as 0.  The GNU C library instead assumes a value of (AI_V4MAPPED | AI_ADDRCONFIG) for this case, since this value is considered an improvement on the specification."

The patch makes sure that krb5_parse_address works consistently on both GNU libc and systems that follow POSIX.1-2001 to the letter. Some incorrect Fedora 17 patches managed to break IPv6 connectivity and were later backed out. It is still my recommendation to apply the patch and resolve the incompatibility.

net9rok:
The exact symptoms were: krb5_parse_address didn't parse ::1 correctly and IPv6 connectivity failed. Test cases were failing on IPv6 systems.

jaltman:
I would like to see the commit message updated with these details and a new pull request submitted.

ktdreyer:
I've amended the commit and added the comments from @jaltman and @net9rok to the commit log message.

jaltman:
Thank you Ken.

PR #124, opened 2015-04-10
Fix build when OpenSSL has no EGD support

Sp1l:
Hi,

Current 1.5.3 heimdal does not build with LibreSSL as it does not have EGD support. This fixing and upstreaming is part of the work done to enable all ports to work with LibreSSL on FreeBSD, see https://wiki.freebsd.org/LibreSSL and the PR to fix this is https://bugs.freebsd.org/198527
I believe it is safe to remove EGD altogether as this is a relic from the past, nothing recent requires EGD (not even Windows as can be seen in the master branch).
Latest versions requiring EGD
IRIX 6.5.19 Feb 2003
Solaris 2.6 Jul 1997
AIX 5.2 Oct 2002
Tru64 5.1B Sep 2002
HP-UX 11i v2 Sep 2003

This patch detects EGD support in the OpenSSL (or LibreSSL) libcrypto and adds ifdef's where relevant. Looks a bit kludgy now that this has been made conditional in the reverse way for Windows.

Hope you will add this to your next release!

Thanks, Bernard Spil

### DIFF
--- a/cf/crypto.m4
+++ b/cf/crypto.m4
@@ -118,6 +118,7 @@ if test "$crypto_lib" = "unknown" -a "$with_openssl" != "no"; then
 			break;
 		fi
 	done
+	AC_CHECK_LIB(crypto, RAND_egd, AC_DEFINE(HAVE_RAND_EGD, 1, [Define if the libcrypto has RAND_egd]))
 	CFLAGS="$save_CFLAGS"
 	LIBS="$save_LIBS"
 fi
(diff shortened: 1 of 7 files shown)


jaltman:
Windows is not going to define HAVE_EGD.  The reason it is a negative is the due to the lack of autoconf testing.   For an autoconf based test the define should be positive.

Sp1l:
Hi Jeffrey,

Sorry for the confusion... It's just that I think that both a positive as well as a negative test for the same feature isn't verry pretty...

Any comments on the rest of the proposed change?

sevan:
@jaltman it be good to clarify any reservations you have with this change so that if it's possible then they can be addressed.

sevan:
@jaltman Are the proposed changes ok the subsequent time?

lha:
Thanks!

PR #128, opened 2015-04-30
Fix compilation error in dlfcn.h

kiransj:
When dlfcn.h is included from a C++ file causes a compilation error due to missing '{'.


lha:
thx

PR #134, opened 2015-07-07
gssapi: Allow a NULL authenticator

abartlet:
Some non-GSSAPI implementations that instead try to create compatible packets by wrapping krb5_mk_req()
can trigger a NULL authenticator here.  Assume this to be equvilent to specifying an all-zero
channel bindings and some reasonable (fixed) flags.

Andrew Bartlett

### DIFF
--- a/lib/gssapi/krb5/accept_sec_context.c
+++ b/lib/gssapi/krb5/accept_sec_context.c
@@ -511,13 +511,8 @@ gsskrb5_acceptor_start(OM_uint32 * minor_status,
 	    return ret;
 	}
 
-	if (authenticator->cksum == NULL) {
-	    krb5_free_authenticator(context, &authenticator);
-	    *minor_status = 0;
-	    return GSS_S_BAD_BINDINGS;
-	}
-
-        if (authenticator->cksum->cksumtype == CKSUMTYPE_GSSAPI) {
+        if (authenticator->cksum != NULL
+	    && authenticator->cksum->cksumtype == CKSUMTYPE_GSSAPI) {
             ret = _gsskrb5_verify_8003_checksum(minor_status,
 						input_chan_bindings,
 						authenticator->cksum,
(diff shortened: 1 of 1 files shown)


abartlet:
The primary issue with this patch is that like the support for 'Samba style' checksums, it allows clients to be in total disregard to the spec, attempting to emulate GSSAPI by raw krb5 routines and ASN.1 wrapping. 

While not a security issue (this is all inside an integrity protected packet, signed and encrypted by the client) It has been rightly put that this should be made optional for the server to accept.  The flip side of the argument is that if we do that, we should make the same option control 'samba-style' use, and doing so would break existing Samba installations until they are updated to request this, via presumably a flag on the gss_creds.

abartlet:
As context, the samba patches are here:
git://git.samba.org/abartlet/samba.git allow-no-krb5-checksum
https://git.samba.org/?p=abartlet/samba.git/.git;a=shortlog;h=refs/heads/allow-no-krb5-checksum

In a private forum, Love expressed a desire to have this protected by a flag, and I've done some work towards that, but I would like to avoid that if possible, to make this more backward compatible.

abartlet:
The official answer from MS on why this is allowed is this:

On Thu, 2015-07-30 at 18:37 +0000, Sreekanth Nadendla wrote:
> Hello Andrew,
> I've verified this and we are adding the following text In MS-KILE 
> section "3.4.5 Message Processing Events and Sequencing Rules", to 
> explain the deviation you have reported.
> 
> When the checksum field is not present, the application server SHOULD 
> process the requests as though none of the flags (RFC 4121 section
> 4.1.1.1) are set and SHOULD NOT check channel binding information (RFC 
> 4121 section 4.1.2.1).

essentially this is updating the spec to match the code, but anyway.  The only question is should we be Microsoft-compatible always, or just when a flag is set?

jaltman:
I see no reason to not be compatible always.   Love?

lha:
thanks Andrew

abartlet:
BTW, the NAS in question with the non-GSSAPI GSSAPI implementation is a Huawei Unified Storage System S5500 V3, using Samba as an AD DC.  Sorry for not making that clear in the commit message.

PR #136, opened 2015-07-20
PK-INIT improvements

HenryJacques:
This PR add:

* more support related to common PIN errors (PIN blocked, incorrect, ...)
* the corresponding messages allowing a more friendly debugging.
* fix a small typo
* allow more than one token to be present (if the first is not usable, the second one can be).

Credit: thanks to Romain Sahut for his help.


lha:
thanks

PR #140, opened 2015-09-24
Add ability to specifiy PKCS#11 slot number when using hx509

madscientist159:
Example usage: kinit -C PKCS11:/usr/lib/opensc-pkcs11.so,slot=3 foo@BAR.TLD

This is required for certain tokens, such as the OpenPGP cards, that default to an incorrect slot.  Slot numbering is consistent with the pkcs11 PAM module.


lha:
little fragile since I'm not sure that developers that reads this next time expects the p->slot[...] that is not select to be zero filled

would be better to use an heim_array and grow as needed when slots are needed

PR #141, opened 2015-09-26
Add ability to store extended principal attributes in LDAP

madscientist159:
When attempting to deploy PKI infrastructure on our Kerbeos servers with LDAP backends it was discovered that the Heimdal Kerbeos extended attribute structure (including PKI mapping fields) was not supported with the LDAP backend.  This patch fixes this critical issue.

Original commit text:
A careful code review was undertaken, and it was determined
that the best way to store the extended attributes was in a
native ASN1 encoded field.  LDAP does not understand the
SEQUENCE of SEQUENCE structures used extensively throughout
the extended attributes structure, and there was already a
precedent set for storing the krb5Key data in a native ASN1
encoded field.

### DIFF
--- a/lib/hdb/hdb-ldap.c
+++ b/lib/hdb/hdb-ldap.c
@@ -2,6 +2,7 @@
  * Copyright (c) 1999-2001, 2003, PADL Software Pty Ltd.
  * Copyright (c) 2004, Andrew Bartlett.
  * Copyright (c) 2003 - 2008, Kungliga Tekniska Högskolan.
+ * Copyright (c) 2015, Timothy Pearson.
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
@@ -91,6 +92,7 @@ static char * krb5kdcentry_attrs[] = {
     "krb5PasswordEnd",
     "krb5PrincipalName",
     "krb5PrincipalRealm",
+    "krb5ExtendedAttributes",
     "krb5ValidEnd",
     "krb5ValidStart",
     "modifiersName",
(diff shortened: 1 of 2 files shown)


lib/hdb/hdb-ldap.c
```diff
@@ -1048,6 +1078,27 @@ LDAP_message2entry(krb5_context context, HDB * db, LDAPMessage * msg,
 #endif
     }
 
+    extensions = ldap_get_values_len(HDB2LDAP(db), msg, "krb5ExtendedAttributes");
+    if (extensions != NULL) {
+	size_t l;
+
+	ent->entry.extensions = malloc(sizeof(*(ent->entry.extensions)));
```
lha: check return value, use ret = krb5_enomem(context);\ngoto out;

lib/hdb/hdb-ldap.c
```diff
@@ -1048,6 +1078,27 @@ LDAP_message2entry(krb5_context context, HDB * db, LDAPMessage * msg,
...
+	ent->entry.extensions->len = ldap_count_values_len(extensions);
+	ent->entry.extensions->val = (HDB_extension *) calloc(ent->entry.extensions->len, sizeof(HDB_extension));
+	if (ent->entry.extensions->val == NULL) {
+	    ret = ENOMEM;
+	    krb5_set_error_message(context, ret, "calloc: out of memory");
```
lha: use krb5_enomem()

lib/hdb/hdb-ldap.c
```diff
@@ -1048,6 +1078,27 @@ LDAP_message2entry(krb5_context context, HDB * db, LDAPMessage * msg,
...
+	    goto out;
+	}
+	for (i = 0; i < ent->entry.extensions->len; i++) {
+	    decode_HDB_extension((unsigned char *) extensions[i]->bv_val,
```
lha: check return value, krb5_set_error_message with the same error code so we get a sane error string

lha:
fix review comments and I'm happy to merge, thanks for fixing this.

preferably the extension should be stored as native ldap options, but this will do for now

madscientist159:
Should be fixed now.  I don't see how we can reasonably break up certain attributes, such as the PKI alias fields, due to their SEQUENCE of SEQUENCE construction.  LDAP only "understands" either a single value per attribute or an unordered list of attributes; trying to add unordered lists of PKI alias fields would require adding a completely new (non-derivative) data type to LDAP, which is unfortunately beyond my capabilities at this time.

lib/hdb/hdb-ldap.c
```diff
@@ -1048,6 +1078,35 @@ LDAP_message2entry(krb5_context context, HDB * db, LDAPMessage * msg,
...
+	ent->entry.extensions->val = (HDB_extension *) calloc(ent->entry.extensions->len, sizeof(HDB_extension));
+	if (ent->entry.extensions->val == NULL) {
+	    ent->entry.extensions->len = 0;
+	    ret = krb5_enomem(context);
+	    goto out;
+	}
+	for (i = 0; i < ent->entry.extensions->len; i++) {
+	    int asn1_decode_ret = decode_HDB_extension((unsigned char *) extensions[i]->bv_val,
+		       (size_t) extensions[i]->bv_len, &ent->entry.extensions->val[i], &l);
+	    if (asn1_decode_ret < 0) {
```
lha: the correct check here is actually != 0 or just if (asn1_decode_ret) {

I think you should just use ret directly

madscientist159: I was going on the documentation here:
http://bellard.org/ffasn1/ffasn1c.html#TOC10

Specifically, "Return the number of consumed bytes or < 0 if error.".  Does Heimdal's ASN1 parser behave differently?

lha: Yes, 0 is sucess, all others return codes are errors.

madscientist159:
OK, updated.

PR #142, opened 2015-10-06
Completely remove RAND_egd support

hasufell:
The EGD daemon is completely unmaintained and has not seen a release
since 13 years which is not an acceptable timeframe for cryptographic
software. It is not packaged in any linux distribution I know of
and definitely not in *BSD.

LibreSSL has already dropped support for RAND_egd

Also see http://sourceforge.net/projects/egd/files/egd/


lha:
thanks

hasufell:
thanks for the quick response

PR #144, opened 2015-10-26
roken: fix rename recursion due to macro

meffie:
rk_rename() function must not be redefined to rename() in rename.c, otherwise
the function becomes an infinite recursion.

### DIFF
--- a/lib/roken/rename.c
+++ b/lib/roken/rename.c
@@ -31,6 +31,7 @@
 
 #include <config.h>
 #include "roken.h"
+#undef rk_rename
 
 /* rename() for platforms where the native implementation doesn't
  * unlink newname. */


SimonWilkinson:
For this to be the case, doesn't there need to be a 
    #define rename rk_rename
somewhere? I might be missing something, but I can't see such a define in current Heimdal master.

meffie:
Hi Simon,

In roken.h.in in heimdal we have;

 694 #ifdef RENAME_DOES_NOT_UNLINK
 695 ROKEN_LIB_FUNCTION int ROKEN_LIB_CALL rk_rename(const char *, const char *);
 696 #else
 697 #define rk_rename(__rk_rn_from,__rk_rn_to) rename(__rk_rn_from,__rk_rn_to)
 698 #endif

When RENAME_DOES_NOT_UNLINK is not defined (which is normally on unix-like systems), then
roken.h has:

#define rk_rename(__rk_rn_from,__rk_rn_to) rename(__rk_rn_from,__rk_rn_to) 

In rename.c, we have:

#include "roken.h"  ( which has #define rk_rename ... rename)
int rk_rename(...)
{
     rename(..) 

After the preprocessor runs we have:

int rename(...) {
      rename(...);

Checking the object code, you can see a rename() function which calls itself.

In openafs, under linux and solaris, rk_rename() is defined as rename(). Under linux we happen to link against the libc rename, but under solaris we call the libroken rename, which recurses until we run over the stack.

jaltman:
OpenAFS should not build roken's rename.c on linux and solaris.  Creating a function with the name rename() will only confuse the linker and mask the system rename()..

The reason the macro exists is that Heimdal calls rk_roken() internally and we want it to use the system rename() function when RENAME_DOES_NOT_UNLINK is undefined.

nicowilliams:
@meffie Is right.  Normally we rename `foo` to `rk_foo` when we need to wrap/replace/provide-an-actual `foo` with one from libroken.  With `rename()` we have the pattern backwards.  Ideally we could just use `#ifdef RENAME_DOES_NOT_UNLINK` in `lib/roken/rename.c`, but empty source files are technically not allowed by C99, so `#undef`ing the macro as in @meffie's fix is correct.

Note that `lib/roken/rename.c` only gets built on Windows, so @jaltman 's concern is taken care of.

The only reason we need an `rk_rename()` macro at all is that `lib/krb5/fcache.c` calls `rk_rename()`.  I don't know why that is, but it seems to be part of having the roken pattern backwards for rename...

To use the normal pattern, nothing should call `rk_rename()` directly, and instead libroken should define a `rename()` macro when `RENAME_DOES_NOT_UNLINK` that expands to a call to `rk_rename()`.  So @meffie 's fix is not complete, but close.

Or did I miss something?

jaltman:
@nicowilliams rk_rename() is used internally to Heimdal and renamed to rename() on platforms that provide a working rename().  We cannot go in the other direction for symbols that exist but have the wrong behavior.  For example, the rename() and close() functions on Windows.  The rule is that lib/roken/rename.c is simply not compiled or linked on platforms that aren't broken.  The only broken platform is Windows.

PR #145, opened 2015-11-28
Lukeh/pkcs11 hcrypto

nicowilliams:
I created a PR for this to make it easier to record my comments and capture the head of lukeh/pkcs11-hcrypto as it was when I reviewed.

### DIFF
--- a/cf/crypto.m4
+++ b/cf/crypto.m4
@@ -135,6 +135,17 @@ if test "$crypto_lib" = "unknown"; then
 
 fi
 
+AC_ARG_WITH(pkcs11-module,
+                       AS_HELP_STRING([--with-pkcs11-module=path],
+                                      [use PKCS11 module in path]),
+                       [pkcs11_module="$withval"],
+                       [])
+
+if test "$pkcs11_module" != ""; then
+  AC_DEFINE_UNQUOTED(PKCS11_MODULE_PATH, "$pkcs11_module", [path to PKCS11 module])
+  openssl=no
+fi
+
 if test "$openssl" = "yes"; then
   AC_DEFINE([HAVE_OPENSSL], 1, [define to use openssl's libcrypto])
 fi
--- a/lib/hcrypto/Makefile.am
+++ b/lib/hcrypto/Makefile.am
@@ -2,7 +2,8 @@
 
 include $(top_srcdir)/Makefile.am.common
 
-AM_CPPFLAGS += -I$(srcdir)/libtommath -DUSE_HCRYPTO_LTM=1
+AM_CPPFLAGS += -I$(top_srcdir)/lib/hx509 \
+        -I$(srcdir)/libtommath -DUSE_HCRYPTO_LTM=1
 
 lib_LTLIBRARIES = libhcrypto.la
 check_LTLIBRARIES = libhctest.la
@@ -11,6 +12,7 @@ libhcrypto_la_LDFLAGS = -version-info 5:0:1
 libhcrypto_la_LIBADD = \
 	$(top_builddir)/lib/asn1/libasn1.la \
 	$(LIB_dlopen) \
+	$(LIB_heimbase) \
 	$(LIBADD_roken)
 
 hcryptoincludedir = $(includedir)/hcrypto
@@ -29,6 +31,7 @@ hcryptoinclude_HEADERS =	\
 	evp.h			\
 	evp-hcrypto.h		\
 	evp-cc.h		\
+	evp-pkcs11.h		\
 	hmac.h			\
 	md2.h			\
 	md4.h			\
@@ -114,6 +117,7 @@ libhcrypto_la_SOURCES =	\
 	evp.h		\
 	evp-hcrypto.c	\
 	evp-cc.c	\
+	evp-pkcs11.c	\
 	engine.c	\
 	engine.h	\
 	hash.h		\
--- a/lib/hcrypto/evp-pkcs11.c
+++ b/lib/hcrypto/evp-pkcs11.c
@@ -1,0 +1,775 @@
+/*
+ * Copyright (c) 2015, Secure Endpoints Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ * - Redistributions of source code must retain the above copyright
+ *   notice, this list of conditions and the following disclaimer.
+ *
+ * - Redistributions in binary form must reproduce the above copyright
+ *   notice, this list of conditions and the following disclaimer in
+ *   the documentation and/or other materials provided with the
+ *   distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ * FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ * COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT,
+ * INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+ * STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+ * OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/* PKCS#11 provider */
+
+#include "config.h"
+
+#include <assert.h>
+#ifdef HAVE_DLFCN_H
+#include <dlfcn.h>
+#ifndef RTLD_LAZY
+#define RTLD_LAZY 0
+#endif
+#ifndef RTLD_LOCAL
+#define RTLD_LOCAL 0
+#endif
+#ifndef RTLD_GROUP
+#define RTLD_GROUP 0
+#endif
+#ifndef RTLD_NODELETE
+#define RTLD_NODELETE 0
+#endif
+#else
+#error PKCS11 support requires dlfcn.h
+#endif
+
+#include <roken.h>
+#include <heimbase.h>
+
+#include <evp.h>
+#include <evp-hcrypto.h>
+#include <evp-pkcs11.h>
+
+#include <ref/pkcs11.h>
+
+#if __sun && !defined(PKCS11_MODULE_PATH)
+# if _LP64
+# define PKCS11_MODULE_PATH "/usr/lib/64/libpkcs11.so"
+# else
+# define PKCS11_MODULE_PATH "/usr/lib/libpkcs11.so"
+# endif
+#endif
+
+static CK_FUNCTION_LIST_PTR p11_module;
+
+static int
+p11_cleanup(EVP_CIPHER_CTX *ctx);
+
+struct pkcs11_cipher_ctx {
+    CK_SESSION_HANDLE hSession;
+    CK_OBJECT_HANDLE hSecret;
+};
+
+struct pkcs11_md_ctx {
+    CK_SESSION_HANDLE hSession;
+};
+
+static void
+p11_module_init_once(void *context)
+{
+    CK_RV rv;
+    CK_FUNCTION_LIST_PTR module;
+    CK_RV (*C_GetFunctionList_fn)(CK_FUNCTION_LIST_PTR_PTR);
+    void *handle = NULL;
+
+    if (!issuid()) {
+        char *pkcs11ModulePath = getenv("PKCS11_MODULE_PATH");
+        if (pkcs11ModulePath != NULL) {
+            handle = dlopen(pkcs11ModulePath, RTLD_LAZY | RTLD_LOCAL |
+                                              RTLD_GROUP | RTLD_NODELETE);
+            if (handle == NULL)
+                fprintf(stderr, "p11_module_init(%s): %s\n", pkcs11ModulePath, dlerror());
+        }
+    }
+#ifdef PKCS11_MODULE_PATH
+    if (handle == NULL) {
+        handle = dlopen(PKCS11_MODULE_PATH, RTLD_LAZY | RTLD_LOCAL |
+                                            RTLD_GROUP | RTLD_NODELETE);
+        if (handle == NULL)
+            fprintf(stderr, "p11_module_init(%s): %s\n", PKCS11_MODULE_PATH, dlerror());
+    }
+#endif
+    if (handle == NULL) {
+        rv = CKR_LIBRARY_LOAD_FAILED;
+        goto cleanup;
+    }
+
+    C_GetFunctionList_fn = (CK_RV (*)(CK_FUNCTION_LIST_PTR_PTR))
+                           dlsym(handle, "C_GetFunctionList");
+    if (C_GetFunctionList_fn == NULL) {
+        rv = CKR_LIBRARY_LOAD_FAILED;
+        goto cleanup;
+    }
+
+    rv = C_GetFunctionList_fn(&module);
+    if (rv != CKR_OK)
+        goto cleanup;
+
+    rv = module->C_Initialize(NULL);
+    if (rv == CKR_CRYPTOKI_ALREADY_INITIALIZED)
+        rv = CKR_OK;
+    if (rv == CKR_OK)
+        *((CK_FUNCTION_LIST_PTR_PTR)context) = module;
+
+cleanup:
+    if (handle != NULL && p11_module == NULL)
+        dlclose(handle);
+    /* else leak handle */
+}
+
+static CK_RV
+p11_module_init(void)
+{
+    static heim_base_once_t init_module = HEIM_BASE_ONCE_INIT;
+
+    heim_base_once_f(&init_module, &p11_module, p11_module_init_once);
+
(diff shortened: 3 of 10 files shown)


cf/crypto.m4
```diff
@@ -135,6 +135,17 @@ if test "$crypto_lib" = "unknown"; then
 
 fi
 
+AC_ARG_WITH(pkcs11-module,
```
nicowilliams: Why shouldn't this just be a configuration parameter in krb5.conf?  Why bother with autoconf'ing this?  My guess: to provide a platform-specific/sane default?

configure.ac
```diff
@@ -6,6 +6,7 @@ AC_INIT([Heimdal],[1.6.99],[heimdal-bugs@h5l.org])
 AC_CONFIG_SRCDIR([kuser/kinit.c])
 AC_CONFIG_HEADERS(include/config.h)
 AC_CONFIG_MACRO_DIR([cf])
+AC_USE_SYSTEM_EXTENSIONS
```
nicowilliams: I'm not sure this is a good idea.  We should know where and why we need extensions rather than enable them globally, otherwise we risk breaking builds more often.  Mind you, we can always fix them after the fact.  But I'd rather that developers trip over the need to enable extensions and then do whatever the right thing is, such as adding compatibility wrappers in libroken.

kdc/connect.c
```diff
@@ -965,7 +965,7 @@ loop(krb5_context context, krb5_kdc_configuration *config,
 
 #ifdef __APPLE__
 static void
-bonjour_kid(krb5_context, krb5_kdc_configuration *config, int islive)
+bonjour_kid(krb5_context context, krb5_kdc_configuration *config, int islive)
```
nicowilliams: If possible make this a separate commit.  If not I'll do it when ready to integrate.

lib/hcrypto/Makefile.am
```diff
@@ -2,7 +2,8 @@
 
 include $(top_srcdir)/Makefile.am.common
 
-AM_CPPFLAGS += -I$(srcdir)/libtommath -DUSE_HCRYPTO_LTM=1
+AM_CPPFLAGS += -I$(top_srcdir)/lib/base -I$(top_srcdir)/lib/hx509/ref \
```
nicowilliams: Maybe `-I$(top_srcdir)/lib/hx509`, without the `/ref`, to make sure that `#include <ref/pkcs11.h>` means that one, compared with `#include <pkcs11.h>` perhaps getting a system one?

lib/hcrypto/evp-pkcs11.c
```diff
@@ -0,0 +1,744 @@
...
+struct pkcs11_md_ctx {
+    CK_SESSION_HANDLE hSession;
+};
+
+static CK_RV
+p11_module_init(void)
+{
+    CK_RV rv = CKR_OK;
+
+    if (p11_module == NULL) {
```
nicowilliams: Would it be better to use `heim_base_once`?  Normally that would be the preferred way to do atomic one-time global initialization in Heimdal.

lib/hcrypto/evp-pkcs11.c
```diff
@@ -0,0 +1,744 @@
...
+
+        if (!issuid()) {
+            char *pkcs11ModulePath = getenv("PKCS11_MODULE_PATH");
+            if (pkcs11ModulePath != NULL) {
+                handle = dlopen(pkcs11ModulePath, RTLD_LAZY | RTLD_LOCAL | RTLD_GROUP);
+                if (handle == NULL)
+                    fprintf(stderr, "p11_module_init(%s): %s\n", pkcs11ModulePath, dlerror());
+            }
+        }
+#ifdef PKCS11_MODULE_PATH
```
nicowilliams: Getting this from krb5.conf if not available in the `environ` would be nice.

Also, the "there's two paths, one for the system 32-bit PKCS11 provider, and the other for the system 64-bit PKCS11 provider" problem is likely to come up elsewhere.  Maybe we need a lib/base utility for constructing dlopen()/LoadLibraryEx()/whatever paths that supports a template where whatever appropriate token for the bit-ness of the running process can be inserted.

On more recent Solaris the are `/32/` and `/64/` symlinks in all the lib directories.  On older ones it's necessary to use `/./` (or empty string) for 32-bit, and `/<isa-name>/` for 64-bit.  I.e., `/usr/lib/libfoo.so` and `/usr/lib/amd64/libfoo.so` on older systems, and `/usr/lib/32/libfoo.so` and `/usr/lib/64/libfoo.so` on newer.

lib/hcrypto/evp-pkcs11.c
```diff
@@ -0,0 +1,744 @@
...
+
+    if (p11_module == NULL) {
+        CK_RV (*C_GetFunctionList_fn)(CK_FUNCTION_LIST_PTR_PTR);
+        CK_FUNCTION_LIST_PTR module;
+        void *handle = NULL;
+
+        if (!issuid()) {
+            char *pkcs11ModulePath = getenv("PKCS11_MODULE_PATH");
+            if (pkcs11ModulePath != NULL) {
+                handle = dlopen(pkcs11ModulePath, RTLD_LAZY | RTLD_LOCAL | RTLD_GROUP);
```
nicowilliams: If we're gonna leak the handle we might as well use `RTLD_NODELETE`.

Make sure to

    #ifndef RTLD_NODELETE
    #define RTLD_NODELETE 0
    #endif

lib/hcrypto/evp-pkcs11.c
```diff
@@ -0,0 +1,744 @@
...
+                if (handle == NULL)
+                    fprintf(stderr, "p11_module_init(%s): %s\n", pkcs11ModulePath, dlerror());
+            }
+        }
+#ifdef PKCS11_MODULE_PATH
+        if (handle == NULL) {
+            handle = dlopen(PKCS11_MODULE_PATH, RTLD_LAZY | RTLD_LOCAL | RTLD_GROUP);
```
nicowilliams: Ditto.

nicowilliams: And what, no PKCS#11 support on Windows?  :)  (It's OK, just teasing.  Though it'd be nice eventually, but I'm OK with letting whoever needs it first add that support.)

lib/hcrypto/evp-pkcs11.c
```diff
@@ -0,0 +1,744 @@
...
+            if (handle == NULL)
+                fprintf(stderr, "p11_module_init(%s): %s\n", PKCS11_MODULE_PATH, dlerror());
+        }
+#endif
+        if (handle == NULL) {
+            rv = CKR_LIBRARY_LOAD_FAILED;
+            goto cleanup;
+        }
+
+        C_GetFunctionList_fn = (CK_RV (*)(CK_FUNCTION_LIST_PTR_PTR))
```
nicowilliams: There's no need to have that cast (`dlsym()` returns `void *`).

lib/hcrypto/evp-pkcs11.c
```diff
@@ -0,0 +1,744 @@
...
+    if (pSlotList == NULL) {
+        rv = CKR_HOST_MEMORY;
+        goto cleanup;
+    }
+
+    rv = p11_module->C_GetSlotList(CK_FALSE, pSlotList, &ulSlotCount);
+    if (rv != CKR_OK)
+        goto cleanup;
+
+    for (i = 0; i < ulSlotCount; i++) {
```
nicowilliams: I realize that this works on Solaris, and probably elsewhere by accident.  If we're going to allow specifying a provider by env var and/or krb5.conf, we might as well also allow a slot to be specified.  But don't implement this.  Let's leave a cookie crumb explaining that this is a bit of a hack and that it's not clear what's the best thing to do.

There's an RFC for PKCS#11 URIs, `RFC7512`, that ideally we could use for all of this, perhaps with even a URI per-mechanism.

lib/hcrypto/evp-pkcs11.c
```diff
@@ -0,0 +1,744 @@
...
+        static const EVP_CIPHER *__cipher;                              \
+                                                                        \
+        if (__cipher == NULL) {                                         \
+            const EVP_CIPHER *cipher;                                   \
+                                                                        \
+            cipher = hc_EVP_pkcs11_ ##name();                           \
+            if (cipher == NULL)                                         \
+                cipher = hc_EVP_hcrypto_ ##name();                      \
+                                                                        \
+            (void)heim_base_exchange_pointer(&__cipher, cipher);        \
```
nicowilliams: This is a lot easier than using `heim_base_once` though...  It's racy but idempotent.

lib/hcrypto/libhcrypto-exports.def
```diff
@@ -175,6 +175,23 @@ EXPORTS
 ;!	hc_EVP_cc_aes_192_cfb8
 ;!	hc_EVP_cc_aes_256_cfb8
 
+;!	hc_EVP_pkcs11_md2
```
nicowilliams: Why the comment?

kdc/connect.c
lhoward: It was but I foolishly squashed it. I'll see if I can resurrect it then commit that directly to master.

configure.ac
lhoward: OK. I think Solaris 11 build failed without this but I'll try again (does it work for you?).

cf/crypto.m4
lhoward: Yes, but also hcrypto does not link against krb5 (circular dependency).

lib/hcrypto/evp-pkcs11.c
lhoward: Isn't there a dlopen wrapper in roken for Windows? Besides, WinCNG.

lib/hcrypto/evp-pkcs11.c
lhoward: Necessary to avoid warning with Solaris Studio.

lib/hcrypto/evp-pkcs11.c
lhoward: I hadn't seen heim_base_once, I'll take a look.

nicowilliams:
The updates to `pkcs11.h` should probably be in a separate commit.

lib/hcrypto/libhcrypto-exports.def
lhoward: True, we can export this on Windows. Will fix.

lib/hcrypto/evp-pkcs11.c
lhoward: How do we read krb5.conf from hcrypto?

nicowilliams:
The `lib/libedit` changes should also be in a separate commit.

My git-fu is strong, so if you prefer to leave that to me I can do it for you at integration time.

lhoward:
Please.

lib/hcrypto/evp-pkcs11.c
nicowilliams: Yeah, there's a dlopen wrapper in lib/roken.  I'd completely forgotten :(

kdc/connect.c
nicowilliams: All the separate commit requests can be left to me.

lib/hcrypto/evp-pkcs11.c
lhoward: I guess we could export a "set pkcs#11 path" API from hcrypto but - as I understand it hcrypto should be a subset of libcrypto (although maybe this isn't a hard and fast rule) and also it would require some care not to be racy (unless we only let it be set once).

I don't have the code in front of me but is a krb5 configuration associated with a krb5_context? If so note we don't have a context associated with hcrypto (well, it might be possible but it'd be an intrusive change and again would deviate from the libcrypto API).

Basically, doing this properly may be complicated.

configure.ac
nicowilliams: Thanks for noting that the S11 build requires `AC_USE_SYSTEM_EXTENSIONS`.  I think what I'd like to do is build it myself (it's been a while) and see which extensions, where, and do the right thing on a case-by-case basis.  Or alternatively use `AC_USE_SYSTEM_EXTENSIONS` only on Solaris hosts.

cf/crypto.m4
nicowilliams: Ahh!  Got it.  Thanks.  Also, there's no place in sight where we can store any configuration.

nicowilliams:
LGTM.  I do need to go over the PKCS#11 calling code carefully, but what I looked at carefully looked good to me, and it was easy to read (thanks for that!  as always, your code is awesome).

cf/crypto.m4
lhoward: See other mail, it could be done but doing it properly would require a lot of thought and probably some API changes that would make hcrypto a superset (well, intersecting set) of libcrypto.

configure.ac
lhoward: If you have a chance to do that great, else I'll pop it behind a Solaris specific check (just for S11 yes?)

lhoward:
Well I do love writing it! I just wish we could get rid of the macro mess in the PKCS#11 and WinCNG backends by multiplexing all hcrypto APIs through EVP_get_{cipher,digest}bynid for example. Or replace hcrypto with something eke entirely...

lib/hcrypto/evp-pkcs11.c
lhoward: Well, you will get the first slot that works even if there's no meta-slot provider but I agree it may not get you the fastest or more desired provider.

lhoward:
BTW I wonder if this will deliver much performance enhancement - the OpenSolaris Developer Conference paper suggested that OpenSSL was actually faster than the PKCS#11 soft token (and so perhaps the kernel provider should be used directly), and that also hardware offloading was slower for small blocks.

lib/hcrypto/evp-pkcs11.c
lhoward: Will fix, note this will require libhcrypto to link against libheimbase (not that this should be a problem).

lhoward:
OK, I believe I have addressed all of your comments (except for making the path configurable and breaking up commits).

kdc/connect.c
jaltman: The baselocl.h changes and the "s/tripple/triple" changes can also be their own unrelated commit.

jaltman:
AuriStor needs to support PKCS#11 for AES-NI on Solaris 10 and 11.

jaltman:
At the moment Heimdal is not building hcrypto against OpenSSL so there is no AES-NI.  Anything is better than AES-NI.  Now that we have the PCKS#11 and OpenSSL we can run performance tests.

lhoward:
Not sure if I understand the last comment @jaltman, could you not use Oracle's PKCS#11-enabled OpenSSL instead of hcrypto? (Indeed, why didn't we think of this earlier?)

lib/hcrypto/evp-pkcs11.c
lhoward: > On 29 Nov 2015, at 11:05 AM, Nico Williams <notifications@github.com> wrote:
> 
> Also, the "there's two paths, one for the system 32-bit PKCS11 provider, and the other for the system 64-bit PKCS11 provider" problem is likely to come up elsewhere. Maybe we need a lib/base utility for constructing dlopen()/LoadLibraryEx()/whatever paths that supports a template where whatever appropriate token for the bit-ness of the running process can be inserted.
> 
Is this a common issue though? On Solaris it seems to be a problem because uname/autoconf doesn’t give us a way to determine if we’re building 64-bit. On other platforms I think it’s a non-issue – OS X has universal binaries, and every other platform I can think of has 64-bit as a separate target.

— Luke

jaltman:
There are thread safety issues using OpenSSL from a library.  We have been unable to confirm that these issues have been addressed on Solaris.

Jeffrey Altman
Secure Endpoints Inc

vdukhovni:
On Sat, Nov 28, 2015 at 10:12:24PM -0800, Jeffrey Altman wrote:

> > On Nov 28, 2015, at 10:04 PM, Luke Howard <notifications@github.com> wrote:
> > 
> > Not sure if I understand the last comment @jaltman, could you not use
> > Oracle's PKCS#11-enabled OpenSSL instead of hcrypto? (Indeed, why didn't
> > we think of this earlier?)
> 
> There are thread safety issues using OpenSSL from a library.  We have been
> unable to confirm that these issues have been addressed on Solaris.

I would not expect any speedup from PKCS#11 (engine) support in
OpenSSL.  Is the goal here to gain access to hardware signature
tokens, or to a fast AES-NI implementation for SPARC?

The OpenSSL thread safety issues are primarily in the global PRNG
state, but perhaps not exclusively so.  It is not yet clear when
they will be addressed, though there is a glimmer of hope for 1.1.0,
more likely in 1.2.0.

-- 
	Viktor.

lhoward:
> I would not expect any speedup from PKCS#11 (engine) support in
> OpenSSL. Is the goal here to gain access to hardware signature
> tokens, or to a fast AES-NI implementation for SPARC?

The latter (I think for Intel actually but I guess Jeff cares about both architectures?).

libhx509 already has PKCS#11 glue for the former, I believe.

— Luke

jaltman:
The primary goal of pkcs11-hcrypto is to provide for Solaris the ability to use the native OS crypto engine on whatever hardware platform the OS is shipped for.   Much as CommonCrypto (evp-cc.c) does for OSX and iOS and Crypto Next Generation (evp-wincng.c) does for Windows.   This ensures that if AES-NI is available on Intel/AMD processors that the hcrypto consumer will benefit from it but also that if there is support for a custom hardware accelerator that it will be used.

The secondary goal of pkcs11-hcrypto is to permit hcrypto to be built so it can use any PKCS#11 hardware or software token for applications that require it.

vdukhovni:
On Sun, Nov 29, 2015 at 07:21:49AM -0800, Jeffrey Altman wrote:

> The primary goal of pkcs11-hcrypto is to provide for Solaris the ability
> to use the native OS crypto engine on whatever hardware platform the OS
> is shipped for.   Much as CommonCrypto (evp-cc.c) does for OSX and iOS
> and Crypto Next Generation (evp-wincng.c) does for Windows.   This ensures
> that if AES-NI is available on Intel/AMD processors that the hcrypto
> consumer will benefit from it but also that if there is support for a
> custom hardware accelerator that it will be used.
> 
> The secondary goal of pkcs11-hcrypto is to permit hcrypto to be built so
> it can use any PKCS#11 hardware or software token for applications that
> require it.

Right, thanks.  The first goal IIRC cannot be met via OpenSSL, and
in any case there are the aforementioned threading issues.  So
direct PKCS#11 support is required.  It remains to be seen whether
bulk crypto (AES rather than PKI) is sufficiently fast via PKCS#11.
If you're able to measure relative performance, please share the
results.

-- 
	Viktor.

PR #149, opened 2015-12-04
draft-ietf-kitten-aes-cts-hmac-sha2-07.txt implementation

lhoward:
Assuming draft-ietf-kitten-aes-cts-hmac-sha2-07.txt does not change before going to last call (premature assumption?), this is ready for review, will just need assigned numbers to be updated before merge.


jaltman:
Already merged into the tree.

PR #150, opened 2015-12-16
gss_wrap/gss_unwrap_aead implementation

lhoward:
Support for the gss_wrap_aead() and gss_unwrap_aead() API which has existed in MIT Kerberos since 2009 and is specified in draft-howard-gssapi-aead.

Differences to the MIT implementation:

* It is an API only, not an SPI. Mechanisms should implement the IOV API if they wish to support AEAD. This is based on practical experience – all of the AEAD-supporting third-party mechanisms of which I'm aware support export gss_wrap_iov() and friends. Also avoids APIxSPI cartesian explosion.

* Actually using it with Kerberos requires support for GSS_IOV_BUFFER_TYPE_STREAM, which is not yet in the Kerberos mechanism. This will come in a future patch (or see my aes-gcm branch).

### DIFF
--- a/lib/gssapi/mech/gss_aeap.c
+++ b/lib/gssapi/mech/gss_aeap.c
@@ -214,3 +214,121 @@ gss_context_query_attributes(OM_uint32 *minor_status,
 
     return GSS_S_FAILURE;
 }
+
+/*
+ * AEAD wrap API for a single piece of associated data, for compatibility
+ * with MIT and as specified by draft-howard-gssapi-aead-00.txt.
+ *
+ * @ingroup gssapi
+ */
+GSSAPI_LIB_FUNCTION OM_uint32 GSSAPI_LIB_CALL
+gss_wrap_aead(OM_uint32 *minor_status,
+	      gss_ctx_id_t context_handle,
+              int conf_req_flag,
+              gss_qop_t qop_req,
+              gss_buffer_t input_assoc_buffer,
+              gss_buffer_t input_payload_buffer,
+              int *conf_state,
+              gss_buffer_t output_message_buffer)
+{
+    OM_uint32 major_status, tmp, flags = 0;
+    gss_iov_buffer_desc iov[5];
+    size_t i;
+    unsigned char *p;
+
+    memset(iov, 0, sizeof(iov));
+
+    iov[0].type = GSS_IOV_BUFFER_TYPE_HEADER;
+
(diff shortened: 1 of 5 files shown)


lib/gssapi/mech/gss_aeap.c
```diff
@@ -214,3 +214,121 @@ gss_context_query_attributes(OM_uint32 *minor_status,
...
+ * with MIT and as specified by draft-howard-gssapi-aead-00.txt.
+ *
+ * @ingroup gssapi
+ */
+GSSAPI_LIB_FUNCTION OM_uint32 GSSAPI_LIB_CALL
+gss_wrap_aead(OM_uint32 *minor_status,
+	      gss_ctx_id_t context_handle,
+              int conf_req_flag,
+              gss_qop_t qop_req,
+              gss_buffer_t input_assoc_buffer,
```
nicowilliams: Should the inputs be gss_const_buffer_t?  They aren't in MIT krb5, but I think they should be.

lib/gssapi/mech/gss_aeap.c
```diff
@@ -214,3 +214,121 @@ gss_context_query_attributes(OM_uint32 *minor_status,
...
+    gss_iov_buffer_desc iov[3];
+
+    memset(iov, 0, sizeof(iov));
+
+    iov[0].type = GSS_IOV_BUFFER_TYPE_STREAM;
+    iov[0].buffer = *input_message_buffer;
+
+    iov[1].type = GSS_IOV_BUFFER_TYPE_SIGN_ONLY;
+    if (input_assoc_buffer)
+	iov[1].buffer = *input_assoc_buffer;
```
nicowilliams: Of course, using `gss_const_buffer_t` would then require a cast here. and at 318...

lib/gssapi/mech/gss_aeap.c
nicowilliams: I'll merge it as-is and fix it in a separate commit.

nicowilliams:
Cherry-picked onto master as 33ce593.  Thanks!

lib/gssapi/mech/gss_aeap.c
lhoward: Sure, makes sense. Actually porting mech_eap to current Heimdal was a bit of a pain owing to all the const changes - I wasn't entirely clear if const context and cred handles made sense always, eg sequence numbers changing in contexts.

PR #151, opened 2015-12-17
Refactoring conditional directives for incomplete if conditions.

wmarquesr:
A suggestion to compile entire statements and expressions, as suggested by code style guidelines from the Linux Kernel and practitioners.
- [Linux kernel coding style](https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/CodingStyle#n892)
- [Living in the #ifdef Hell](https://www.cqse.eu/en/blog/living-in-the-ifdef-hell/)

It might improve code understanding, maintainability and error-proneness.

### DIFF
--- a/lib/krb5/get_in_tkt.c
+++ b/lib/krb5/get_in_tkt.c
@@ -382,6 +382,7 @@ krb5_get_in_cred(krb5_context context,
     krb5_preauthdata *my_preauth = NULL;
     unsigned nonce;
     int done;
+		int check_directive;
 
     opts = int2KDCOptions(options);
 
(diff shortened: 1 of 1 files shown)


wmarquesr:
I understand, but what you think about this kind of refactoring?

jaltman:
I agree with elric1.  The history of the #f 0 block is that it was part of commit 4e5a924e40432daaddb9a8e70aed5bf950aa7a1b.  It has never been compiled as part of Heimdal and should just be removed.

As far as the Linux guideline to use IS_ENABLED(), Heimdal builds on platforms that are not Linux and use compilers other than gcc.   C Preprocessor logic is used to isolate platform specific functionality and we must restrict functionality to the lowest common denominator.

elric1:
I've pushed the change that removes the #if 0'd code entirely.

PR #153, opened 2016-01-28
Fix parallel build

brianmay:
Fix problem due to missing makefile dependency on a generated file.

See http://bugs.debian.org/800728 for details


nicowilliams:
Cherry-picked with 6b59333.  Thanks!

PR #157, opened 2016-02-20
Fix race condition with global _gsskrb5_keytab

yaheath:
gsskrb5_acceptor_start() was making a copy of the global pointer
_gsskrb5_keytab to use later. This invites a race condition where
another thread could call gsskrb5_register_acceptor_identity()
(thus invalidating the target of the copied pointer) before it is
used by gsskrb5_acceptor_start().

So instead, clone the keytab to a new one while protected by the
mutex lock (similar to get_keytab() in acquire_cred.c).


nicowilliams:
Merged as 545b5b4.  Thanks!

PR #161, opened 2016-03-09
Fix bias in ordering SRV RR results by weight.

elric1:
In lib/roken/resolve.c, we find rk_dns_srv_order() which re-orders
the results of an SRV RR lookup by the algorithm in RFC2782.  We
fix a bias in the random weight sorting by changing the order of
operations when selecting rnd.  rnd should be a non-zero random
number less than the sum of the weights at a particular priority,
but zero was included as a legitimate output thus biasing the
selection process.  rk_random() % sum is still biased as a 32
bit int modulo a number which doesn't divide 2^32 does not have
a uniform distribution, but the bias should be small enough to
live with for our purposes here.

### DIFF
--- a/lib/roken/resolve.c
+++ b/lib/roken/resolve.c
@@ -660,16 +660,20 @@ rk_dns_srv_order(struct rk_dns_reply *r)
     headp = &r->head;
 
     for(ss = srvs; ss < srvs + num_srv; ) {
-	int sum, rnd, count;
+	int sum, zeros, rnd, count;
 	struct rk_resource_record **ee, **tt;
 	/* find the last record with the same priority and count the
            sum of all weights */
-	for(sum = 0, tt = ss; tt < srvs + num_srv; tt++) {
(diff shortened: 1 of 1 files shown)


lib/roken/resolve.c
```diff
@@ -690,7 +696,9 @@ rk_dns_srv_order(struct rk_dns_reply *r)
 	    (*tt)->next = *headp;
 	    *headp = *tt;
 	    headp = &(*tt)->next;
-	    sum -= (*tt)->u.srv->weight;
+	    sum -= (*tt)->u.srv->weight * zeros;
+	    if ((*tt)->u.srv->weight == 0)
+		sum--;
 	    *tt = NULL;
```
vdukhovni: This (*tt = NULL) part had me confused for a bit...  While it is not new with this patch and should be improved separately, for the record instead of filling the array with NULLs to skip, and adding code to skip NULLs, ... it would surely be much simpler to swap the selected element with (*ss) and then increment ss.  May well do that, when I get a chance...

PR #162, opened 2016-03-10
lib/krb5/pac.c: fix NULL check typo

aaptel:
ucs2 was already NULL checked earlier, this is never reached (and is doing a free(NULL), which man free(3) says it's a no-op) so no worries there.

You actually want to NULL-check s2 who has just received a malloc() result.


lha:
thanks

PR #163, opened 2016-03-10
lib/gssapi/mech/gss_compare_name.c: fix memcmp() call

aaptel:
Make memcmp() compare the name1 and name2 value instead of comparing
name1 with itself.

The memcmp() is only executed if the left-hand side of the || is false
i.e. when both length are equal so the length argument is correct (no out-of-bounds reads).

I'm confused about the `$FreeBSD: ...$` tag, is this originally from FreeBSD SVN? Should I send this change on their mailing list?


lha:
thanks!

PR #167, opened 2016-04-07
Fix build when OpenSSL does not have EGD support

Sp1l:
- EGD is disabled by default in OpenSSL 1.1.0
- EGD was removed in LibreSSL
- No modern platform requires EGD


Sp1l:
This has been successfully tested on FreeBSD 10.3

PR #172, opened 2016-04-22
Fix QAS Bug 21711: Kerberos BAD_INTEGRITY error with 2008 R2 RODC

DavidMulder:
BAD_INTEGRITY error when trying to kinit:

ERROR: VAS_ERR_KRB5: Failed to obtain credentials. Client: SLED10-32$@F.QAS,
Service: SLED10-32$@F.QAS, Server: ad2-f.f.qas
   Caused by:
   KRB5KRB_AP_ERR_BAD_INTEGRITY (-1765328353): Decrypt integrity check failed

MS changed code for RODCs in 2008R2.  MS does state that ALL krgtgt/REALM tickets SHOULD be sent using principal name type of KRB5_NT_SRV_INST, instead of KRB5_NT_PRINCIPAL.

From Microsoft:

"I believe we discovered the problem.  There isn’t a bug in Windows. 
There’s been a code change to address another issue which puts in additional
checks for Kerberos tickets.  The problem is with the Unix clients when the
client request a TGT.  The Unix clients are using Name-type Principal
[KRB_NT_PRINCIPAL  (1)] instead of using Name-type Service and Instance
[KRB_NT_SRV_INST  (2)]...."  

Further correspondence suggests that other tickets should be formatted to use:

"TGT’s should always be KRB5_NT_SRV_INST.  As far as principal type, it
depends on the name type being used.  If it’s in the form of http/ or
mssqlsvc/, etc then it would use KRB5_NT_SRV_INST.  If it is host/ then it
would be KRB_NT_SRV_HST.  If it’s in the form of a UPN then it would be
KRB_NT_ENTERPRISE_PRINCIPAL."

### DIFF
--- a/lib/krb5/principal.c
+++ b/lib/krb5/principal.c
@@ -211,6 +211,10 @@ krb5_parse_name_flags(krb5_context context,
     int enterprise = flags & KRB5_PRINCIPAL_PARSE_ENTERPRISE;
     int ignore_realm = flags & KRB5_PRINCIPAL_PARSE_IGNORE_REALM;
     int no_def_realm = flags & KRB5_PRINCIPAL_PARSE_NO_DEF_REALM;
+    int srv_and_inst = 0;
+
+    if( !strncasecmp( name, KRB5_TGS_NAME, 6 ) )
+        srv_and_inst = 1;
 
     *principal = NULL;
 
@@ -348,6 +352,9 @@ krb5_parse_name_flags(krb5_context context,
     }
     if (enterprise)
 	(*principal)->name.name_type = KRB5_NT_ENTERPRISE_PRINCIPAL;
+    else
+    if( srv_and_inst )
+    (*principal)->name.name_type = KRB5_NT_SRV_INST;
     else
 	(*principal)->name.name_type = KRB5_NT_PRINCIPAL;
     (*principal)->name.name_string.val = comp;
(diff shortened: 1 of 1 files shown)


jaltman:
The formatting for this change needs to be fixed before it can be merged into the tree.  

 Space between "if and (" because "if" is not a function

 Indentation of body of if statement

The commit message should describe the problem and not just the solution.

Thank you.

DavidMulder:
I modified the code style and updated the commit message. Is that sufficient?

jaltman:
The final result in correct but please squash the changes into a single commit.

kaduk:
FWIW you can force-push to your personal github branch with the updated (squashed) commit, and the pull request will update accordingly automatically; there's no need to close and open a new pull request.

DavidMulder:
I normally do. For some reason it wasn't accepting the force push. I'm sure I was doing something wrong.

> git reset --hard HEAD~2
HEAD is now at ca0dff8 Check HCRYPTO_FALLBACK in W32 backend
> patch principal.c <~/code/Bug21711.patch
patching file principal.c
> git commit principal.c 
[detached HEAD 9103a2e] Fix QAS Bug 21711: Kerberos BAD_INTEGRITY error with 2008 R2 RODC
 1 file changed, 6 insertions(+)
> git push -f origin master
Everything up-to-date

DavidMulder:
I think I should have done a push force after the reset, to rewrite the branch, then patch and push.

kaduk:
I am not entirely sure how you got there (as "git reset --hard" should keep you on a branch), but this is the important part:

> [detached HEAD 9103a2e] 

So you were not on branch 'master' anymore.  "git push origin +HEAD:master" should have worked, though.

PR #173, opened 2016-04-22
Fix QAS Bug 21711: Kerberos BAD_INTEGRITY error with 2008 R2 RODC

dmulder:
BAD_INTEGRITY error when trying to kinit:

ERROR: VAS_ERR_KRB5: Failed to obtain credentials. Client: SLED10-32$@F.QAS,
Service: SLED10-32$@F.QAS, Server: ad2-f.f.qas
Caused by:
KRB5KRB_AP_ERR_BAD_INTEGRITY (-1765328353): Decrypt integrity check failed

MS changed code for RODCs in 2008R2. MS does state that ALL krgtgt/REALM tickets SHOULD be sent using principal name type of KRB5_NT_SRV_INST, instead of KRB5_NT_PRINCIPAL.

From Microsoft:

"I believe we discovered the problem. There isn’t a bug in Windows. 
There’s been a code change to address another issue which puts in additional
checks for Kerberos tickets. The problem is with the Unix clients when the
client request a TGT. The Unix clients are using Name-type Principal
[KRB_NT_PRINCIPAL (1)] instead of using Name-type Service and Instance
[KRB_NT_SRV_INST (2)]...."

Further correspondence suggests that other tickets should be formatted to use:

"TGT’s should always be KRB5_NT_SRV_INST. As far as principal type, it
depends on the name type being used. If it’s in the form of http/ or
mssqlsvc/, etc then it would use KRB5_NT_SRV_INST. If it is host/ then it
would be KRB_NT_SRV_HST. If it’s in the form of a UPN then it would be
KRB_NT_ENTERPRISE_PRINCIPAL."

### DIFF
--- a/lib/krb5/principal.c
+++ b/lib/krb5/principal.c
@@ -348,6 +348,10 @@ krb5_parse_name_flags(krb5_context context,
     }
     if (enterprise)
 	(*principal)->name.name_type = KRB5_NT_ENTERPRISE_PRINCIPAL;
+    else if (ncomp > 1 && !strcmp(comp[0], KRB5_TGS_NAME))
+        (*principal)->name.name_type = KRB5_NT_SRV_INST;
+    else if (ncomp > 1 && !strcmp(comp[0], "host"))
+        (*principal)->name.name_type = KRB5_NT_SRV_HST;
     else
 	(*principal)->name.name_type = KRB5_NT_PRINCIPAL;
     (*principal)->name.name_string.val = comp;


lib/krb5/principal.c
```diff
@@ -211,6 +211,10 @@ krb5_parse_name_flags(krb5_context context,
     int enterprise = flags & KRB5_PRINCIPAL_PARSE_ENTERPRISE;
     int ignore_realm = flags & KRB5_PRINCIPAL_PARSE_IGNORE_REALM;
     int no_def_realm = flags & KRB5_PRINCIPAL_PARSE_NO_DEF_REALM;
+    int srv_and_inst = 0;
+
+    if (!strncasecmp(name, KRB5_TGS_NAME, 6))
```
jaltman: A couple of things.  First, instead of "6" the constant KRB5_TGS_NAME_SIZE should be used if it were appropriate to be using strncasecmp() but it isn't safe to do so.  What this statement compares is up to 6 characters of name to "krbtgt".  But it will succeed with if the first component of name is "krbtgt" or if it is "krbtgtxxxxxxxxxx".  

What is necessary is to ensure that the first component is a match for both string and length.

A second question that comes to mind is whether or not this is required to be a case insensitive comparison.   Although AD is not case sensitive, Kerberos is case sensitive.  I will ask therefore for some evidence that name can ever be a case variant of "krbtgt".

jaltman:
Thank you for the updated commit message.  Now we just have to get the change to be functionally correct.

lib/krb5/principal.c
jaltman: Later on in this function the comp[] array will be populated with C-strings.   The string that should be compared to KRB5_TGS_NAME is comp[0].

lib/krb5/principal.c
```diff
@@ -348,6 +352,8 @@ krb5_parse_name_flags(krb5_context context,
     }
     if (enterprise)
 	(*principal)->name.name_type = KRB5_NT_ENTERPRISE_PRINCIPAL;
+    else if (srv_and_inst)
```
jaltman: We can avoid the srv_and_inst variable entirely and use a comparison of 

   if (ncomp > 1 && strcmp(comp[0], KRB5_TGS_NAME) == 0)

I would feel safer with the comparison being case sensitive.
What do you think?

jaltman:
Another question that stems from the commit message.   Should another conditional block

   else if (ncomp > 1 && strcmp("host", comp[0]) == 0)
        (*principal)->name.name_type = KRB5_NT_SRV_HST;

be added?

nicowilliams:
@jaltman: "no", no such conditional should be added.

This whole thing has me peeved.  Name types in Kerberos do not partition the namespace (unlike in GSS generic query syntax), and are informative, not normative.  No KDC has any business in demanding that a client use a particular name type or other when referring to any principals (client or service principals alike), and IIRC there is no text in RFC4120 (or any other related RFC, nor 1510) supporting such a requirement.

Indeed, neither MIT Kerberos nor Heimdal even keep track of name-types in their KDC principal databases, because the protocol never required them to.

DavidMulder:
@nicowilliams I tend to agree. If this violates standards, then I'm not sure it belongs in heimdal.
@jaltman what is your opinion?

abartlet:
I would strongly argue for maximum compatibility with AD here, due to wide deployment. 

Otherwise, you just push the problem on to the calling applications, like Samba, which have had to work around this for years now.

DavidMulder:
@nicowilliams 
Check RFC4120 page 99 for principle name types.
NT-PRINCIPAL      1    Just the name of the principal as in DCE,
                            or for users
NT-SRV-INST       2    Service and other unique instance (krbtgt)
NT-SRV-HST        3    Service with host name as instance
                            (telnet, rcommands)
etc...

jaltman:
@DavidMulder I'm unlikely to take this change as it.   I want to merge something that handles the other cases as well.   I want to review the history of RFC1510 from the mailing lists.  I simply haven't had time this week but should be able to research next week.

DavidMulder:
@jaltman Have you had a chance to research this?

jaltman:
@DavidMulder My apologies for the delay in responding but the subject matter required a bit of thought.  

Let me first address the question of principals with the service name "krbtgt".    My reading of RFC4120 is that all principals with a service name of "krbtgt" should be treated as type NT-SRV_INST either whenever the principal type is NT-UNKNOWN.   NT-SRV-INST principal types are treated special in that a request for a referral is implied without requiring the canonicalize flag.   In Heimdal this special casing of NT-SRV-INST occurs in kdc/krb5tgs.c need_referral().    Problems will occur if the client creates a principal of type NT-PRINCIPAL when the actual type is unknown.  

One question that this raises is, what does Windows do when the service name is "krbtgt" but the principal type is NT-UNKNOWN?   By my reading of the text, Windows should work fine in that case.

I agree that krb5_parse_name_flags() should be modified to set the type to KRB5_NT_SRV_INST when the service name is "krbtgt" (aka KRB5_TGS_NAME).   Lets commit that as a single isolated change.

RFC4120 says 

> If the first component of a name identifies a service and the remaining components identify an instance of the service in a server-specified manner, then the name type of SRV-INST SHOULD be   used.  An example of this name type is the Kerberos ticket-granting service whose name has a first component of krbtgt and a second component identifying the realm for which the ticket is valid.

This implies that a principal of type NT-SRV-INST should not be assumed to be a "krbtgt" or that the same rules apply to the principal that apply to "krbtgt".

Heimdal's kdc/krb5tgs.c need_referral() behavior needs to be altered to ensure that if the principal type is NT-UNKNOWN and the service name is "krbtgt" then the behavior is the same as if the type was NT-SRV-INST and service name "krbtgt".  need_referral() currently does not validate the service name nor does it handle the case of NT-UNKNOWN and "krbtgt".

I believe that the KDC should whenever it receives an incoming request with principal type NT-UNKNOWN apply transformation rules to assign an appropriate type for internal processing.  However, the service name principal type returned to the caller must be NT-UNKNOWN if the client specified NT-UNKNOWN.  Otherwise, the Java 1.6 Kerberos client implementation will fail to accept the response.  It would be worth testing to determine if this behavior has been altered in subsequent releases.   The fixes for the need_referral() behavior should be a separate commits.

The next question is should a principal whose first component is the service name "host" be treated as type NT-SRV_HST when the type is NT-UNKNOWN.   I believe the answer is "yes" both in the krb5 client, the kdc and in GSS.   RFC 4120 says:

> If the first component of a name identifies a service and there is a single component following the service name identifying the instance as the host on which the server is running, then the name type   SRV-HST SHOULD be used.  This type is typically used for Internet services such as telnet and the Berkeley R commands.  

As such, designating an unknown principal with first component "host" must only be considered NT-SRV-HST when the number of components is 2.

RFC 4120 continues with:

> If the separate components of the host name appear as successive components following the name of the service, then the name type SRV-XHST SHOULD be used.

Heimdal currently does not properly handle NT-SRV-XHST in lib/krb5 and in lib/gssapi/krb5.

RFC 4120 finishes with:

> A name type of UNKNOWN SHOULD be used when the form of the name is not known.  When comparing names, a name of type UNKNOWN will match principals authenticated with names of any type.  A principal authenticated with a name of type UNKNOWN, however, will only match other names of type UNKNOWN.

However, Heimdal's krb5_parse_name_flags() sets the principal type to NT-PRINCIPAL not NT-UNKNOWN when all other options are exhausted.

RFC 5179 "Generic Security Service Application Program Interface (GSS-API) Domain-Based Service Names Mapping for the Kerberos V GSS Mechanism" defines NT-SRV-HST-DOMAIN which is similar to NT-SRV-HST but supports 3 components.  First component is the service, second is the host, and third is an application specific domain.  Heimdal currently does not support NT-SRV-HST-DOMAIN.  It does not apply to the service "host".

Finally, Heimdal requires configuration support to permit specifying service names that should be treated as NT-SRV-HST or NT-SRV-HST-DOMAIN when processing krb5_parse_name_flags(), in the kdc, and in gss.  I propose 

  [service_principal_type]
       service = type
       service = type
          ....
        service = type

where the types are those specified in lib/krb5/principal.c nametypes[].    This mapping should be supported by the krb5 client, the kdc and gss.   This mechanism will permit the appropriate assignment of principal types by the client and the appropriate mapping of NT-UNKNOWN to principal type by the KDC.

There are a significant number of to do items and quite a bit of interop testing that will be required.  Can your employer assist with this work?

jaltman:
There are many more places that just krb5_parse_name_flags() that fails to set the proper principal type for "krbtgt".

nicowilliams:
Pushed:

22790e4 Test RODC interop fix
9e2b696 Make kdc name type strictness configurable
961f543 Set princ type to NT-SMTP-NAME when parsing
6a1db3f princ type NT-UNKNOWN + "host" == NT-SRV-HST
5aef50c gss-krb5: do_delegate remove dead comment
09bdb3a Set the right name type for anon princ (client)
020f2c7 kdc: principals of type NT-UNKNOWN can be anonymous
a59bb71 When building a princ name pick a sane def type

PR #176, opened 2016-05-07
Revert "KVNOs are krb5uint32 in RFC4120, make it so"

urisimchoni:
This reverts commit 1124c4872dfb81bec9c4b527b8927ca35e39a599.

A reason for deviation from RFC4120 is backward compatibility.
With Windows RODC's, it is likely to have a kvno for the tgt which
has bit 31 set. Encoding such a kvno as unsigned would lead to a
5-byte number, whereas Windows, which treats kvno as int32, expects
up to 4 bytes. MIT Kerberos does the same for the exact same reason.

The encoding has already been reverted in commit
060474df1626bcc597ef6e786cc58d9b2daac8db. After reverting the encoding,
there's no reason to use krb5uint32 in the context of kvno, so this
commit reverts the rest, plus removes one more use of krb5uint32 in
the context of kvno.

Signed-off-by: Uri Simchoni <uri@samba.org>

### DIFF
--- a/kdc/krb5tgs.c
+++ b/kdc/krb5tgs.c
@@ -1572,6 +1572,8 @@ tgs_build_reply(krb5_context context,
 	hdb_entry_ex *uu;
 	krb5_principal p;
 	Key *uukey;
+	krb5uint32 second_kvno = 0;
+	krb5uint32 *kvno_ptr = NULL;
 
 	if(b->additional_tickets == NULL ||
 	   b->additional_tickets->len == 0){
@@ -1588,8 +1590,12 @@ tgs_build_reply(krb5_context context,
 	    goto out;
 	}
 	_krb5_principalname2krb5_principal(context, &p, t->sname, t->realm);
+	if(t->enc_part.kvno){
+	    second_kvno = *t->enc_part.kvno;
+	    kvno_ptr = &second_kvno;
+	}
 	ret = _kdc_db_fetch(context, config, p,
-			    HDB_F_GET_KRBTGT, t->enc_part.kvno,
+			    HDB_F_GET_KRBTGT, kvno_ptr,
 			    NULL, &uu);
 	krb5_free_principal(context, p);
 	if(ret){


jaltman:
Key version numbers should never be negative.  For that reason I would prefer that an unsigned type be used.

jaltman:
To be clear what the RODC use of the 31st bit is being used for.  It is not part of the kvno.  It's a flag to indicate that it came from an RODC.   If we want to do something with it we should be masking it off and storing the rest as the kvno.

lhoward:
You saw this yes?

https://blogs.msdn.microsoft.com/openspecification/2011/05/11/notes-on-kerberos-kvno-in-windows-rodc-environment/ <https://blogs.msdn.microsoft.com/openspecification/2011/05/11/notes-on-kerberos-kvno-in-windows-rodc-environment/>

tlyu:
My recollection is that the high-bit 32-bit kvno is being encoded as a negative ASN.1 INTEGER, therefore taking only 4 bytes at most in the encoding. (An unsigned 32-bit kvno with the high bit set would be encoded in 5 bytes, because all ASN.1 INTEGER values are encoded as signed in BER.) It is of course your choice how to represent it internally, as long as the wire encoding is as expected.

jaltman:
@lhoward I've read that post the n the past.  Heimdal needs to encode as signed integer for the asn.1 to be properly generated but I think we should be representing internally as an unsigned integer,   Note that Microsoft is capping the max kvno value at 0xFFFF.  The upper 16-bits represent the unsigned RODC id.  

@urisimchoni  Is Samba supporting the RODC mode?

urisimchoni:
Here's a modified patch which keeps the interfaces unsigned, and just fixes a signed-unsigned pointer mismatch.

Samba supports RODC mode. The db glue code extracts the RODC-id from the kvno.

urisimchoni:
Ping...
Can you please review the modified patch? it just handles what seems to be an implicit conversion of incompatible pointers (signed and unsigned). GCC generally allows this but "picky developer builds" and code checkers don't like it.

Thanks,
Uri.

PR #177, opened 2016-05-15
Don't put system architecture in generated headers.

jelmer:
Including HOST causes the build to be unreproducible.

See also https://tests.reproducible-builds.org/rb-pkg/testing/i386/heimdal.html

### DIFF
--- a/include/bits.c
+++ b/include/bits.c
@@ -141,9 +141,6 @@ int main(int argc, char **argv)
 	}
 	f = fopen(argv[1], "w");
     }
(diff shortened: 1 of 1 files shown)


abartlet:
This is important, the extra detail is not worth making the thing unreproducible.

jaltman:
@abartlet thank you for merging this and the others you have merged over the last few days.  I'm receiving e-mail notifications for the merge but not for the original pull request.  I need to figure out why.

PR #181, opened 2016-06-17
krb5_c_random_make_octets incorrectly implemented

UnitedMarsupials:
The function, found in `lib/krb5/mit_glue.c`, is currently using `krb5_generate_random_keyblock()`. This compiles because warning-level is not high enough, but does not work. At runtime the `krb5_generate_random_keyblock()` interprets the second argument as the `krb5_enctype` (rather than a _length_ of anything) and tries to verify it.

When the length does not match any known enctype, as usually happens, the function fails and returns an error. If the length happened to correspond to an enctype, the function would likely crash due to misinterpreting its third argument as a valid `krb5_keyblock`.

The change uses `krb5_generate_random_block()` instead. This function does not return anything -- upon detecting failure it will cause the entire application to exist instead...

### DIFF
--- a/lib/krb5/mit_glue.c
+++ b/lib/krb5/mit_glue.c
@@ -378,7 +378,8 @@ krb5_c_prf(krb5_context context,
 KRB5_LIB_FUNCTION krb5_error_code KRB5_LIB_CALL
 krb5_c_random_make_octets(krb5_context context, krb5_data * data)
 {
-    return krb5_generate_random_keyblock(context, data->length, data->data);
+    krb5_generate_random_block(context, data->length, data->data);
+    return 0;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


jaltman:
While the overall change is correct could you please squash the two commits into a single commit and resubmit the pull request.

UnitedMarsupials:
> resubmit the pull request

@jaltman, seriously? Should I also wear a jacket, while doing it? It is a one-line change, for crying out loud!

It is bad enough, you did not _thank me_, you want me to do more work? Ah, and then you were going to ask me to submit a separate pull-request for each of Heimdal's mutliple branches, were you not? 

Sorry. I raised your awareness of the problem -- and even provided a fix. If it is not in a format you can use in the browser -- too bad.

jaltman:
@UnitedMarsupials Agreed.  Its a one line change submitted as a pull request with two commits that change the same line twice.  I can't accept the pull up request as-is.   One of us will need to do more work.  As a user of Heimdal you have gained the benefit of the work that others have put into its development and maintenance.  No one is paid to work on Heimdal.  The project relies entirely upon volunteer efforts.  I am closing this request and manually submit the change as a single commit. 

9f6baf00f6b7c6f4bb6a231cf1753234b9f6148d

PR #185, opened 2016-07-21
Remove duplicate symbols from libroken version-script.map

emaste:
Commit efed5633 (r24759) prefixed some symbols with rk_, but introduced 6 duplicate symbols in the version script (because the rk_-prefixed versions of the symbols were already present).


lha:
thanks

PR #186, opened 2016-07-23
Honor krb5.conf-configured etypes for GSSAPI TGS request

urisimchoni:
When sending a TGS request as part of GSSAPI client-side
context negotiation, use for etypes the value in the
credentials, and if that's not set - the values in the
configuration.

When the krb5 context is created, and krb5.conf is read, the allowed enctyption types are set to whatever is in krb5.conf (or the full set as default). Then, without this patch, if no enc types are associated with the credentials, the enc types are reset to the full set. This would not allow limiting the encryption types via configuration.

It seems like the current code is intentional. I was just unable to figure out the intent (I get the part of overriding the etypes with the stuff in the credentials, but not the part of resetting to default). Therefore I'm sending over this pull request - to be accepted or rejected with an explanation.

### DIFF
--- a/lib/krb5/context.c
+++ b/lib/krb5/context.c
@@ -48,6 +48,11 @@
         }								\
     } while(0)
 
+static krb5_error_code
+copy_enctypes(krb5_context context,
+	      const krb5_enctype *in,
+	      krb5_enctype **out);
+
 /*
  * Set the list of etypes `ret_etypes' from the configuration variable
  * `name'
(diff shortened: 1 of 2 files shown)


jaltman:
@lha do you remember the reason for resetting to the default set of enctypes in d3807ad88238f9e63794f0acafbe6a9fd2c7b959 ?

urisimchoni:
Ping...

abartlet:
This change is needed for Samba, but there is an issue with it generally.

I've looked over it carefully, and the reason that it is set to NULL is that otherwise it will tatoo the long-term context with the encryption types of any one credential.

We need to extend the krb5_context to have a config_etypes variable, initially set the same, and reset etypes to that variable if NULL is specified.

abartlet:
To be clear, the lifetime of context in init_auth is very long, as under the hood it is essentially per-thread static via the GSSAPI_KRB5_INIT() -> _gsskrb5_init() code.

urisimchoni:
New patch version tackles the problem as Andrew suggested. I've verified that it also fixes things for Samba, with original patch reverted.

PR #187, opened 2016-08-02
Merge fixes from Dell into heimdal master

DavidMulder:
A small sample of fixes in Dell's heimdal branch. More will follow after extensive code review.
- The --disable-afs-support doesn't really disable afs. It still gets built, and kinit still links to it, although it doesn't use it.
- Plus various fixes for compiler errors on various platforms (mostly hpux, aix, solaris, and mac).


jaltman:
@DavidMulder I would prefer to see separate patches for each of the various platforms.  Ideally the commit messages would include the error/warning message from the compiler along with the compiler version information as well as the platform.

One patch for HPUX, one for AIX, one for Solaris, etc.

The --disable-afs-support change would also be its own commit.

DavidMulder:
I'll see what I can do.

PR #188, opened 2016-08-02
--disable-afs-support doesn't really disable afs

DavidMulder:
The --disable-afs-support configure flag doesn't really disable afs. It still gets built, and kinit still links to it, although it doesn't use it.


jaltman:
thank you.   commit merged.

PR #190, opened 2016-08-02
The evp-w32.h header should not be included on all platforms

DavidMulder:
This caused undefined symbol errors on multiple platforms. I can't remember which. Seems like it was x86 solaris and ia64 hpux.
gcc 4.6.1


jaltman:
The conditional should be

  #ifdef _WIN32

not _WIN32_

DavidMulder:
Fixed

PR #191, opened 2016-08-02
Redefinition errors on ppc64 linux

DavidMulder:
These defines cause redefinition warnings on most platforms/compilers. On ppc64 linux with gcc 4.7 it caused an error.

### DIFF
--- a/lib/roken/roken.h.in
+++ b/lib/roken/roken.h.in
@@ -1230,6 +1230,7 @@ int ROKEN_LIB_FUNCTION rk_socket(int, int, int);
 #endif
 
 /* Microsoft VC 2010 POSIX definitions */
+#if _WIN32_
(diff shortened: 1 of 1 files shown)


jaltman:
The correct define for Windows is `_WIN32` but the correct fix for this is to include the header file in roken.h.in that defines `EAFNOSUPPORT` and `EWOULDBLOCK`.

DavidMulder:
Those defines are in errno.h, which if you look in roken.h.in, it's already included.
Maybe the failure here is that HAVE_ERRNO_H isn't defined? That should be defined in config.h, but I don't see config.h included anywhere in roken.h.in.

lhoward:
@dmulder, where are the redefinition warnings coming from? Perhaps the problem is that they are not including config.h before roken.h?

PR #192, opened 2016-08-03
Solaris 8 x86 uses ->d_fd not ->dd_fd or dirfd()

dmulder:
Here's a stack overflow article for reference:
http://stackoverflow.com/questions/28022507/solaris-10-alternative-to-dirfd

On Solaris 8 x86 there is no dirfd() or ->dd_fd, so we need to check for ->d_fd also.


nicowilliams:
Thanks!

PR #195, opened 2016-08-09
krb5_check_transited() ret variable is unused

DavidMulder:
On aix 5.3 and solaris 10 x64

transited.c: In function 'krb5_check_transited':
transited.c:607:21: error: variable 'ret' set but not used [-Werror=unused-but-set-variable]

gcc 4.6.2 on both systems.

Also,

transited.c: In function '_krb5_find_capath':
transited.c:487:16: error: 'len' may be used uninitialized in this function [-Werror=uninitialized]
transited.c:450:12: note: 'len' was declared here

Same systems and gcc

Fixing commit 1501740952a815fbe10ab8345bb38cfb0ded9504

### DIFF
--- a/lib/krb5/transited.c
+++ b/lib/krb5/transited.c
@@ -447,7 +447,7 @@ hier_init(struct hier_iter *state, const char *local_realm, const char *server_r
 {
     size_t llen;
     size_t slen;
-    size_t len;
+    size_t len = 0;
     const char *lr;
     const char *sr;
 
@@ -604,13 +604,12 @@ krb5_check_transited(krb5_context context,
 		     unsigned int num_realms,
 		     int *bad_realm)
 {
-    krb5_error_code ret = 0;
     char **capath = NULL;
     size_t num_capath = 0;
     size_t i = 0;
     size_t j = 0;
 
-    ret = _krb5_find_capath(context, client_realm, client_realm, server_realm,
+    _krb5_find_capath(context, client_realm, client_realm, server_realm,
                             TRUE, &capath, &num_capath);
 
     for (i = 0; i < num_realms; i++) {


lib/krb5/transited.c
```diff
@@ -447,7 +447,7 @@ hier_init(struct hier_iter *state, const char *local_realm, const char *server_r
 {
     size_t llen;
     size_t slen;
-    size_t len;
+    size_t len = 0;
```
jaltman: unrelated change?

DavidMulder: All related to Viktor Dukhovni recent changes.

lib/krb5/transited.c
```diff
@@ -604,13 +604,12 @@ krb5_check_transited(krb5_context context,
...
 {
-    krb5_error_code ret = 0;
     char **capath = NULL;
     size_t num_capath = 0;
     size_t i = 0;
     size_t j = 0;
 
-    ret = _krb5_find_capath(context, client_realm, client_realm, server_realm,
+    _krb5_find_capath(context, client_realm, client_realm, server_realm,
                             TRUE, &capath, &num_capath);
```
jaltman: why isn't the answer to check the return code?

DavidMulder: I assumed he ignored the return code on purpose. I don't know his reasoning.
I could add a
`if (ret) return ret;`

jaltman:
@nicowilliams Please review and fix since you merged Viktor's change.

@DavidMulder Please include the sha1 of the commit you are fixing in the commit message.

lib/krb5/transited.c
vdukhovni: That's the correct fix, if properly indented.  I can commit and push that.  Or you can go ahead and update the PR.

vdukhovni: I went ahead and pushed both changes.

PR #197, opened 2016-08-10
Related to vas segfault from re-defining krb5_krbhst_data

DavidMulder:
At some point a VAS dev re-defined krb5_krbhst_data because we only get a stub from the header file. When I updated our heimdal version, the VAS re-definition was different from the heimdal definition, which ended up corrupting some values in the structure when passed to VAS source.

I moved the definition of krb5_krbhst_data into krb5.h to solve that problem, but figured I'd push this upstream since it just makes since to be there instead of in the source file.

### DIFF
--- a/lib/krb5/krb5.h
+++ b/lib/krb5/krb5.h
@@ -805,16 +805,6 @@ typedef struct krb5_verify_opt {
 
 #define KPASSWD_PORT 464
 
-/* types for the new krbhst interface */
-struct krb5_krbhst_data;
-typedef struct krb5_krbhst_data *krb5_krbhst_handle;
-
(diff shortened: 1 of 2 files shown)


jaltman:
krb5.h is for public data structures only.   krb5_krbhst_data is a private data structure that should be opaque to public consumers of krb5.h.   

Why does VAS need to access the private data structure?

DavidMulder:
Looks like it's part of our hairy sendto code.
It's because we're injecting our own sendto code inside of krb5_sendto_context() (used to be krb5_sendto()).
Feel free to reject this. I'll be pulling those changes out anyway. Sorry I missed that.

PR #198, opened 2016-08-10
On hpux, pthread_cond_t is undefined when pthread is disabled

dmulder:
Since pthread_cond_t is undefined, if I disable pthread with --disable-pthread-support, it should try to create this struct. It isn't used anyway, since the only place it's used is inside a #if ENABLE_PTHREAD_SUPPORT block.

I think this was on hpux pa 11.11 with gcc 4.7.1.

### DIFF
--- a/lib/ipc/common.c
+++ b/lib/ipc/common.c
@@ -94,7 +94,9 @@ _heim_ipc_create_cred(uid_t uid, gid_t gid, pid_t pid, pid_t session, heim_icred
 #ifndef HAVE_GCD
 struct heim_isemaphore {
     HEIMDAL_MUTEX mutex;
+#ifdef ENABLE_PTHREAD_SUPPORT
     pthread_cond_t cond;
+#endif
     long counter;
 };
 #endif


lib/ipc/common.c
```diff
@@ -91,7 +91,7 @@ _heim_ipc_create_cred(uid_t uid, gid_t gid, pid_t pid, pid_t session, heim_icred
     return 0;
 }
 
-#ifndef HAVE_GCD
+#if !defined(HAVE_GCD) && defined(ENABLE_PTHREAD_SUPPORT)
```
jaltman: This will break the build for !defined(HAVE_GCD) && defined(ENABLE_PTHREAD_SUPPORT) because the heim_issemaphore type will be undefined.  The type is used in the prototype of heim_ipc_semaphore_create() at line 102.

Perhaps #ifdef ENABLE_PTHREAD_SUPPORT the 'cond' field in the heim_isemaphore structure.

DavidMulder: Ah, I missed that probably because of the stub for heim_isemaphore in heim-ipc.h. It compiled, but you're right, this is wrong.

PR #201, opened 2016-08-30
Use C99 designated initializers in a couple places

kaduk:
I had considered preparing a broader change to use designated initializers everywhere, or everywhere in hcrypto, but it's unclear that the benefit of designated initializers is worth the global churn, so I figured I would make a pull request with the targeted changes and see what response it got.

### DIFF
--- a/lib/hcrypto/rand-timer.c
+++ b/lib/hcrypto/rand-timer.c
@@ -183,6 +183,16 @@ timer_status(void)
 #endif
 }
 
+#if defined(__GUNC__) || (defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901)
+const RAND_METHOD hc_rand_timer_method = {
+    .seed = timer_seed,
+    .bytes = timer_bytes,
+    .cleanup = timer_cleanup,
+    .add = timer_add,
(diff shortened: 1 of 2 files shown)


lib/hcrypto/rand-timer.c
```diff
@@ -191,6 +192,16 @@ const RAND_METHOD hc_rand_timer_method = {
     timer_pseudorand,
     timer_status
 };
+#else
+const RAND_METHOD hc_rand_timer_method = {
```
DavidMulder: Both of these blocks have the same definition in them. Is this a typo?

kaduk: Yes, thanks for spotting it.

lha:
either we do c99, or we don't.

now that there is the clang fronting to vc++, do we need to keep pre-c99 for windows ? @jaltman

jaltman:
@lha We will not be converting to clang on Windows for the next release and it is unclear at this point what versions of Windows we can build for with the clang front-end.

jaltman:
This was merged by @nicowilliams as aa87e08cc72fdf708cfc353020a713970991a21b

PR #203, opened 2016-09-19
Fix typo in rand-fortuna.c

sdigit:
`__GNUC__` was written as `__GUNC__`, which could have interesting results as it'll never be defined.


adaugherity:
Commit aa87e08 had the typo in both lib/hcrypto/rand-fortuna.c and .../rand-timer.c.  This PR fixed rand-fortuna.c but rand-timer.c still has the `__GUNC__` typo.

PR #204, opened 2016-10-03
lib/krb5/fcache.c: return errors in fcc_destroy() rather than always 0.

elric1:


lib/krb5/fcache.c
```diff
@@ -277,6 +277,9 @@ _krb5_erase_file(krb5_context context, const char *filename)
     if (unlink(filename) < 0) {
 	_krb5_xunlock(context, fd);
         close (fd);
+	krb5_set_error_message(context, errno,
+	    N_("krb5_cc_destroy: unlinking \"%s\": %s", ""),
+	    filename, strerror(errno));
         return errno;
```
jaltman: This isn't safe.  The value of errno can be altered by the krb5_xunlock and the close.   In order to return "errno" must save the value before doing other things and then return the saved value.

vdukhovni:
+1

PR #208, opened 2016-11-11
Fix interop with 2008 R2 RODC (issue #173)

nicowilliams:
Based on patches by @jaltman.

### DIFF
--- a/kdc/kerberos5.c
+++ b/kdc/kerberos5.c
@@ -1544,7 +1544,8 @@ generate_pac(kdc_request_t r, Key *skey)
 krb5_boolean
 _kdc_is_anonymous(krb5_context context, krb5_principal principal)
 {
-    if (principal->name.name_type != KRB5_NT_WELLKNOWN ||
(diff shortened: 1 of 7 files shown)


jaltman:
@nicowilliams Thanks for the revised patches.

One of the items that was going to be worked on was to permit assignment of principal types and the associated parsing rules based upon matching component values and the number of components. 

I will review and test over the weekend.

nicowilliams:
Pushed:

22790e4 Test RODC interop fix
9e2b696 Make kdc name type strictness configurable
961f543 Set princ type to NT-SMTP-NAME when parsing
6a1db3f princ type NT-UNKNOWN + "host" == NT-SRV-HST
5aef50c gss-krb5: do_delegate remove dead comment
09bdb3a Set the right name type for anon princ (client)
020f2c7 kdc: principals of type NT-UNKNOWN can be anonymous
a59bb71 When building a princ name pick a sane def type

PR #217, opened 2016-12-03
Don't include build user, date and host in binaries.

jelmer:
This makes the build unreproducible.

### DIFF
--- a/include/bits.c
+++ b/include/bits.c
@@ -141,9 +141,6 @@ int main(int argc, char **argv)
 	}
 	f = fopen(argv[1], "w");
     }
-    fprintf(f, "/* %s -- this file was generated for %s by\n", fn, HOST);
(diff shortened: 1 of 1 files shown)


nicowilliams:
There's also `include/NTMakefile`.

The user/date/host info might be useful in some cases.  Maybe we should have a configuration option for it.

quanah:
It's fairly traditional for lots of software to include that information, and it can be extremely useful information to have.

nicowilliams:
@quanah Absolutely, however, for non-release builds it does make the build non-repeatable.  That's a problem.  We should want a `--release` or some such option to `./configure`, or perhaps a make variable.

nicowilliams:
I'll try and add a `--release` or `--enable-release` option to `./configure` before 7.1.

PR #218, opened 2016-12-03
Increase timeout when waiting for iprop-slave.

jelmer:
This is required when running on slower platforms. In Debian, we're hitting the timeout building on mips.

### DIFF
--- a/tests/kdc/wait-kdc.sh
+++ b/tests/kdc/wait-kdc.sh
@@ -36,7 +36,7 @@ log=${2:-messages.log}
 waitfor="${3:-${name} started}"
 
 t=0
-waitsec=35
+waitsec=65
 
 echo "Waiting for ${name} to start, looking logfile ${log}"
 


nicowilliams:
My goal is to make all the daemons daemonize properly, detaching only when ready.  In fact, they do, now (at least `kdc`, `digest-service`, `kcm`, `ipropd-slave`, `ipropd-master`, `kadmind`, and `kpasswdd`, but not `hpropd` nor any others).

The thing that's left to do is to make sure all the daemons have pidfiles or output their PID and then make the tests start them daemonized and stop them by killing their PIDs.

jelmer:
Checking pid files definitely seems like a more robust strategy than relying on timeouts.

jelmer:
It would be nice if the timeout could be increased in the meantime though, as this without the higher timeout this test fails on some hardware.

nicowilliams:
> Checking pid files definitely seems like a more robust strategy than
> relying on timeouts. 

The pidfile is only needed to kill the daemon.  The nice thing is that the daemon fork()s first, sets up things on the child side, and the parent doesn't exit until the child indicates that it's ready.  So all a script has to do is start the daemon and there's no need to wait for anything: as soon as the daemon exit(0)s (completes daemonizing), it's ready.

nicowilliams:
> It would be nice if the timeout could be increased in the meantime
> though, as this without the higher timeout this test fails on some
> hardware.

Sure.

nicowilliams:
I think the only thing to do to make the tests do the right thing is to make sure to set `HEIM_PIDFILE_DIR` in the tests.  However, we use `pidfile()`, which can be available from the OS (e.g., on NetBSD), in which case we may have no control over where the pidfile goes, and that's the reason (ISTR) I've not made this change yet...

Hmmm, nowadays on NetBSD `pidfile()` does take a path (if it isn't a basename): http://www.unix.com/man-page/netbsd/3/pidfile/.  But OpenBSD's still doesn't...  So I'm concerned about making this change.

One possibility would be to use `rk_pidfile()` instead of `pidfile()` and make sure we always build `rk_pidfile()`, and have `rk_pidfile()` check `$HEIM_PIDFILE_DIR`, and if not set then call the real `pidfile()` if we have it, else work as usual.

vdukhovni:
We should probably use (not override) the system pidfile(), however we could add a command-line
option that instead writes pids to an alternate location.  The system pidfile
directly may not even be writable by non-root users.

nicowilliams:
@vdukhovni It'd be better to use `rk_pidfile()` and have it writer wherever Heimdal wants to and also call the system `pidfile()` when there is one.

nicowilliams:
Ugh, libroken's `pidfile()` does not match the prototype of NetBSD's (nor OpenBSD's, which at least is the same as NetBSD's).  libroken's returns `void`, while the others return `int`.

nicowilliams:
BTW, I've patches to make the tests use `--detach` and pidfiles.  See the `misc` branch of my clone.  `check-iprop` still uses the `wait_kdc` script though, to wait for `ipropd-slave` to catch up to the master.  So we'll still need this patch...

nicowilliams:
Do please test-drive my `misc` changes.  I really should have done this when I added the `roken_detach_prep()` and `roken_detach_finish()` functions...

PR #219, opened 2016-12-03
Use actual libdir/includedir in heimdal-gssapi.pc.

jelmer:
This fixes the library path on e.g. systems like Debian which specify a custom --includedir and --libdir to configure.


nicowilliams:
Thanks.

Issue #246, opened 2017-01-24
hdb_generate_key_set_password signature changed in Heimdal 7.0


jaltman:
The signature of hdb_generate_key_set_password() was changed in Heimdal 7.0 by 

  commit af011f57fc4ae6e865bab471c20aa9047e4e19d4  
  Author: Roland C. Dowdeswell <elric@imrryr.org>
  Date:   Mon Nov 28 15:18:52 2011 +0000

    Provide server side kadm5_chpass_principal_3() with ks_tuple implementation.

    We enable kadm5_chpass_principal_3() in the server side of the
    library.  The client kadm5 library calls will still return the
    error KAMD5_KS_TUPLE_NO_SUPP.

This change broke compatibility with the OpenLDAP smbk5pwd module, http://www.openldap.org/devel/gitweb.cgi?p=openldap.git;a=blob;f=contrib/slapd-modules/smbk5pwd/smbk5pwd.c;h=6f917bcccc4ae7c367530c2fe7ca051561e8f310;hb=HEAD

Can the original signature be restored and export a new function with the additional parameter for the new kadm functionality?

elric1:
Are we interested in API or ABI backwards compatibility here?  The reason that I ask is that the new signature is compatible with the MIT Kerberos function of the same name and it might be an idea to preserve that at least on an API level.

jaltman:
I think we care about API and ABI backward compatibility because its a public function that is in use by external modules.   There are many functions in Heimdal and MIT that have divergent signatures.  Applications already have to conditionally test for Heimdal or MIT.   By changing the signature we have now forced applications to test for which version of Heimdal is in use and we provide no version information on which to base that decision.

nicowilliams:
On Tue, Jan 24, 2017 at 08:18:12AM -0800, Roland C. Dowdeswell wrote:
> Are we interested in API or ABI backwards compatibility here?  The
> reason that I ask is that the new signature is compatible with the MIT
> Kerberos function of the same name and it might be an idea to preserve
> that at least on an API level.

The interface that matters is the hdb one, not the kadm5 one.  We're
compatible with MIT as to the kadm5 one, but we changed an hdb one that
is depended on elsewhere.

We should restore the original hdb_generate_key_set_password() and add a
version that takes the additional arguments.

Nico
--

vdukhovni:
[...]

PR #247, opened 2017-01-26
Fix signature of hdb_generate_key_set_password()

elric1:
The change to the signature of hdb_generate_key_set_password() in
Heimdal 7.1 broke API/ABI compatibility with previous releases.  We
fix this by renaming it hdb_generate_key_set_password_with_ks_tuple()
and creating a new hdb_generate_key_set_password() which calls our
new function with zeroes for the added arguments.

Issue #246      https://github.com/heimdal/heimdal/issues/246


vdukhovni:
+1. Looks good.

PR #248, opened 2017-02-17
Add include/includedir directives for krb5.conf

nicowilliams:
Includes docs and one test.  includedir is not tested in-tree, but I tested it manually.

### DIFF
--- a/lib/krb5/config_file.c
+++ b/lib/krb5/config_file.c
@@ -41,6 +41,7 @@
 
 /* Gaah! I want a portable funopen */
 struct fileptr {
+    krb5_context context;
     const char *s;
     FILE *f;
 };
@@ -363,7 +364,7 @@ krb5_config_parse_debug (struct fileptr *f,
 	    ++p;
 	if (*p == '#' || *p == ';')
 	    continue;
-	if (*p == '[') {
+        if (*p == '[') {
 	    ret = parse_section(p, &s, res, err_message);
 	    if (ret)
 		return ret;
@@ -371,6 +372,22 @@ krb5_config_parse_debug (struct fileptr *f,
 	} else if (*p == '}') {
 	    *err_message = "unmatched }";
 	    return KRB5_CONFIG_BADFORMAT;
+        } else if (strncmp(p, "include", sizeof("include") - 1) == 0 &&
+            isspace(p[sizeof("include") - 1])) {
+            p += sizeof("include");
+            while (isspace(*p))
+                p++;
+            ret = krb5_config_parse_file_multi(f->context, p, res);
+	    if (ret)
+		return ret;
+        } else if (strncmp(p, "includedir", sizeof("includedir") - 1) == 0 &&
+            isspace(p[sizeof("includedir") - 1])) {
+            p += sizeof("includedir");
+            while (isspace(*p))
+                p++;
+            ret = krb5_config_parse_dir_multi(f->context, p, res);
+	    if (ret)
+		return ret;
 	} else if(*p != '\0') {
 	    if (s == NULL) {
 		*err_message = "binding before section";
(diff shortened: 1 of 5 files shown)


lib/krb5/config_file.c
```diff
@@ -397,6 +414,64 @@ is_plist_file(const char *fname)
...
+    if ((d = opendir(dname)) == NULL)
+        return errno;
+
+    while ((entry = readdir(d)) != NULL) {
+        char *p = entry->d_name;
+        char *path;
+        int is_valid = 1;
+
+        while (*p) {
+            if (!isalpha(*p) && *p != '_' && *p != '-' &&
```
vdukhovni: Please change isalpha to isalnum here.

lib/krb5/config_file.c
```diff
@@ -397,6 +414,64 @@ is_plist_file(const char *fname)
...
+            }
+            p++;
+        }
+        if (!is_valid)
+            continue;
+
+        if (asprintf(&path, "%s/%s", dname, entry->d_name) == -1 ||
+            path == NULL)
+            return krb5_enomem(context);
+        ret = krb5_config_parse_file_multi(context, path, res);
```
vdukhovni: Skip (sub)directories here by stat() or fstat before attempting to read as a config file...

lib/krb5/config_file.c
```diff
@@ -397,6 +414,64 @@ is_plist_file(const char *fname)
...
+                break;
+            }
+            p++;
+        }
+        if (!is_valid)
+            continue;
+
+        if (asprintf(&path, "%s/%s", dname, entry->d_name) == -1 ||
+            path == NULL)
+            return krb5_enomem(context);
```
vdukhovni: Missing closedir() here

lib/krb5/config_file.c
```diff
@@ -397,6 +414,64 @@ is_plist_file(const char *fname)
...
+        ret = krb5_config_parse_file_multi(context, path, res);
+        free(path);
+        if (ret == ENOMEM)
+            return krb5_enomem(context);;
```
vdukhovni: Ditto re closedir()

vdukhovni:
Closing in favour of #253

lib/krb5/config_file.c
```diff
@@ -363,14 +364,30 @@ krb5_config_parse_debug (struct fileptr *f,
...
 	    b = NULL;
 	} else if (*p == '}') {
 	    *err_message = "unmatched }";
 	    return KRB5_CONFIG_BADFORMAT;
+        } else if (strncmp(p, "include", sizeof("include") - 1) == 0 &&
+            isspace(p[sizeof("include") - 1])) {
+            p += sizeof("include");
+            while (isspace(*p))
+                p++;
+            ret = krb5_config_parse_file_multi(f->context, p, res);
```
vdukhovni: This code admits relative pathnames.  Should absolte pathnames be enforced?

nicowilliams:
@vdukhovni Regarding relative paths...  MIT's docs say that absolute paths must be used, but its implementation does not reject relative paths.  Portably determining whether a path is absolute is non-trivial...  @elric1 proposes that I check `#ifndef WIN32`, and not check on `WIN32`.  I... don't really care since one should really not include relative paths, but I'll probably do what @elric1 proposes.

jaltman:
`static int
is_absolute_path(const char *path)
{
    /*
     * An absolute path is one that refers to an explicit object
     * without ambiguity.
     */
#ifdef WIN32
    size_t len = strlen(path);

    /* UNC path is by definition absolute */
    if (len > 2
         && ISPATHSEP(path[0])
         && ISPATHSEP(path[1]))
        return 1;

    /* A drive letter path might be absolute */
    if (len > 3
         && isalpha(path[0])
         && path[1] == ':'
         && ISPATHSEP(path[2]))
        return 1;

    /*
     * if no drive letter but first char is a path
     * separator then the drive letter must be obtained
     * from the including file.
     */
#else
    /* UNIX is easy, first char '/' is absolute */
    if (ISPATHSEP(path[0]))
        return 1;
#endif
    return 0;
}`

nicowilliams:
@jaltman And what about UNCs?

Also, MSFT might introduce new such conventions some day, and, indeed, so can POSIX, since `//` is allowed to be special under POSIX (IIRC because of Apollo).

The only reliable, portable way I know of to determine if a path is absolute is to repeatedly apply `dirname()` (to a copy of the path) until its output is repeated, and then to check that the last output is a prefix of the original path.

vdukhovni:
We're not going to do the dirname thing.  The code Jeff looks just fine.  Or just support Unix with a leading "/".  We don't need perfection here.  Or leave it be with a comment in the code, and a warning in the docs.

nicowilliams:
I'm OK with leaving the code as-is.  I'm OK with including a check only on Unix (indeed, I've pushed that to my `master`).  I won't bother with the `dirname()` thing.  I could add the code that @jaltman proposes (perhaps we *never* want to include files from UNC paths anyways!).

nicowilliams:
What I won't do is add @jaltman's code to `lib/roken` -- I do feel like a proper `is_absolute_path()` general-purpose utility function needs to do the `dirname()` thing (or similar and equivalent).

(Also, re: rejecting UNCs here... why not also reject /net/... paths?  There could be a never-ending list of things to do / not do.)

vdukhovni:
We must not reject any absolute paths, however unsafe they might be.  All that's called for is that the path be unambiguous. If someone configures an unsafe remote path, that's their problem, not ours.  We can document that the paths SHOULD be unambiguous (absolute), or enforce it and document that they MUST be unambiguous.

nicowilliams:
@vdukhovni That was already the case yesterday...

jaltman:
@nicowilliams The code I provided handles UNC paths.   UNC paths are absolute by definition.

nicowilliams:
@jaltman Oh, yes it does.

jaltman:
is_absolute_name() does not belong in libroken.  libroken is for supplying otherwise portable functions that are missing or implemented incorrectly on some platforms.   heimbase is where an is_absolute_path() or similar function would be added.

nicowilliams:
`@!*(&*!`

There are other places with lame abs path checks that could use this function.  No good deed goes unpunished.

nicowilliams:
To make things worse, `master`'s Travis build fails tests...  I was wondering why my Travis builds were failing.  Of course, the Travis build isn't very informative on failure, so we should fix that too now.

vdukhovni:
I built master on MacOS Sierra, and "make check" passes.

nicowilliams:
@vdukhovni Yeah, Travis "helpfully" shows the recent-most PR not `master`.  So never mind that.

jaltman:
I'm building master on Fedora 25 with --enable-ldap=/usr.   It would be useful to indicate which configure options were specified when indicating that "make check" succeeds or fails.

vdukhovni:
I would have expected "--with-openldap=/usr", does the above work as intended?

nicowilliams:
@jaltman Never mind.  The problem is in my branch.  Travis was showing me the result of the latest PR, now `heimdal/heimdal`'s `master`, even though I was on the page for `heimdal/heimdal` -- very annoying.  The indication that it wasn't `master` was subtle.

nicowilliams:
Plus is passes in my clone, so I'm pushing (to my `master`) an update to cat the test-suite.log files on failure in Travis.

PR #253, opened 2017-03-09
Fix CR comments on include/includedir

nicowilliams:


lib/krb5/config_file.c
```diff
@@ -553,8 +563,7 @@ krb5_config_parse_file_multi (krb5_context context,
 	if (ret) {
 	    krb5_set_error_message(context, ret,
 				   "Failed to parse plist %s", fname);
-	    if (newfname)
-		free(newfname);
+            free(newfname);
 	    return ret;
 	}
 #else
```
vdukhovni: The `#else` codepath is, I think, missing a decrement of `context->config_include_depth`

jaltman: agreed.  introduced in fe43be8558

nicowilliams:
@jaltman There are things in lib/roken that don't belong there if `is_absolute_path()` doesn't belong there either:

 - `parse_*.c`
 - `cloexec.c`
 - `environment.c`
 - `mini_inetd.c`
 - `net_*.c`
 - `simple_exec*.c`
 - `detach.*`
 - `swab.c`
 - `ct.c`
 - `copyhostent.c`
 - `getarg.*`

Meanwhile, `lib/base` does have some portability things in it that really belong in `lib/roken`:

 - everything to do with threads

and otherwise consists mainly of an implementation of JSON-ish things.

If we were extracting these things into separate projects, I think we should:

 - put all portability things (including `lib/base`'s threading things) in `lib/roken`
     - this includes `is_absolute_path()`
 - leave only the JSON-ish things in `lib/base` and give it a more appropriate name

nicowilliams:
I surmise that my Travis-CI build must be failing because... either `$srcdir` or `$objdir` are not absolute...

nicowilliams:
Yepper doodles, that's the issue.  From the build output, where I replaced the rejection of relative paths with an `fprintf(stderr, ...)`:

```
WELL THERE MATEY, ../../../tests/gss/include-krb5.conf is NOT an absolute path
```

vdukhovni:
Sweet. For the test, this'll have to converted to an absolute path for the "include" directive...
Good test!

nicowilliams:
My fix for this is simply to canonicalize `$srcdir` in `configure.ac`.  Testing a Travis build...

PR #270, opened 2017-03-31
Further improvements to issuid()

nicowilliams:
This set of changes will make issuid() much more likely to return the correct answer on:

 - Linux with old glibc
 - FreeBSD
 - NetBSD

and should continue to return the correct answer on:

 - Linux with newer glibc
 - Illumos
 - OpenBSD

### DIFF
--- a/lib/roken/issuid.c
+++ b/lib/roken/issuid.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1998 - 2001 Kungliga Tekniska Högskolan
+ * Copyright (c) 1998 - 2017 Kungliga Tekniska Högskolan
  * (Royal Institute of Technology, Stockholm, Sweden).
  * All rights reserved.
  *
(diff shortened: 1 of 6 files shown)


nicowilliams:
We have asked MIT to adopt Heimdal's `issuid()` in RT#8566.  @greghudson wants `issuid()` to be a bit smarter, and he's quite right.  This PR does that.

lib/roken/issuid.c
```diff
@@ -62,23 +62,84 @@
...
+#ifdef HAVE_GETAUXVAL
 #ifdef GETAUXVAL_SETS_ERRNO
     return getauxval(type);
 #else
-    unsigned long ret = getauxval(type);
+    unsigned long ret;
+    unsigned long ret2;
+    static int getauxval_sets_errno = -1;
+    int save_errno = errno;
+    int fd;
```
vdukhovni: Unused `fd`?

lib/roken/issuid.c
```diff
@@ -62,23 +62,84 @@
...
-    return ret;
+    errno = 0;
+    ret = getauxval(type);
+    if (ret != 0 || errno == ENOENT || getauxval_sets_errno == 1) {
+        if (ret != 0)
+            errno = save_errno;
+        return ret;
+    }
+
+    if (!getauxval_sets_errno) {
```
vdukhovni: I think that here `== 0` is more clear than `!`

lib/roken/issuid.c
nicowilliams: Correct.  I need to find a way to make this compile always so I can test it, no?

lib/roken/issuid.c
```diff
@@ -118,98 +179,220 @@ issuid(void)
...
+#endif
+
+    /*
+     * Fall through if we have rk_getauxval() but we didn't have (or
+     * don't know if we don't have) the aux entries that we needed.
+     * We're done with it.
+     */
+
+#if defined(HAVE_ISSETUGID)
+    if (issetugid()) {
```
vdukhovni: This has already been done, so the return value is surely non-zero at this point...

nicowilliams: Yes.  I can return unconditionally here, but only if I add a `#else` to avoid dead code warnings... though... if the compiler is smart enough it would emit them anyways, so, ok, I'll add the `#else`.

lib/roken/issuid.c
nicowilliams: Do you want me to make a change here?

lib/roken/issuid.c
vdukhovni: Not necessary at present, just drop the unused variable.

lib/roken/issuid.c
vdukhovni: Yes `if (getauxval_sets_errno == 0)`

lib/roken/issuid.c
vdukhovni: I see your point. Perhaps leave as-is given that we memoize the result...

nicowilliams:
I believe this code will work even on Illumos, but `rk_getprocauxval()` will not, as the entries in `/proc/self/auxv` are different there.  However, it would be nice to have `rk_getprocauxval()` and `rk_getauxval()` working on Illumos too.

PR #272, opened 2017-04-10
No tgt aliasing

vdukhovni:


nicowilliams:
@jaltman Viktor and I basically authored this together after much debugging.

These commits fix a bug observed only with patches we have that are not in `master` where for services in certain realms we only forward TGTs of the form `krbtgt/<service-realm>@<service-realm>`.  The breakage occurred when using referrals (i.e., no capaths on the client) and one realm in the middle, so `client@A` transiting `B` to talk to `host/foo@C`.  This can't be exercised in `master` today, but you can see the problems:

1. `krb5_build_ap_req()` does something it *really* ought not to do, and has always done it this way.
2. `store_cred()` does something it shouldn't do too: it was storing two ccache entries when the ticket we're storing is a TGT and was the result of a referral.  We never want anything like referral entries for TGTs in the ccache.
3. Lastly, and unrelatedly, but still obviously wrong (I think!  I could be wrong) `get_cred_kdc_referral()` was comparing the principal of the ticket in a TGS-REP to the requested one without comparing the realm name.  When this fails we were then (and still are) immediately checking that the returned ticket is a TGT.

Either (1) or (2) are sufficient to make this failure case go away, while (3) is completely harmless whether applied or not, and not really necessary in this case, but we think our patch for (3) is correct.

lib/krb5/build_ap_req.c
```diff
@@ -52,14 +52,8 @@ krb5_build_ap_req (krb5_context context,
   ap.ap_options.use_session_key = (ap_options & AP_OPTS_USE_SESSION_KEY) > 0;
   ap.ap_options.mutual_required = (ap_options & AP_OPTS_MUTUAL_REQUIRED) > 0;
 
-  ap.ticket.tkt_vno = 5;
-  copy_Realm(&cred->server->realm, &ap.ticket.realm);
-  copy_PrincipalName(&cred->server->name, &ap.ticket.sname);
-
   decode_Ticket(cred->ticket.data, cred->ticket.length, &t, &len);
```
nicowilliams: Let's get rid of the `t` local variable and decode directly into `&ap.ticket`.  I checked, and that's [what MIT does](https://github.com/krb5/krb5/blob/master/src/lib/krb5/krb/mk_req_ext.c#L142).

lib/krb5/get_cred.c
```diff
@@ -1051,9 +1051,7 @@ get_cred_kdc_referral(krb5_context context,
 	}
 
 	/* Did we get the right ticket ? */
-	if (krb5_principal_compare_any_realm(context,
-					     referral.server,
-					     ticket.server))
+	if (krb5_principal_compare(context, referral.server, ticket.server))
```
nicowilliams: FYI, MIT does allow completely different ticket sname@srealm when canonicalization is requested, but IIRC this requires FAST and other functionality we don't have in Heimdal.

vdukhovni:
Addressed comments and added commit for better handling of referrals for (remote) TGTs

lib/krb5/get_cred.c
```diff
@@ -1011,14 +1012,32 @@ get_cred_kdc_referral(krb5_context context,
...
+     * point will be krbtgt/B@START_REALM.  Only once we obtain a ticket for
+     * krbtgt/B@some-transit, do we switch to requesting krbtgt/A@B on our
+     * final request.
+     */
     referral = *in_creds;
-    ret = krb5_copy_principal(context, in_creds->server, &referral.server);
+    if (!want_tgt)
+        ret = krb5_copy_principal(context, in_creds->server, &referral.server);
+    else
+	ret = krb5_make_principal(context, &referral.server, start_realm, KRB5_TGS_NAME,
```
nicowilliams: Whitespace issue?

lib/krb5/get_cred.c
```diff
@@ -963,6 +963,7 @@ get_cred_kdc_referral(krb5_context context,
     krb5_creds **referral_tgts = NULL;  /* used for loop detection */
     int loop = 0;
     int ok_as_delegate = 1;
+    int want_tgt = krb5_principal_is_krbtgt(context, in_creds->server);
```
nicowilliams: This needs to be true IFF `krb5_principal_is_krbtgt(context, in_creds->server)` AND `in_creds->server->realm` is not the "referral realm", i.e., the empty string.

nicowilliams: The reason for this is that we want to be able to use `kgetcred(1)` to get a `krbtgt/B@` using referrals.

It's only when we want `krbtgt/B@A` that we have to start by asking for a `krbtgt/A`.

nicowilliams: Also, `want_tgt` is kind of a misnomer.  What it really is is "want no referrals on the last hop", but that's kinda wordy!

nicowilliams: All that said, I must say that this is a very nice patch!

nicowilliams: We'll leave testing for later, as we need to improve `klist(1)` to print the Ticket `sname`@`srealm` when it's not the same as the `cred->server` for each ccache entry.

vdukhovni:
Rebased

PR #274, opened 2017-04-12
Add secure_getenv()

nicowilliams:
And other stuff.  And use it.

### DIFF
--- a/appl/su/su.c
+++ b/appl/su/su.c
@@ -364,7 +364,7 @@ main(int argc, char **argv)
     else
 	su_user = argv[optidx++];
 
-    if (!issuid() && getuid() != 0)
+    if (geteuid() != 0)
 	warnx("Not setuid and you are not root, expect this to fail");
 
     pwd = k_getpwnam(su_user);
--- a/lib/hx509/softp11.c
+++ b/lib/hx509/softp11.c
@@ -818,18 +818,16 @@ func_not_supported(void)
 static char *
 get_config_file_for_user(void)
 {
-    char *fn = NULL;
+    char *fn;
 
 #ifndef _WIN32
-    char *home = NULL;
+    char *home;
     int ret;
 
-    if (!issuid()) {
-        fn = getenv("SOFTPKCS11RC");
-        if (fn)
-            fn = strdup(fn);
-        home = getenv("HOME");
-    }
+    fn = secure_getenv("SOFTPKCS11RC");
+    if (fn)
+        fn = strdup(fn);
+    home = secure_getenv("HOME");
     if (fn == NULL && home == NULL) {
         struct passwd *pw = getpwuid(getuid());
         if(pw != NULL)
--- a/lib/krb5/cache.c
+++ b/lib/krb5/cache.c
@@ -443,10 +443,7 @@ environment_changed(krb5_context context)
 	 strncmp(context->default_cc_name, "API:", 4) == 0))
 	return 1;
 
-    if(issuid())
-	return 0;
-
-    e = getenv("KRB5CCNAME");
+    e = secure_getenv("KRB5CCNAME");
     if (e == NULL) {
 	if (context->default_cc_name_env) {
 	    free(context->default_cc_name_env);
@@ -518,15 +515,13 @@ krb5_cc_set_default_name(krb5_context context, const char *name)
     if (name == NULL) {
 	const char *e = NULL;
 
-	if (!issuid()) {
-	    e = getenv("KRB5CCNAME");
-	    if (e) {
-		p = strdup(e);
-		if (context->default_cc_name_env)
-		    free(context->default_cc_name_env);
-		context->default_cc_name_env = strdup(e);
-	    }
-	}
+        e = secure_getenv("KRB5CCNAME");
+        if (e) {
+            p = strdup(e);
+            if (context->default_cc_name_env)
+                free(context->default_cc_name_env);
+            context->default_cc_name_env = strdup(e);
+        }
 
 #ifdef _WIN32
 	if (p == NULL) {
--- a/lib/roken/getauxval.c
+++ b/lib/roken/getauxval.c
@@ -1,0 +1,254 @@
+/*
+ * Copyright (c) 2016 - 2017 Kungliga Tekniska Högskolan
+ * (Royal Institute of Technology, Stockholm, Sweden).
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * 3. Neither the name of the Institute nor the names of its contributors
+ *    may be used to endorse or promote products derived from this software
+ *    without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE INSTITUTE AND CONTRIBUTORS ``AS IS'' AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED.  IN NO EVENT SHALL THE INSTITUTE OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+ * OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ */
+
+#include <config.h>
+
+#ifdef HAVE_SYS_AUXV_H
+#include <sys/auxv.h>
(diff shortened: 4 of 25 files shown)


nicowilliams:
This needs a test.  And since we can't test set-uid in-tree, it probably needs some form of injection, else a manual test.

appl/su/su.c
```diff
@@ -364,7 +364,7 @@ main(int argc, char **argv)
     else
 	su_user = argv[optidx++];
 
-    if (!issuid() && getuid() != 0)
+    if (!geteuid() != 0)
```
vdukhovni: Drop the leading "!"

lib/roken/getauxval.c
```diff
@@ -0,0 +1,248 @@
...
+    if ((fd = open("/proc/self/auxv", O_RDONLY)) == -1) {
+        if (errno == ENOENT)
+            has_proc_auxv = 0;
+        ret = errno;
+        errno = save_errno;
+        return ret;
+    }
+
+    got_proc_auxv = 0;
+    memset(auxv, 0, sizeof(auxv));
```
nicowilliams: This, the memset(), and the read() are not thread-safe.  Making this thread-safe is non-trivial: because the `HEIM_ONCE` stuff is in `lib/base`.

At some point we should really move all of the threading primitives from `lib/base` into `lib/roken`!

appl/su/su.c
nicowilliams: Oops!  Nice catch!

lib/krb5/cache.c
```diff
@@ -443,10 +443,7 @@ environment_changed(krb5_context context)
 	 strncmp(context->default_cc_name, "API:", 4) == 0))
 	return 1;
 
-    if(issuid())
-	return 0;
-
-    e = getenv("KRB5CCNAME");
+    e = secure_getenv("KRB5CCNAME");
```
vdukhovni: This appears appears to be a bit of a change when suid, since (e == NULL) does not always just `return 0`.  Is that OK?

nicowilliams: Yes, actually.  I pondered that for a bit myself.  `context->default_cc_name_env` basically gets the value of `secure_getenv("KRB5CCNAME")`, so if `issuid()` is true then `secure_getenv("KRB5CCNAME")` is `NULL` and `context->default_cc_name_env` is too, so we still `return 0` here.

lib/hx509/softp11.c
```diff
@@ -824,12 +824,10 @@ get_config_file_for_user(void)
     char *home = NULL;
```
vdukhovni: Arguably, this no longer needs initialization...

lib/roken/getauxval.c
```diff
@@ -0,0 +1,308 @@
...
+#if defined(ENABLE_PTHREAD_SUPPORT) && defined(HAVE_PTHREAD_H)
+#include <pthread.h>
+#endif
+
+#include <errno.h>
+
+#include "roken.h"
+#include "getauxval.h"
+
+int injected_auxv = 0; /* shared with issuid() for testing */
```
vdukhovni: As a global, perhaps `injected_auxv` should have a more Heimdal-specific name? Would `heim_injected_auxv` or perhaps `rk_injected_auxv` be better?  While we don't export the symbol, it will be visible with static linking and may conflict with imported symbols...  The odds of any problems are of course rather low.

lib/roken/getauxval.c
```diff
@@ -0,0 +1,308 @@
...
+static int has_proc_auxv = 1;
+static int proc_auxv_ret = 0;
+
+#if defined(ENABLE_PTHREAD_SUPPORT) && defined(HAVE_PTHREAD_H)
+pthread_once_t readprocauxv_once = PTHREAD_ONCE_INIT;
+#endif
+
+/* There's no standard maximum; 64 is PLENTY */
+#define MAX_AUXV_COUNT 64
+static auxv_t auxv[MAX_AUXV_COUNT];
```
vdukhovni: Famous last words: `64 is PLENTY`...

lib/roken/getauxval.c
```diff
@@ -0,0 +1,308 @@
...
+            has_proc_auxv = 0;
+        goto out;
+    }
+
+    do {
+        if ((bytes = read(fd, p, sz)) > 0) {
+            sz -= bytes;
+            p += bytes;
+        }
+    } while (sz && ((bytes == -1 && errno == EINTR) || bytes > 0));
```
vdukhovni: This should error out somehow if we we've not consumed the entire auxv.  Also, do we need any special code for string-valued elements?

lib/roken/getauxval.c
```diff
@@ -0,0 +1,308 @@
...
+ *
+ * @return The value of the ELF auxiliary value for the given type, or
+ * zero and sets errno to ENOENT.
+ */
+ROKEN_LIB_FUNCTION unsigned long ROKEN_LIB_CALL
+rk_getauxval(unsigned long type)
+{
+#ifdef HAVE_GETAUXVAL
+#ifdef GETAUXVAL_SETS_ERRNO
+    if (injected_auxv)
```
vdukhovni: How does injected_auxv affect `rk_getprocauxval()`?  Why are we calling that function in this case?

lib/roken/getauxval.c
```diff
@@ -0,0 +1,308 @@
...
+        return rk_getprocauxval(type);
+    return getauxval(type);
+#else
+    unsigned long ret;
+    unsigned long ret2;
+    static int getauxval_sets_errno = -1;
+    int save_errno = errno;
+
+    if (injected_auxv)
```
vdukhovni: Ditto.

lib/roken/getauxval.c
```diff
@@ -0,0 +1,308 @@
...
+/**
+ * Like the nearly-standard getauxval().  If the auxval is not found
+ * returns NULL and always sets errno to ENOENT.  Otherwise if auxval is
+ * found it leaves errno as it was, even if the value is zero.
+ *
+ * @return The value of the ELF auxiliary value for the given type, or
+ * NULL and sets errno to ENOENT.
+ */
+ROKEN_LIB_FUNCTION void * ROKEN_LIB_CALL
+rk_getauxptr(unsigned long type)
```
vdukhovni: Does this yield a valid pointer when reading /proc/self/auxv?  If not, how do we avoid problems?

lib/roken/getauxval.c
```diff
@@ -0,0 +1,308 @@
...
+/**
+ * Like the nearly-standard getauxval().  If the auxval is not found
+ * returns NULL and always sets errno to ENOENT.  Otherwise if auxval is
+ * found it leaves errno as it was, even if the value is zero.
+ *
+ * @return The value of the ELF auxiliary value for the given type, or
+ * NULL and sets errno to ENOENT.
+ */
+ROKEN_LIB_FUNCTION void (*
+rk_getauxfnc(unsigned long type))(void) ROKEN_LIB_CALL
```
vdukhovni: And how about this one and /proc?

lib/roken/getauxval.c
nicowilliams: It's not exported...

lib/hx509/softp11.c
nicowilliams: Oh sure.

lib/roken/getauxval.c
nicowilliams: Heheh, I know.  The factor of ten was purposeful :)

The thing is though that we'll discover when 64 is no longer enough.

Oh, and Linux nowadays has around 18 of these, so maybe 64 is a bit low.

lib/roken/getauxval.c
nicowilliams: Errors are handled: by storing `errno` in `proc_auxv_ret`, which is checked in `readprocauxv()`.

As for string-valued auxv elements: there is no such thing.  They're all `uintptr_t`s that are either interpreted as a data pointer, a function pointer, or an unsigned integer.  Some of the pointed-to things will be in the stack after the auxv terminator, much as with the argument and environment vectors, but they're not _in_ the vector.

lib/roken/getauxval.c
nicowilliams: So, we try to use the C library's `getauxval()` where possible, but if we've injected auxv entries for testing `issuid()` and `secure_getenv()`, then we can't use the C library `getauxval()`, and we have to call `rk_getprocauxval()`.

lib/roken/getauxval.c
nicowilliams: Same answer.

lib/roken/getauxval.c
nicowilliams: Yes, it does!  And it gets tested, since the test does check that the C library `getauxval()` and `rk_getprocauxval()` return the same answers.

lib/roken/getauxval.c
nicowilliams: Same answer.

lib/roken/getauxval.c
nicowilliams: Oh, and the reasons this code tries the true C library's `getauxval()` are:

 - `getauxval()` seems preferable to use the official interface where it works, no?
 - `getauxval()` will be faster, since it does not have to read `/proc/self/auxv`, since the RTLD should save the pointer to the auxv where the C library can get at it

lib/roken/getauxval.c
```diff
@@ -0,0 +1,308 @@
...
+{
+    size_t i;
+    int ret;
+
+    /*
+     * This function is racy, but as an internal function never meant to
+     * be called in a threaded program, we don't care.
+     */
+
+    if (e->a_type == 0) {
```
nicowilliams: This is about truncating the auxv to be empty.  But we don't need nor use this.  And anyways, this should follow the call to `readprocauxv()`.

nicowilliams: So I'll remove this.

lib/roken/getauxval.c
nicowilliams: @vdukhovni Points out that in a static link archive case this could collide with a symbol from a library.  Of course, that's just *always* a risk with static linking, but sure, I shall add an `rk_` prefix to it.

lib/roken/secure_getenv.c
```diff
@@ -0,0 +1,51 @@
...
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ */
+
+#include <config.h>
+
+#ifdef HAVE_SYS_AUXV_H
+#include <sys/auxv.h>
```
vdukhovni: Final review pass: I don't think we need this include here.

nicowilliams:
Thanks!

PR #275, opened 2017-04-18
Fix check-iprop race

vdukhovni:
Creation of a random key HDB entry takes 3 updates, not one.  Instead of
hard-coding this magic count, wait for the appearance of the expected
entry with no "invalid" flag among its attributes.


nicowilliams:
Thanks!

PR #285, opened 2017-05-20
Fix memory leak in _gss_ntlm_release_cred.

alexis-m:
ntlm_cred is always allocated with calloc, so we need to free the cred
object too, similarly to what _gsskrb5_release_cred does.


nicowilliams:
Thanks!

PR #286, opened 2017-05-25
Improve ccache and keytab performance

nicowilliams:


lib/krb5/store_stdio.c
```diff
@@ -0,0 +1,231 @@
...
+
+    /* similar pattern to net_write() to support pipes */
+    while (rem > 0) {
+	count = fwrite(cbuf, 1, rem, F(sp));
+	if (count < 0) {
+	    if (errno == EINTR)
+		continue;
+	    else
+		return count;
+	}
```
vdukhovni: I don't see the `` else if (size == rem) `` pattern here...  With partial writes, should this not return the number of bytes written?  Or are partial writes always fatal at this layer?

lib/krb5/store_stdio.c
```diff
@@ -0,0 +1,231 @@
...
+    errno = save_errno;
+    return ftello(F(sp));
+}
+
+static int
+stdio_trunc(krb5_storage * sp, off_t offset)
+{
+    int save_errno = errno;
+
+    if (fflush(F(sp)) == EOF)
```
vdukhovni: Perhaps `!= 0` rather than `== EOF`?

lib/krb5/store_stdio.c
```diff
@@ -0,0 +1,231 @@
...
+    if (fseeko(F(sp), offset, SEEK_SET) == -1)
+        return errno;
+    errno = save_errno;
+    return 0;
+}
+
+static int
+stdio_sync(krb5_storage * sp)
+{
+    if (fflush(F(sp)) == EOF)
```
vdukhovni: Ditto I guess...

lib/krb5/store_stdio.c
```diff
@@ -0,0 +1,231 @@
...
+	return errno;
+    return 0;
+}
+
+static void
+stdio_free(krb5_storage * sp)
+{
+    int save_errno = errno;
+    if (fclose(F(sp)) == 0)
+        errno = save_errno;
```
vdukhovni: This closes the underlying file!  But IIRC code in the keytab and fcc layers both frees the sp and and also closes the underlying fd again?  Am I missing something?

lib/krb5/store_stdio.c
```diff
@@ -0,0 +1,231 @@
...
+     * This function used to try to pass the input to
+     * _get_osfhandle() to test if the value is a HANDLE
+     * but this doesn't work because doing so throws an
+     * exception that will result in Watson being triggered
+     * to file a Windows Error Report.
+     */
+    fd = _dup(fd_in);
+#else
+    fd = dup(fd_in);
+#endif
```
vdukhovni: I see a dup!

lib/krb5/store_stdio.c
```diff
@@ -0,0 +1,231 @@
...
+
+    if (fd < 0)
+        return NULL;
+
+    if ((fl & O_RDWR))
+        mode = "r+";
+    else if ((fl & O_WRONLY))
+        mode = "w";
```
vdukhovni: The O_RDONLY, O_WRONLY and O_RDWR flags are not required to be independent bits!  Instead switch on (fl & O_ACCMODE)

http://pubs.opengroup.org/onlinepubs/009695399/functions/fcntl.html

lib/krb5/store_stdio.c
```diff
@@ -0,0 +1,231 @@
...
+{
+    if (fflush(F(sp)) == EOF)
+	return errno;
+    if (fsync(fileno(F(sp))) == -1)
+	return errno;
+    return 0;
+}
+
+static void
+stdio_free(krb5_storage * sp)
```
vdukhovni: Who frees the sp itself?  Is that the generic sp layer?

lib/krb5/store_stdio.c
nicowilliams: They're not fatal in other `lib/krb5/store_*.c` `*_fetch()` implementations!  They support partial reads.

lib/krb5/store_stdio.c
nicowilliams: What's the difference?

lib/krb5/store_stdio.c
nicowilliams: That's the idea.  I should probably `NULL` the `f` field and guard the `fclose()`, but otherwise this is exactly what's intended.

lib/krb5/store_stdio.c
nicowilliams: Yes: because `fdopen()` doesn't dup.

lib/hx509/libhx509-exports.def
```diff
@@ -30,6 +30,7 @@ EXPORTS
 	hx509_request_set_name
 	_hx509_request_to_pkcs10
 	_hx509_request_to_pkcs10
+	_hx509_sel_yywrap
```
vdukhovni: Do we still need this export?

lib/kadm5/log.c
```diff
@@ -1206,9 +1210,9 @@ kadm5_log_rename(kadm5_server_context *context,
         ret = krb5_store_principal(sp, source);
     if (ret == 0) {
         errno = 0;
-        if (krb5_storage_write(sp, value.data, value.length) !=
-            (krb5_ssize_t)value.length)
-            ret = errno ? errno : EIO;
+        bytes = krb5_storage_write(sp, value.data, value.length);
+        if (bytes != (krb5_ssize_t)value.length)
+            ret = errno ? errno : ENOMEM;
```
vdukhovni: Typo here, the test condition is surely `bytes == -1`

lib/krb5/keytab_file.c
```diff
@@ -72,18 +74,20 @@ krb5_kt_ret_string(krb5_context context,
...
     *data = malloc(size + 1);
     if (*data == NULL)
 	return krb5_enomem(context);
-    ret = krb5_storage_read(sp, *data, size);
+    bytes = krb5_storage_read(sp, *data, size);
     (*data)[size] = '\0';
-    if(ret != size)
-	return (ret < 0)? errno : KRB5_KT_END;
+    if (bytes != size)
+	return (bytes < 0)? errno : KRB5_KT_END;
```
vdukhovni: Same.

lib/krb5/keytab_file.c
```diff
@@ -92,13 +96,15 @@ krb5_kt_store_data(krb5_context context,
...
-    ret = krb5_storage_write(sp, data.data, data.length);
-    if(ret != (int)data.length){
-	if(ret < 0)
+    if (ret != 0)
+        return ret;
+    bytes = krb5_storage_write(sp, data.data, data.length);
+    if (bytes != (int)data.length){
+	if (bytes < 0)
 	    return errno;
 	return KRB5_KT_END;
```
vdukhovni: Use common idiom?

lib/krb5/keytab_file.c
```diff
@@ -109,14 +115,16 @@ static krb5_error_code
...
+    if (ret != 0)
 	return ret;
-    ret = krb5_storage_write(sp, data, len);
-    if(ret != (int)len){
-	if(ret < 0)
+    bytes = krb5_storage_write(sp, data, len);
+    if (bytes != (int)len){
+	if(bytes < 0)
 	    return errno;
 	return KRB5_KT_END;
```
vdukhovni: Same.

lib/krb5/keytab_file.c
```diff
@@ -382,32 +391,34 @@ fkt_start_seq_get_int(krb5_context context,
 	close(c->fd);
 	return ret;
     }
-    c->sp = krb5_storage_from_fd(c->fd);
+    if ((flags & O_ACCMODE) == O_RDWR && (flags & O_APPEND))
+        stdio_mode = "ab+";
+    else if ((flags & O_ACCMODE) == O_RDWR)
+        stdio_mode = "rb+";
+    else if ((flags & O_ACCMODE) == O_WRONLY)
+        stdio_mode = "wb+";
```
vdukhovni: Perhaps just "wb" for this one?  We don't have read access...

lib/krb5/keytab_file.c
```diff
@@ -733,11 +762,20 @@ fkt_remove_entry(krb5_context context,
...
 	    memset(buf, 0, sizeof(buf));
 	    while(len > 0) {
-		krb5_storage_write(cursor.sp, buf,
+		bytes = krb5_storage_write(cursor.sp, buf,
 		    min((size_t)len, sizeof(buf)));
+                if (bytes != min((size_t)len, sizeof(buf))) {
+                    if (bytes == -1)
+                        ret = errno;
+                    else
+                        ret = KRB5_KT_END;
```
vdukhovni: Ternary idiom?

lib/krb5/store_stdio.c
```diff
@@ -0,0 +1,227 @@
...
+    /* similar pattern to net_read() to support pipes */
+    while (rem > 0) {
+	count = fread(cbuf, 1, rem, F(sp));
+	if (count < 0) {
+	    if (errno == EINTR)
+		continue;
+	    else
+		return count;
+	} else if (count == 0) {
+	    return count;
```
vdukhovni: So we return 0 when we could not read all the requested data. Worth a comment?  With stdio partial reads make more sense than partial writes, so I might be tempted to return the partial count here (on the read side).

lib/krb5/store_stdio.c
```diff
@@ -0,0 +1,227 @@
...
+stdio_store(krb5_storage * sp, const void *data, size_t size)
+{
+    const char *cbuf = (const char *)data;
+    ssize_t count;
+    size_t rem = size;
+
+    /* similar pattern to net_write() to support pipes */
+    while (rem > 0) {
+	count = fwrite(cbuf, 1, rem, F(sp));
+	if (count < 0) {
```
vdukhovni: Should we do anything with (count == 0)?

lib/krb5/store_stdio.c
nicowilliams: Yes.

lib/hx509/libhx509-exports.def
nicowilliams: Probably not.  I'll see about removing it.

lib/kadm5/log.c
nicowilliams: No, it's not.  However, the `ret` assignment needs to change to `ret = bytes == -1 ? errno : ENOMEM;`.

lib/krb5/keytab_file.c
```diff
@@ -52,18 +52,20 @@ krb5_kt_ret_data(krb5_context context,
...
     data->length = size;
     data->data = malloc(size);
     if (data->data == NULL)
 	return krb5_enomem(context);
-    ret = krb5_storage_read(sp, data->data, size);
-    if(ret != size)
-	return (ret < 0)? errno : KRB5_KT_END;
+    bytes = krb5_storage_read(sp, data->data, size);
+    if (bytes != size)
+	return (bytes < 0) ? errno : KRB5_KT_END;
```
nicowilliams: I don't have an opinion on this one.

lib/krb5/keytab_file.c
nicowilliams: Sure.  I went for the minimal, easiest to review change.

lib/krb5/keytab_file.c
nicowilliams: Ops, yes, sorry.  Good catch.

lib/krb5/keytab_file.c
nicowilliams: Yes, will change it.

lib/krb5/store_stdio.c
nicowilliams: Oh, no, that should be `return size - rem;`.

lib/krb5/store_stdio.c
nicowilliams: Yes, we should: to avoid an infinite loop if the fd is `O_NONBLOCK`, right.  What to return?  `-1`?

lib/krb5/keytab_file.c
vdukhovni: I don't either, but I think we should use the same test everywhere.

lib/krb5/store_stdio.c
vdukhovni: I think that -1 is fine.

lib/hx509/libhx509-exports.def
nicowilliams: Removing it is ok, yes.

nicowilliams:
Thx!

PR #296, opened 2017-06-19
Fixes issue#294

quanah:
Fixes issue#294 when building Heimdal on x86 Solaris 11.


quanah:
going to close and regenerate from its own unique branch so it doesn't pick up the other cruft.

Issue #308, opened 2017-08-03
Solaris 10: Build fails in lib/libedit/src/chared.c


quanah:
Solaris 10 builds currently fail when attempting to compile lib/libedit/src/chared

Adding the following to lib/libedit/src/sys.h allows it to move past this build failure:

```
#ifdef HAVE_SYS_TYPES_H
#include <sys/types.h>
#endif
```

I will do a pull request shortly (Working on other solaris 10 build failures as well)

PR #311, opened 2017-08-08
Fixes https://github.com/heimdal/heimdal/issues/308

quanah:
Only include sys/types.h if we actually have it.


nicowilliams:
Thanks!

Issue #310, opened 2017-08-08
Solaris 10: kdc/connect.c uses PF_LOCAL instead of PF_UNIX


quanah:
From https://www.gnu.org/software/libc/manual/html_node/Address-Formats.html

```
AF_UNIX

    This is a synonym for AF_LOCAL. Although AF_LOCAL is mandated by POSIX.1g, AF_UNIX is portable to more systems. AF_UNIX was the traditional name stemming from BSD, so even most POSIX systems support it. It is also the name of choice in the Unix98 specification. (The same is true for PF_UNIX vs. PF_LOCAL).
```

PR #312, opened 2017-08-08
Fixes https://github.com/heimdal/heimdal/issues/310

quanah:
Use PF_UNIX instead of PF_LOCAL


nicowilliams:
Does Solaris 10 not have `PF_LOCAL`??

quanah:
@nicowilliams Correct, Solaris 10 does not have PF_LOCAL available.

nicowilliams:
Thanks!

PR #313, opened 2017-08-08
Issue309

quanah:
This adds a new getline routine, and adds it to the compile suite.  Note that I have not currently tested this when building on Linux or other OSes that have their own native getline yet.


nicowilliams:
Thanks!

PR #319, opened 2017-08-10
Set default principal for password change to credential client

rra:
The default client principal for `krb5_set_password`, if the principal argument were `NULL`, was `krb5_get_default_principal`.  But `krb5_set_password` requires credentials for the password change service be passed in, and those credentials are already associated with a client principal that's much more likely to be the correct choice for a default.  Use that principal instead of `krb5_get_default_principal`.


nicowilliams:
Thanks!

PR #320, opened 2017-08-10
Always check password quality

rra:
Previous versions bypassed the password quality check in `kadmind` and `kpasswdd` if the the password was being changed by an administrator instead of the user.  However, password quality checks are how password history should be implemented, and administrator password changes should also go into history.  Check password quality unconditionally.

### DIFF
--- a/kadmin/server.c
+++ b/kadmin/server.c
@@ -318,9 +318,10 @@ kadmind_dispatch(void *kadm_handlep, krb5_boolean initial,
 	 *
 	 * a) allowed by sysadmin
 	 * b) it's for the principal him/herself and this was an
-	 *    initial ticket, but then, check with the password quality
-	 *    function.
+	 *    initial ticket
 	 * c) the user is on the CPW ACL.
+	 *
+	 * All changes are checked for password quality.
 	 */
 
 	if (krb5_config_get_bool_default(contextp->context, NULL, TRUE,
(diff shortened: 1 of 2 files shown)


kadmin/server.c
```diff
@@ -329,6 +330,12 @@ kadmind_dispatch(void *kadm_handlep, krb5_boolean initial,
 	    && krb5_principal_compare (contextp->context, contextp->caller,
 				       princ))
 	{
+	    ret = 0;
```
nicowilliams: Wouldn't it be better to just reverse the sense of the if condition and then drop this `ret = 0; } else {` bit?

macrotex:
The changes in test #293 were written to address this exact patch. When you say "This definitely needs to be bundled with #293" do you mean that #320 should be altered to include the patch from #293, and then #293 deleted?

jaltman:
I believe that always checking password quality should be a configuration option.

lhoward:
It's interesting (but I suppose there may be historical reasons for this) that `allow_self_change_password` is honored only by kadmind (not kpasswdd).

lhoward:
Anyway, I'll integrate with a configuration knob, default to existing behavior?

lhoward:
Someone come up with a better name than `check_on_admin_set` for the option :)

lhoward:
Closing, will resubmit new pull request.

nicowilliams:
> It's interesting (but I suppose there may be historical reasons for this) that allow_self_change_password is honored only by kadmind (not kpasswdd).

I think the reason is just that users were supposed to use kpasswdd for password changes.

PR #321, opened 2017-08-11
lib: Fix printing a short into portstr

cryptomilk:
The size of portstr is too small to print an integer and we should print
a short anyway.

This fixes building with GCC 7.1


nicowilliams:
Thanks!

PR #329, opened 2017-08-29
kdc: fix dh->q allocation check in get_dh_param()

metze-samba:
Please merge these fix, so we can backport it into Samba's copy of Heimdal.

Thanks!


nicowilliams:
Thanks!

PR #332, opened 2017-09-04
Allow to override build date

bmwiedemann:
in order to make builds reproducible.
See https://reproducible-builds.org/ for why this is good
and https://reproducible-builds.org/specs/source-date-epoch/
for the definition of this variable.

This date call is designed to work with both GNU date and BSD date.

### DIFF
--- a/cf/krb-version.m4
+++ b/cf/krb-version.m4
@@ -15,9 +15,15 @@ if test -f include/version.h && cmp -s include/newversion.h.in include/version.h
 	rm -f include/newversion.h.in
 else
  	echo "creating include/version.h"
- 	User=${USER-${LOGNAME}}
- 	Host=`(hostname || uname -n) 2>/dev/null | sed 1q`
- 	Date=`date`
+	if [ -n "$SOURCE_DATE_EPOCH" ] ; then
+		User=reproducible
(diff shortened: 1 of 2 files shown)


nicowilliams:
Why don't we just get rid of the Date altogether?

nicowilliams:
Actually, for releases we shouldn't want the USER, the HOST, nor the DATE.

The only things of interest are:

 - the actual version
 - the HEAD commit hash

bmwiedemann:
Closing after this was solved slightly differently in master with
6d27e0048963955b547584f004109bbbad6e9baf
469366ddb8b38c57c1b443002612078d2760b946

nicowilliams:
@bmwiedemann Was my fix sufficient for your purporses?

bmwiedemann:
@nicowilliams it was sufficient, even though I did not set the SOURCE_HOST variable and was building with different hostnames - that is the part I did not understand, but looking into the build VM's filesystem I see
```c
VERSION_HIDDEN const char *heimdal_long_version = "@(#)$Version: Heimdal 7.4.0 by abuild on localhost (x86_64-suse-linux-gnu) 2017-08-03T12:00:00Z $";
```

PR #336, opened 2017-09-26
Drop build date

bmwiedemann:
and build user and build host
in order to make builds reproducible.
See https://reproducible-builds.org/ for why this is good.

Made this as an alternative approach to #332

### DIFF
--- a/cf/krb-version.m4
+++ b/cf/krb-version.m4
@@ -6,7 +6,7 @@ dnl
 
 AC_DEFUN([AC_KRB_VERSION],[
 cat > include/newversion.h.in <<FOOBAR
-const char *${PACKAGE_TARNAME}_long_version = "@(#)\$Version: $PACKAGE_STRING by @USER@ on @HOST@ ($host) @DATE@ \$";
+const char *${PACKAGE_TARNAME}_long_version = "@(#)\$Version: $PACKAGE_STRING \$";
 const char *${PACKAGE_TARNAME}_version = "$PACKAGE_STRING";
 FOOBAR
 
@@ -15,10 +15,7 @@ if test -f include/version.h && cmp -s include/newversion.h.in include/version.h
 	rm -f include/newversion.h.in
 else
  	echo "creating include/version.h"
- 	User=${USER-${LOGNAME}}
- 	Host=`(hostname || uname -n) 2>/dev/null | sed 1q`
- 	Date=`date`
 	mv -f include/newversion.h.in include/version.h.in
-	sed -e "s/@USER@/$User/" -e "s/@HOST@/$Host/" -e "s/@DATE@/$Date/" include/version.h.in > include/version.h
+	sed -e "" include/version.h.in > include/version.h
 fi
 ])
--- a/configure.ac
+++ b/configure.ac
@@ -667,7 +667,7 @@ cat > include/newversion.h.in <<EOF
 #ifndef VERSION_HIDDEN
 #define VERSION_HIDDEN
 #endif
-VERSION_HIDDEN const char *heimdal_long_version = "@([#])\$Version: $PACKAGE_STRING by @USER@ on @HOST@ ($host) @DATE@ \$";
+VERSION_HIDDEN const char *heimdal_long_version = "@([#])\$Version: $PACKAGE_STRING \$";
 VERSION_HIDDEN const char *heimdal_version = "AC_PACKAGE_STRING";
 EOF
 
@@ -676,9 +676,6 @@ if test -f include/version.h && cmp -s include/newversion.h.in include/version.h
 	rm -f include/newversion.h.in
 else
  	echo "creating include/version.h"
- 	User=${USER-${LOGNAME}}
- 	Host=`(hostname || uname -n || echo unknown) 2>/dev/null | sed 1q`
- 	Date=`date`
 	mv -f include/newversion.h.in include/version.h.in
-	sed -e "s/@USER@/$User/" -e "s/@HOST@/$Host/" -e "s/@DATE@/$Date/" include/version.h.in > include/version.h
+	sed -e "" include/version.h.in > include/version.h
 fi


abartlet:
I strongly support this.  Samba made a very similar change for 4.7. 
while back in the 2000s this was a handy thing to embed, the work of
reproducible builds is much more important now. 

Andrew Bartlett
-- 
Andrew Bartlett                       http://samba.org/~abartlet/
Authentication Developer, Samba Team  http://samba.org
Samba Developer, Catalyst IT          http://catalyst.net.nz/services/samba

nicowilliams:
@abartlet So do I.  Is there any reason to keep any of `HOST`, `USER`, or `DATE` ever?  I'd rather remove them altogether.  The only things I want are the actual version, maybe the branch, closest tag, HEAD commit, and/or indication of whether the clone built in is dirty.

abartlet:
On Tue, 2017-09-26 at 15:29 +0000, Nico Williams wrote:
> @abartlet So do I. Is there any reason to keep any of HOST, USER, or DATE ever? 

It used to be handy to track when/how a binary was created, but we
didn't get any complaints (and I waited time and really expected
someone to have a view) when we dropped it form Samba for the same
reasons. 

> I'd rather remove them altogether. 

I agree. 

> The only things I want are the actual version, maybe the branch, closest tag, HEAD commit, and/or indication of whether the clone built in is dirty.

Yes.

Andrew Bartlett

-- 
Andrew Bartlett                       http://samba.org/~abartlet/
Authentication Developer, Samba Team  http://samba.org
Samba Developer, Catalyst IT          http://catalyst.net.nz/services/samba

nicowilliams:
I'll push my own version of this.  I'll just use various git things.  I'm testing this:

```
diff --git a/configure.ac b/configure.ac
index 806b639..4d4237f 100644
--- a/configure.ac
+++ b/configure.ac
@@ -667,7 +667,7 @@ cat > include/newversion.h.in <<EOF
 #ifndef VERSION_HIDDEN
 #define VERSION_HIDDEN
 #endif
-VERSION_HIDDEN const char *heimdal_long_version = "@([#])\$Version: $PACKAGE_STRING by @USER@ on @HOST@ ($host) @DATE@ \$";
+VERSION_HIDDEN const char *heimdal_long_version = "@([#])\$Version: $PACKAGE_STRING @BRANCH@ @TAG@ ($host) @COMMIT@ \$";
 VERSION_HIDDEN const char *heimdal_version = "AC_PACKAGE_STRING";
 EOF

@@ -676,9 +676,19 @@ if test -f include/version.h && cmp -s include/newversion.h.in include/version.h
        rm -f include/newversion.h.in
 else
        echo "creating include/version.h"
-       User=${USER-${LOGNAME}}
-       Host=`(hostname || uname -n || echo unknown) 2>/dev/null | sed 1q`
-       Date=`date`
+        if test -d "$srcdir/.git"; then
+            GitCommit=`git rev-parse HEAD`
+            GitBranch=`git rev-parse --abbrev-ref HEAD`
+            if test "x$GitBranch" = master; then
+                GitDesc=`git describe --all --dirty`
+            else
+                GitDesc=`git describe --tags --match 'heimdal-*' --dirty`
+            fi
+        else
+            GitCommit='<commit-unknown>'
+            GitBranch='<branch-unknown>'
+            GitDesc='<tag-unknown>'
+        fi
        mv -f include/newversion.h.in include/version.h.in
-       sed -e "s/@USER@/$User/" -e "s/@HOST@/$Host/" -e "s/@DATE@/$Date/" include/version.h.in > include/version.h
+       sed -e "s;@BRANCH@;$GitBranch;" -e "s;@TAG@;$GitDesc;" -e "s;@COMMIT@;$GitCommit;" include/version.h.in > include/version.h
 fi
```

Similarly for `cf/krb-version.m4`.  Why do we have that duplication??  We don't use the `AC_KRB_VERSION` macro...  I'll try removing it.

nicowilliams:
One thing that's clear is that we need a heimdal-8.0pre tag so we don't get this:

```
#ifndef VERSION_HIDDEN
#define VERSION_HIDDEN
#endif
VERSION_HIDDEN const char *heimdal_long_version = "@(#)$Version: Heimdal 7.99.1 master heimdal-1.5pre2-1992-gaef3843-dirty (x86_64-pc-linux-gnu) aef3843b552c4694cdfe99cca2896579561a1a2c $";
VERSION_HIDDEN const char *heimdal_version = "Heimdal 7.99.1";
#ifndef VERSION_HIDDEN
#define VERSION_HIDDEN
#endif
VERSION_HIDDEN const char *heimdal_long_version = "@(#)$Version: Heimdal 7.99.1 @BRANCH@ @TAG@ (x86_64-pc-linux-gnu) @COMMIT@ $";
VERSION_HIDDEN const char *heimdal_version = "Heimdal 7.99.1";
```

(And, oy, where's that duplication coming from?)

nicowilliams:
What do you think of this version of these changes:

https://github.com/heimdal/heimdal/compare/master...nicowilliams:master?expand=1

?

jaltman:
We recently uncovered a situation where a modified binary Heimdal distribution which included malware was distributed via sourceforge.  One of the details that made it obvious that the build details (timestamp, host, user, etc) were present in the build.   That evidence was important when it came to convincing sourceforge to take down the binaries.

My preference is for the default build to include these details in the version string and provide an option that can be specified for reproducible builds

kaduk:
IIUC the recommended "option" to enable the reproducible build is to detect that the SOURCE_DATE_EPOCH environment variable is set (and use that as the datestamp if needed).

nicowilliams:
I see.  Mind you, the git info my variant of this patch uses... would have revealed the sourceforge copy to be illegitimate too.

nicowilliams:
I've pushed another version: https://github.com/nicowilliams/heimdal/commit/6d27e0048963955b547584f004109bbbad6e9baf

This one uses these env vars: `SOURCE_HOST`, `SOURCE_USER`, and `SOURCE_DATE_EPOCH`.  Only `SOURCE_DATE_EPOCH` appears to be "standard".  I'm still adding git branch/tag/commit hash information when building from a git clone.

nicowilliams:
I've pushed an alternative version.

nicowilliams:
Thanks!

bmwiedemann:
@nicowilliams you used `date -d` which only works with GNU date, not BSD date, which is why #332 used a more complex date call ... or is BSD not a supported platform?

nicowilliams:
@bmwiedemann Oh, `date -u -d "@$SOURCE_DATE_EPOCH" 2>/dev/null || date -u -r "$SOURCE_DATE_EPOCH" 2>/dev/null || date -u`?  Sorry, I just simply missed that.  I'll fix it!

nicowilliams:
Illumos has nothing like `-r`/`-d`.  Sigh.

nicowilliams:
I've pushed a fix.  Thanks for noticing this!

PR #340, opened 2017-10-10
Do not fsync() every record when loading HDB from a dump (kadmin and iprop)

nicowilliams:
This PR makes `kadmin load` and `ipropd-slave`'s `receive_everything()` much faster by not doing any `fsync()`s or `fdatasync()`s or similar (e.g., `O_SYNC`, `sync(2)`, ...).

This should prevent a situation in iprop where a slave receiving a full dump can cause all slaves to persistently need to receive full dumps.  This is not a complete fix for that problem, however.  A more complete fix will follow.

### DIFF
--- a/kadmin/load.c
+++ b/kadmin/load.c
@@ -404,6 +404,7 @@ doit(const char *filename, int mergep)
 	fclose(f);
 	return 1;
     }
+    (void) db->hdb_set_sync(context, db, 0);
     line = 0;
     ret = 0;
     while(fgets(s, sizeof(s), f) != NULL) {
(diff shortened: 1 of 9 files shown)


lib/kadm5/ipropd_slave.c
```diff
@@ -563,6 +565,9 @@ receive_everything(krb5_context context, int fd,
 
     reinit_log(context, server_context, vno);
 
+    ret = mydb->hdb_set_sync(context, mydb, 1);
+    if (ret)
+        krb5_warn(context, ret, "failed to sync the received HDB");
```
vdukhovni: I think that sync failure needs to be fatal.  You're clobbering `ret` with the return value of `hdb_close`.

kadmin/load.c
```diff
@@ -547,6 +548,9 @@ doit(const char *filename, int mergep)
 	    break;
 	}
     }
+    ret = db->hdb_set_sync(context, db, 1);
+    if (ret)
+        krb5_warn(context, ret, "failed to sync the HDB");
```
vdukhovni: This one is better.

PR #344, opened 2017-10-17
Further improve kdc child process management

vdukhovni:
- Simplify child process creation rate limit by moving
  select_sleep() to happen right after the parent fork().

- Consider child pid table slots to be empty if the pid is
  non-positive, rather than just -1 or just 0.

- Log warnings should we ever spawn a child with no free
  slot to track it, or reap a child that does not match
  a tracked slot.

### DIFF
--- a/kdc/connect.c
+++ b/kdc/connect.c
@@ -1025,44 +1025,49 @@ reap_kid(krb5_context context, krb5_kdc_configuration *config,
     char *what;
     int status;
     int i = 0; /* quiet warnings */
+    int ret = 0;
+    int level = 0;
+    const char *sev = "";
 
     pid = waitpid(-1, &status, options);
-    if (pid < 1)
+    if (pid <= 0)
 	return 0;
 
-    if (pid != bonjour_pid) {
+    if (pid == bonjour_pid) {
+        bonjour_pid = (pid_t)-1;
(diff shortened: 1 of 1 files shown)


kdc/connect.c
```diff
@@ -1122,11 +1127,9 @@ start_kdc(krb5_context context,
     if (max_kdcs < 1)
 	max_kdcs = 1;
 
-    pids = malloc(max_kdcs * sizeof(*pids));
```
jaltman: I would prefer that this block be left alone and that an empty slot always be (pid_t)-1 instead of  0 or -1.

kdc/connect.c
```diff
@@ -1199,15 +1187,22 @@ start_kdc(krb5_context context,
                 break;
             default:
 		for (i = 0; i < max_kdcs; i++) {
-		    if (pids[i] == (pid_t)-1) {
```
jaltman: ditto.

vdukhovni: Hence tolerate either 0 or -1 here, create as 0, and assign to -1 when clearing a slot, ensuring that we exercise both cases.  And log any unexpected behaviour if this still breaks.

Not religious on this point, but I think it may be slightly more resilient to bitrot.

PR #354, opened 2017-12-07
Fix size types in Heimdal

cryptomilk:
Hi,

this patchset fixes several size types found by -Wstrict-overflow=2.

Andreas

### DIFF
--- a/lib/asn1/gen.c
+++ b/lib/asn1/gen.c
@@ -513,7 +513,7 @@ is_primitive_type(int type)
 }
 
 static void
-space(int level)
+space(size_t level)
 {
     while(level-- > 0)
 	fprintf(headerfile, "  ");
@@ -542,7 +542,7 @@ have_ellipsis(Type *t)
 }
 
 static void
-define_asn1 (int level, Type *t)
+define_asn1 (size_t level, Type *t)
 {
     switch (t->type) {
     case TType:
--- a/lib/hx509/cert.c
+++ b/lib/hx509/cert.c
@@ -1995,7 +1995,8 @@ hx509_verify_path(hx509_context context,
 {
     hx509_name_constraints nc;
     hx509_path path;
-    int ret, proxy_cert_depth, selfsigned_depth, diff;
+    int ret, diff;
+    size_t proxy_cert_depth, selfsigned_depth;
     size_t i, k;
     enum certtype type;
     Name proxy_issuer;
--- a/lib/krb5/config_file.c
+++ b/lib/krb5/config_file.c
@@ -1081,7 +1081,7 @@ krb5_config_vget_strings(krb5_context context,
 			 va_list args)
 {
     char **strings = NULL;
-    int nstr = 0;
+    size_t nstr = 0;
     const krb5_config_binding *b = NULL;
     const char *p;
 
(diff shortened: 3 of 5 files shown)


quanah:
I wonder if this will fix Solaris so it works.

quanah:
Unfortunately, doesn't resolve the issues with Solaris.

jaltman:
While this pull request might suppress theoretical overflows, I don't believe any of the locations are actually at risk of overflow given the usage.

abartlet:
To be clear, these patches are important to Samba not because of a confirmed overflow, but because it allows Samba (which hopes to rebase on a current git snapshot eventually) to build with this warning locked on with -Werror.  We do this to detect other possibly more serious issues.

I plan to review and merge these shortly.

jaltman:
The reason that I have not reviewed these changes are because I can't tell from the commit messages what is supposed to be fixed.  I would like the commit messages to include the explicit warning messages that are being fixed by the proposed changes. 

As an example, the lib/asn1/gen.c define_asn1() and space() change of the size of level makes no sense to me.  All of the integer values that are passed to define_asn1() as level are small integers.    I can't see how size_t vs int should make any difference.  

If the problem is that there is compiler being stupid, then I would also like the compiler version information included in the commit message.

abartlet:
Thanks for the more detailed feedback, that is a reasonable request and gives Andreas something to work with.

cryptomilk:
> As an example, the lib/asn1/gen.c define_asn1() and space() change of the size of level makes no sense to me.

* Signed overflows are undefined in C: https://port70.net/~nsz/c/c11/n1570.html#3.4.3p3
* Unsigned integers never overflow: https://port70.net/~nsz/c/c11/n1570.html#6.2.5p9
* As signed overflows are undefined, the compilers is free to assume that they don't occur and the code is thrown away by the optimizer. It assume the overflow never occurred.

The solution is to use unsigned arithmetic, which never overflows (in the C Standard sense of the word).

Here is an example which is the same as the `space()` function in `lib/asn1/gen.c`. I just call it dot to print a `.` to stdout.

```
$ cat integer_overflow.c 
#include <stdio.h>
#include <limits.h>

static void dot(int level)
{
    while(level-- > 0) {
        printf(".");
    }
}

int main(void)
{
    dot(1 - INT_MAX);
    printf("WHAT A WURST!\n");

    return 0;
}
$ gcc -O2 integer_overflow.c 
$ ./a.out 
WHAT A WURST!
```

As you can see, the optimizer has thrown away the while-loop. Newer gcc versions don't produce the warning anymore. I've opened a bug for that. The above example demonstrates the problem gcc warned about. That's the reason why I fixed the size types here.

jaltman:
@cryptomilk Thank you for taking the time to post your example but dot() is called with a negative value.   The conditional in the first iteration of the while loop will be `((1 - INT_MAX) > 0)` which evaluates to FALSE.  Therefore the loop never executes.   I'm sorry but the example doesn't demonstrate what you are attempting to show.

jaltman:
In the latest example the compiler knows at compile time that an overflow has occurred and so is free to do what it wants.   That is not true for parameters passed to functions whose values are determined at runtime from external inputs.   Any compiler that refuses to generate code because some input might be bad is broken.   The compiler is free to implement a behavior of its choice in the undefined case; its not permitted to refuse to process values that are within the defined range.

cryptomilk:
Then just close this PR. I can't reproduce it as I don't get those warnings anymore ...

PR #365, opened 2018-03-22
getauxval: return 0 in case returning 0 when getauxval doesn't set errno

dariaphoebe:
if when we need to determine if getauxval sets errno, we determine it doesn't
after getting a 0 return code, make sure we return a value, since we didn't
previously.

### DIFF
--- a/lib/roken/getauxval.c
+++ b/lib/roken/getauxval.c
@@ -213,6 +213,7 @@ rk_getauxval(unsigned long type)
 
     getauxval_sets_errno = 0;
     errno = save_errno;
+    return 0;
 #endif
 #else
     const auxv_t *a;


lib/roken/getauxval.c
```diff
@@ -213,6 +213,7 @@ rk_getauxval(unsigned long type)
 
     getauxval_sets_errno = 0;
     errno = save_errno;
+    return 0;
```
jaltman: Reading the getauxval() the correct return value here should be "ret2" which is the value successfully returned by the getauxval() call on line 207.

dariaphoebe: your choices are 0 or ret. ret2 is a throwaway value. i think ret (not ret2) is correct

dariaphoebe: actually, look at the code above this. shouldn't it
   
    if ((a = rk_getauxv(type)) == NULL) {
        errno = ENOENT;
        return 0;
    }
    return a->a_un.a_val;

lib/roken/getauxval.c
```diff
@@ -213,6 +213,7 @@ rk_getauxval(unsigned long type)
...
 #endif
 #else
```
jaltman: @nicowilliams Did you mean for the code to fall-through and this #else should really be a #endif?

dariaphoebe: can't simply fall though without moving the declaration of a to the top of the function, but that's what it looks like it should do

dariaphoebe: alternate version of this in 366, which i think is right

PR #375, opened 2018-04-19
Solaris10

dariaphoebe:
deal with older versions of tr and lex that exist on solaris 10.

### DIFF
--- a/lib/com_err/lex.l
+++ b/lib/com_err/lex.l
@@ -53,8 +53,6 @@ static int getstring(void);
 
 %}
 
-%option nounput
-
 %%
 et			{ return ET; }
 error_table		{ return ET; }
--- a/lib/libedit/src/makelist
+++ b/lib/libedit/src/makelist
@@ -127,7 +127,7 @@ case $FLAG in
 #
 -fh)
     cat $FILES | $AWK '/el_action_t/ { print $3 }' | \
-    sort | tr '[:lower:]' '[:upper:]' | $AWK '
+    sort | tr [a-z] [A-Z] | tr -d ' ' | $AWK '
 	BEGIN {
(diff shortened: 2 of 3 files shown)


jaltman:
@lhoward please review

lib/libedit/src/makelist
```diff
@@ -127,7 +127,7 @@ case $FLAG in
 #
 -fh)
     cat $FILES | $AWK '/el_action_t/ { print $3 }' | \
-    sort | tr '[:lower:]' '[:upper:]' | $AWK '
+    sort | tr [a-z] [A-Z] | tr -d ' ' | $AWK '
```
nicowilliams: Missing quotes.

lib/com_err/lex.l
```diff
@@ -53,8 +53,6 @@ static int getstring(void);
 
 %}
 
-%option nounput
```
nicowilliams: Does this cause problems on systems where lex generates an `unput()`?

lib/libedit/src/makelist
nicowilliams: That is, the `[a-z]` and `[A-Z]` need to be `'[a-z]'` and `'[A-Z]'`.

lib/com_err/lex.l
lhoward: Probably will just cause a warning?

lhoward: Could be also useful to check for flex in configure, and if we're using it, pass --nounput on the command line. Do we want Heimdal to be able to compile without unused function warnings?

dariaphoebe:
Managed to submit as a new PR, sorry.

#454

nicowilliams:
Pushed via #454.

PR #379, opened 2018-05-11
Re-add support for Solaris Doors IPC mechanism

lhoward:
Creating PR to track branch, work is not yet finished

KCM et al used to support doors (a lightweight IPC mechanism in Solaris) but this was removed when the Heimdal IPC library was introduced. These patches re-implement it.

### DIFF
--- a/kcm/config.c
+++ b/kcm/config.c
@@ -40,7 +40,6 @@ static const char *config_file;	/* location of kcm config file */
 
 size_t max_request = 0;		/* maximal size of a request */
 char *socket_path = NULL;
-char *door_path = NULL;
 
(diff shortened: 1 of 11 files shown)


lhoward:
KCM was completely broken on Solaris anyway because IPC did not work (#382). So we can safely assume nobody was using it. Which makes me feel safer about merging this code. @nicowilliams could you take a look?

lhoward:
Merged in 73fde337. I figure that, given Unix domain socket IPC was completely broken on Solaris, this is fairly low impact to merge. Should only affect platforms with HAVE_DOOR_CREATE.

PR #384, opened 2018-05-15
Fully implement iovecs for aes-sha1 encryption types

SimonWilkinson:
This set of changes push iovecs down into the krb5 crypto code, so that we use iovecs for encryption, decryption and checksumming when the _iov variants of the encryption and checksum functions are called by the library user.

iovecs are currently only used for AES-SHA1 based encryption types.


lib/hcrypto/hmac.c
```diff
@@ -80,49 +93,39 @@ HMAC_Init_ex(HMAC_CTX *ctx,
...
+            HMAC_CTX_cleanup(ctx);
+
 	ctx->md = md;
-	if (ctx->buf) {
-	    memset(ctx->buf, 0, ctx->key_length);
-	    free (ctx->buf);
-	}
 	ctx->key_length = EVP_MD_size(ctx->md);
 	ctx->buf = malloc(ctx->key_length);
+        ctx->opad = malloc(blockSize);
```
lhoward: Suppose you could check the return value of malloc but – it's no worse than the code that was already there!

SimonWilkinson: I've actually got another change (not pushed yet) that makes all of these buffers go away. The compilers I have tested generate much more efficient code when the size of buf, opad and ipad are known at compilation time, and it's worth the overhead of setting these to the maximum known hash block size and doing more iterations in order to get the more efficient code.

lib/krb5/crypto-evp.c
```diff
@@ -190,8 +190,387 @@ _krb5_evp_encrypt(krb5_context context,
...
+}
+
+/* Move the cursor along to the start of the next block to be
+ * encrypted */
+static inline void
+_krb5_evp_iov_cursor_nextcrypt(struct _krb5_evp_iov_cursor *cursor)
+{
+    int i;
+
+    for (i = cursor->nextidx; i < cursor->niov; i++) {
```
SimonWilkinson: I think there's a termination problem here - cursor->nextidx needs to be advanced even if the last iovec isn't one that should be encrypted. Otherwise we'll never hit cursor->nextidx == niov which is required for some loop terminations.

PR #389, opened 2018-07-14
S4U2Self: fix impersonate with multiple realms using referrals

frenche:
When impersonating a user from another realm, we first need
to obtain a TGT to TGS in that user's realm.
Only then we can ask for a ticket for self and follow back
referrals to our own realm.

See MS-SFU sections 3.1.5.1.1.1 and 4.2

More details at:
http://www.h5l.org/pipermail/heimdal-discuss/2018-July/000381.html


frenche:
See also Samba PR (and the thread on Samba mailing list):
https://github.com/samba-team/samba/pull/204

frenche:
Superseded by #403

PR #391, opened 2018-07-17
Some little cleanups

iboukris:


nicowilliams:
Commit 0a67e642bbe56c70c627cfdfd2dcae175d9a924a should be removed.  Just because the ccache isn't currently being used in those functions doesn't mean that it doesn't belong, and we might want to use it in the future for accessing things like `ccconfig` entries.  Builds should not be failing, but we wouldn't object to adding a `(void)id;` to such functions.

nicowilliams:
Cherry-picked and pushed.

PR #395, opened 2018-08-01
del_enctype with specific kvno (cpw --keepold counterpart)

bodik:
usecase for the PR is the krbtgt/ rollover. one can add new keys using `cpw --keepold` but removing old keys/enctypes with specific kvno is not currenly supported.


bodik:
not working as expected, other approach is needed to prune historic keys from hdb

PR #403, opened 2018-09-01
Fix cross realm S4U2Self V2

iboukris:
Including KDC fixes.

Follow-up to #389

### DIFF
--- a/lib/krb5/get_cred.c
+++ b/lib/krb5/get_cred.c
@@ -1036,6 +1036,7 @@ get_cred_kdc_referral(krb5_context context,
     krb5_creds tgt, referral, ticket;
     krb5_creds **referral_tgts = NULL;  /* used for loop detection */
     int loop = 0;
+    krb5_boolean got_tgt = false;
     int ok_as_delegate = 1;
     int want_tgt;
     size_t i;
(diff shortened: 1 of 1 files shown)


nicowilliams:
Thank you for this contribution.  I've not reviewed all of it, though the general quality looks pretty good.  This is quite large and we'll have to review this in the context of the S4U docs, and that will take time.

lib/krb5/pac.c
```diff
@@ -843,6 +849,23 @@ krb5_pac_verify(krb5_context context,
...
+    return krb5_pac_verify_ex(context, pac, authtime, principal,
+			      server, privsvr, false);
+}
+
+/* Verify PAC, with option to include the realm when
+ * comparing the logon-name. Needed for KDC processing
+ * of cross realm s4u2self tickets */
+
+KRB5_LIB_FUNCTION krb5_error_code KRB5_LIB_CALL
+krb5_pac_verify_ex(krb5_context context,
```
lhoward: If this is just for KDC use, then prefix with an underscore

kdc/krb5tgs.c
```diff
@@ -316,54 +306,171 @@ check_PAC(krb5_context context,
...
+	krb5_boolean with_realm;
+
+	/* In case the impersonated principal belongs to our krbtgt realm,
+	 * then we expect the PAC of the impersonating service */
+	if (s4u2self_impersonated_client != NULL) {
+	    pac_principal = client_principal;
+	    db_client = client;
+	    with_realm = false;
+	}
+	else {
```
lhoward: fold line

iboukris:
Thanks for looking at this, note that it has become a bit work-in-progress as I'd like to update it with some insights learned while implementing related work in MIT (like more constraints to cross-realm s4u2self and add in-tree tests without PAC).
However, the main logic is the same and I could use feedback on the current work.

I'm testing it with samba against Windows DC, I've also submitted a similar merge request against Samba with tests that exercise it:
https://gitlab.com/samba-team/devel/samba/commits/iboukris_xrealm_s4u2self_new

PR #406, opened 2018-09-10
Allow zero-length encrypt IOV elements in _krb5_evp_encrypt_iov_cts()

abartlet:
The test here is not a failing test, it will (as discoved via Samba in an attempted heimdal upgrade)
simply spin to 100% CPU.

### DIFF
--- a/lib/krb5/crypto-evp.c
+++ b/lib/krb5/crypto-evp.c
@@ -491,9 +491,12 @@ _krb5_evp_encrypt_iov_cts(krb5_context context,
 		remaining -= wholeblocks;
 	    }
 
-	    /* Then, if we have partial data left, steal enough from subsequent
-	     * iovecs to make a whole block */
-	    if (cursor.current.length > 0 && cursor.current.length < blocksize) {
+	    /* Then steal enough from subsequent iovecs to make a
+	     * whole block.  We need to do this regardless of if there
+	     * is zero data left, for correctness and to avoid a spin
+	     * on remaining > 0
+	     */
+	    if (cursor.current.length < blocksize) {
 		if (encryptp && remaining == blocksize)
 		    lastpos = cursor;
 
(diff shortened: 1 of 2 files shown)


jaltman:
@SimonWilkinson please review

lib/krb5/crypto-evp.c
```diff
@@ -491,9 +491,12 @@ _krb5_evp_encrypt_iov_cts(krb5_context context,
...
 
-	    /* Then, if we have partial data left, steal enough from subsequent
-	     * iovecs to make a whole block */
-	    if (cursor.current.length > 0 && cursor.current.length < blocksize) {
+	    /* Then steal enough from subsequent iovecs to make a
+	     * whole block.  We need to do this regardless of if there
+	     * is zero data left, for correctness and to avoid a spin
+	     * on remaining > 0
+	     */
+	    if (cursor.current.length < blocksize) {
```
SimonWilkinson: This is going to add a single block encryption when we don't need one.

If the iovec is a multiple of the block size, we'll consume the entire iovec in the step above, and cursor.current.length will be 0 here. We'll then end up stealing a whole block of data from the next iovec, and doing a single block encryption with that, regardless of the length of the next iovec. Single block encryption is really expensive with hardware accelerated crypto, and it would be preferable to avoid it where we can.

Is the real problem here not that _krb5_evp_iov_cursor_nextcrypt() should be skipping iovecs of 0 length?

abartlet: OK.  I can't really comment on the performance side of things, I just don't want the CPU spin, making no progress kind of hurts functionality :-)

I wouldn't have though that 1 extra call to anything would be a big drama, ie O(1) operation, not O(N) unless some caller does something (even more) truly silly with the IO vector, like having 100s of empty entries.  Samba provides a single empty vector in an edge case. 

That said, how do we fix this?  I'm even less clear on what has to be done in _krb5_evp_iov_cursor_nextcrypt() than in this patch.  Can you use the tests I wrote and propose a more correct change?

abartlet: Ahh, that makes much more sense now, which is also what worries me, this is complex code beyond my full understanding. Thank you very much for looking into a better fix.

SimonWilkinson:
https://github.com/heimdal/heimdal/pull/422 is an alternative fix. I've taken the liberty of using the tests from this change.

PR #410, opened 2018-09-10
Address some compiler warnings

abartlet:
These compiler warnings come via Samba.

Some were already submitted in #354 but I re-raise them with the justifications requested as far as I've been able to reproduce on Ubuntu 16.04 gcc 5.4.0

I realise not all the patches and chunks have a justification, but increasing the type width in (eg) asn1.c is pretty low-risk. 

It would be very helpful if these could be accepted upstream as it will reduce the number of patches we need to maintain in our lorikeet-heimdal fork.

Thanks!

### DIFF
--- a/lib/krb5/config_file.c
+++ b/lib/krb5/config_file.c
@@ -1081,7 +1081,7 @@ krb5_config_vget_strings(krb5_context context,
 			 va_list args)
 {
     char **strings = NULL;
-    int nstr = 0;
+    size_t nstr = 0;
     const krb5_config_binding *b = NULL;
     const char *p;
 
--- a/lib/krb5/keytab_any.c
+++ b/lib/krb5/keytab_any.c
@@ -222,11 +222,11 @@ any_remove_entry(krb5_context context,
 {
     struct any_data *a = id->data;
     krb5_error_code ret;
-    int found = 0;
+    krb5_boolean found = FALSE;
     while(a != NULL) {
 	ret = krb5_kt_remove_entry(context, a->kt, entry);
 	if(ret == 0)
-	    found++;
+	    found = TRUE;
 	else {
 	    if(ret != KRB5_KT_NOWRITE && ret != KRB5_KT_NOTFOUND) {
 		krb5_set_error_message(context, ret,
--- a/lib/krb5/krbhst.c
+++ b/lib/krb5/krbhst.c
@@ -353,14 +353,15 @@ krb5_krbhst_format_string(krb5_context context, const krb5_krbhst_info *host,
 			  char *hostname, size_t hostlen)
 {
     const char *proto = "";
-    char portstr[7] = "";
     if(host->proto == KRB5_KRBHST_TCP)
 	proto = "tcp/";
     else if(host->proto == KRB5_KRBHST_HTTP)
 	proto = "http://";
-    if(host->port != host->def_port)
-	snprintf(portstr, sizeof(portstr), ":%d", host->port);
-    snprintf(hostname, hostlen, "%s%s%s", proto, host->hostname, portstr);
+    if (host->port != host->def_port) {
+	snprintf(hostname, hostlen, "%s%s:%d", proto, host->hostname, host->port);
+    } else {
+	snprintf(hostname, hostlen, "%s%s", proto, host->hostname);
+    }
     return 0;
 }
 
--- a/lib/krb5/addr_families.c
+++ b/lib/krb5/addr_families.c
@@ -804,7 +804,7 @@ static struct addr_operations at[] = {
     }
 };
 
-static int num_addrs = sizeof(at) / sizeof(at[0]);
+static size_t num_addrs = sizeof(at) / sizeof(at[0]);
 
 static size_t max_sockaddr_size = 0;
 
@@ -815,22 +815,26 @@ static size_t max_sockaddr_size = 0;
 static struct addr_operations *
 find_af(int af)
 {
-    struct addr_operations *a;
+    size_t i;
 
-    for (a = at; a < at + num_addrs; ++a)
-	if (af == a->af)
-	    return a;
+    for (i = 0; i < num_addrs; i++) {
+	if (af == at[i].af) {
+		return &at[i];
+	}
+    }
     return NULL;
 }
 
 static struct addr_operations *
 find_atype(krb5_address_type atype)
 {
-    struct addr_operations *a;
+    size_t i;
 
-    for (a = at; a < at + num_addrs; ++a)
-	if (atype == a->atype)
-	    return a;
+    for (i = 0; i < num_addrs; i++) {
+	if (atype == at[i].atype) {
+		return &at[i];
+	}
+    }
     return NULL;
 }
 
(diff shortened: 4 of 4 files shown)


lib/krb5/krbhst.c
```diff
@@ -353,13 +353,13 @@ krb5_krbhst_format_string(krb5_context context, const krb5_krbhst_info *host,
...
+    char portstr[7] = {0};
     if(host->proto == KRB5_KRBHST_TCP)
 	proto = "tcp/";
     else if(host->proto == KRB5_KRBHST_HTTP)
 	proto = "http://";
     if(host->port != host->def_port)
-	snprintf(portstr, sizeof(portstr), ":%d", host->port);
+	snprintf(portstr, sizeof(portstr), ":%hd", host->port);
     snprintf(hostname, hostlen, "%s%s%s", proto, host->hostname, portstr);
     return 0;
```
jaltman: I would prefer to see this rewritten to remove portstr[] entirely

  if (host->port != host->def_port)
    snprintf(hostname, hostlen, "%s%s%hd", proto, host->hostname, host->port);
  else
    snprintf(hostname, hostlen, "%s%s", proto, host->hostname);

lib/gssapi/mech/gss_set_cred_option.c
```diff
@@ -41,7 +41,7 @@ gss_set_cred_option (OM_uint32 *minor_status,
 	struct _gss_cred *cred = (struct _gss_cred *) *cred_handle;
 	OM_uint32	major_status = GSS_S_COMPLETE;
 	struct _gss_mechanism_cred *mc;
-	int one_ok = 0;
+	OM_uint32 one_ok = 0;
```
jaltman: one_ok is not a protocol variable.  Its a local variable used as a boolean flag.  Its proper type is 'int'.

abartlet: Thanks, I agree.  I can't see how this was ever an issue, I'll drop that hunk.

abartlet:
I've addressed the feedback.

Is this OK to merge now?

Thanks,

Andrew Bartlett

jaltman:
The Travis builders hung when running the tests.

lib/krb5/krbhst.c
```diff
@@ -353,14 +353,15 @@ krb5_krbhst_format_string(krb5_context context, const krb5_krbhst_info *host,
...
     const char *proto = "";
-    char portstr[7] = "";
     if(host->proto == KRB5_KRBHST_TCP)
 	proto = "tcp/";
     else if(host->proto == KRB5_KRBHST_HTTP)
 	proto = "http://";
-    if(host->port != host->def_port)
-	snprintf(portstr, sizeof(portstr), ":%d", host->port);
-    snprintf(hostname, hostlen, "%s%s%s", proto, host->hostname, portstr);
+    if(host->port != host->def_port) {
```
jaltman: Although there are lots of instances of `if()` in the tree instead of `if ()` our preferred style is a space after `if` as `if` is not a function.   We are not going to globally fix this because we do not want to destroy the usefulness of git blame.

abartlet: Thanks, is there a document on this somewhere?  

(This is Samba's style also, for what it is worth)

abartlet:
> The Travis builders hung when running the tests.

I've isolated the patch that does this, and am very glad such tests exist.  I've cut Andreas's patch down to the parts I could reproduce on Ubuntu 16.04 and which I included in the commit message and re-pushed.

abartlet:
Sorry for the drama, I finally worked out what was going on:  I botched the first patch that I was asked to re-write, rather than cherry-pick.  

To be clear, it wasn't Andreas's patch, but I've left it cut-down to the parts I could reproduce the failure for.

lib/krb5/keytab_any.c
```diff
@@ -222,7 +222,7 @@ any_remove_entry(krb5_context context,
 {
     struct any_data *a = id->data;
     krb5_error_code ret;
-    int found = 0;
+    size_t found = 0;
```
jaltman: 'found' should be acceptable as an 'int'.   Its used as a boolean either zero or non-zero.

Given its usage setting `found = 1;` instead of `found++;` at line 229 is probably the correct fix.

lib/krb5/config_file.c
```diff
@@ -1081,7 +1081,7 @@ krb5_config_vget_strings(krb5_context context,
 			 va_list args)
 {
     char **strings = NULL;
-    int nstr = 0;
+    size_t nstr = 0;
```
jaltman: This is fine.

lib/gssapi/mech/gss_set_cred_option.c
abartlet: Looping back to this one, there is no krb5_boolean available in gsspai, is Heimdal willing to use C99 bool yet?

The issue is that on Ubuntu 14.04 with strict compiler options and -O3 we get:


==> /builds/catalyst-samba/samba/samba-o3.stderr <==
../source4/heimdal/lib/gssapi/mech/gss_set_cred_option.c: In function ‘gss_set_cred_option’:
../source4/heimdal/lib/gssapi/mech/gss_set_cred_option.c:112:5: error: assuming signed overflow does not occur when simplifying conditional to constant [-Werror=strict-overflow]
  if (one_ok) {
     ^
cc1: some warnings being treated as errors

If I put that back into the commit message, can I have this hunk accepted?

jaltman: This warning has to be a compiler bug.    
I'm curious, what happens if you change that to 

   if (!one_ok)
       return major_status;

   *minor_status = 0;
   return GSS_S_COMPLETE;

abartlet:
Of course it is a 6 year old compiler, I'm just trying not to carry
local patches.

We can go around and around forever, but I'm just wasting your time and
not getting any closer to reducing the patches I carry locally.

What do you want to do, for the global picture, because this back and
forth is not productive for either of us. 

I don't get to choose the compiler Samba's autobuild runs with, and for
now it is this one, for better or worse.

Neither can I drop the patches, much as would like, because it then I
can't update Samba's copy.  I have far, far bigger issues to address
but we are both bogged down in this, looking for 'perfect'.

So which do you want:
 - the patches in our branch, so we never get to using upstream master
or
 - change the type of the boolean to uint32_t
or
 - change the type of the boolean to bool?

Thanks,

Andrew Bartlett

abartlet:
To try and move this further along, I'll leave the gss_set_cred_option.c change in our branch until we can get Samba off the old compiler for our CI.  Updated patch set pushed, which should, I hope, have addressed all the review comment provided.

lib/krb5/krbhst.c
lhoward: %hd is widely supported I presume

lhoward:
Merged with stylistic fixes

abartlet:
man 3 snprintf suggests at least since SUSv2:

SUSv2 specified only the length modifiers h (in  hd,
       hi, ho, hx, hX, hn) and l (in ld, li, lo, lx, lX, ln, lc, ls)
and L (in Le, LE, Lf, Lg, LG).

abartlet:
> Merged with stylistic fixes

Thanks!

lhoward:
@nicowilliams suggested int cast instead of %hd so went with that.

lib/krb5/krbhst.c
lhoward: I applied Jeff's fix

nicowilliams: That works.  `%hd` is still not needed (since it's an int), but it's fine.

lhoward: I couldn't push to @abartlet's branch to update this PR, but the fix I committed to master uses %d and an int cast.

PR #418, opened 2018-09-16
Fix leaks in contexts 

huaraz:
There seem to be 2 type of leaks.  

Not 100% sure if the free is the best option, but it seem to work.

### DIFF
--- a/lib/gssapi/ntlm/accept_sec_context.c
+++ b/lib/gssapi/ntlm/accept_sec_context.c
@@ -54,10 +54,13 @@ _gss_ntlm_allocate_ctx(OM_uint32 *minor_status, ntlm_ctx *ctx)
     (*ctx)->server = ns_interface;
 
     maj_stat = (*(*ctx)->server->nsi_init)(minor_status, &(*ctx)->ictx);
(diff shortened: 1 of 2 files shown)


lib/gssapi/ntlm/accept_sec_context.c
```diff
@@ -54,10 +54,13 @@ _gss_ntlm_allocate_ctx(OM_uint32 *minor_status, ntlm_ctx *ctx)
...
 
     maj_stat = (*(*ctx)->server->nsi_init)(minor_status, &(*ctx)->ictx);
-    if (maj_stat != GSS_S_COMPLETE)
-	return maj_stat;
+    if (maj_stat == GSS_S_COMPLETE)
+    	return GSS_S_COMPLETE;
 
-    return GSS_S_COMPLETE;
+    if (*ctx) 
+	free(*ctx);
```
jaltman: Set *ctx = NULL after it has been freed.

lhoward: or GSS_C_NO_CONTEXT

huaraz:
resubmitted new pull request

PR #421, opened 2018-09-18
Partial Python3 compatability

abartlet:
Samba aims to build with python3 in the near future, and this set of patches provides that support in the Heimdal code we have within our build system.


abartlet:
Can this be merged please?

PR #422, opened 2018-09-18
Allow zero-length encrypt IOVs in _krb5_evp_encrypt_iov_cts()

SimonWilkinson:
The iovec encryption code doesn't handle 0 length iovecs correctly.
Instead of just skipping them, _krb5_evp_encrypt_iov_cts() will spin
on the 0 length iovec.

Modify the _krb5_evp_iov_cursor_expand helper so that iovec expansion
simply skips 0 length iovecs, and make _krb5_evp_iov_cursor_nextcrypt
do the same.

Original bug report and tests from Andrew Bartlett <abartlet@samba.org>


abartlet:
I've confirmed this also passes Samba's tests when imported into Samba.

Thanks!

Would you like me to merge it?

PR #424, opened 2018-09-19
lib/krb5: Do not re-send the AS-REQ if fast was disabled

abartlet:
Without this, in an AS-REQ that has no local key for FAST
we would send the AS-REQ again, which can bump the bad
password count on the KDC twice.

So only try again if FAST was actually tried.


abartlet:
Can this patch be merged?  Otherwise Samba's bad password count tests fail as Kerberos sends bad passwords twice as often compared to the user input.

Thanks,

Andrew Bartlett

PR #428, opened 2018-09-26
lib/kadm5: kadmin hook_libraries plugin interface

jaltman:
This change adds plugin support to the kadmin libraries for performing
actions before and after a password change is committed to the KDC database
and after a change is made to the attributes of a principal (specifically,
a change to DISALLOW_ALL_TIX).

This change adds a hook_libraries configuration option to the [kadmin]
section of krb5.conf (or kdc.conf if you use that file) that must be set
to load the module.  That configuration option is in the form:

    [kadmin]
      hook_libraries = /usr/local/lib/krb5/plugins/kadm5_hook/krb5_sync.so

where the value is the full path to the plugin that you want to load.  If
this option is not present, kadmind will not load a plugin and the changes
from the patch will be inactive.  If this option is given and the plugin
cannot be loaded, kadmind startup will abort with a (hopefully useful)
error message in syslog.

Any plugin used with this patch must expose a public struct named
kadm5_hook.  That struct must contain the following:

    typedef struct kadm5_hook {
        const char *name;
        int version;
        const char *vendor;

        krb5_error_code (*init)(krb5_context, void **);
        void (*fini)(krb5_context, void *);

        krb5_error_code (*chpass)(krb5_context, void *, enum kadm5_hook_stage,
                                  krb5_principal, const char *);
        krb5_error_code (*create)(krb5_context, void *, enum kadm5_hook_stage,
                                  kadm5_principal_ent_t, uint32_t mask,
                                  const char *password);
        krb5_error_code (*modify)(krb5_context, void *, enum kadm5_hook_stage,
                                  kadm5_principal_ent_t, uint32_t mask);
    } kadm5_hook;

where enum kadm5_hook_stage is:

    enum kadm5_hook_stage {
        KADM5_HOOK_STAGE_PRECOMMIT,
        KADM5_HOOK_STAGE_POSTCOMMIT
    };

init creates a hook context that is passed into all subsequent calls.
chpass is called for password changes, create is called for principal
creation (with the newly-created principal in the kadm5_principal_ent_t
argument), and modify is called when a principal is modified.

These functions should follow the normal Kerberos calling convention of
returning 0 on success and a Kerberos error code on failure, setting the
Kerberos error message in the provided context.

This change is submitted under the following license

  Copyright 2012, 2013
    The Board of Trustees of the Leland Stanford Junior University

  Copying and distribution of this file, with or without modification, are
  permitted in any medium without royalty provided the copyright notice and
  this notice are preserved.  This file is offered as-is, without any
  warranty.

Change-Id: Iebf2c0ac5767f79fe72cc4a1c74521f3cb790d12

### DIFF
--- a/lib/kadm5/kadm5-hook.h
+++ b/lib/kadm5/kadm5-hook.h
@@ -1,0 +1,81 @@
+/*
+ * Copyright 2010
+ *     The Board of Trustees of the Leland Stanford Junior University
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * 3. Neither the name of the Institute nor the names of its contributors
+ *    may be used to endorse or promote products derived from this software
+ *    without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE INSTITUTE AND CONTRIBUTORS ``AS IS'' AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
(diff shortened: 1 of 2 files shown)


jaltman:
The patch was obtained from https://github.com/heimdal/krb5-sync/commit/63780e8a224f122fc53daf3edea79492a636a0da

jaltman:
@lhoward @nicowilliams @vdukhovni please review

lhoward:
I'm confused how this compiles when `kadm5_hook_context` is defined nowhere?

nicowilliams:
@lhoward It doesn't build at all.  It's not even being built.  These are two new files, but the build system doesn't know of them...

lhoward:
OK, I'm going to do this in a new pull request. Will reuse commit message.

PR #430, opened 2018-10-03
hdb-mitdb: Commit 57f1545a broke support of REQUIRES_PWCHANGE during MIT DB imports

fihuer:
Hi,
Here's a small patch to repair some changes made in 57f1545a. It seems like a line has been mistakenly deleted.

Cheers,

Romain


jaltman:
@fihuer thank you

PR #433, opened 2018-10-05
Fix a memory leak in spnego acceptor_start().

yasuoka:


yasuoka:
Previous commit had not fixed the problem.  I pushed updated commit.

nicowilliams:
Did you actually observe a leak?

nicowilliams:
We should make `gss_release_oid()` do nothing -- it should never have existed.  `gss_str_to_oid()` should intern and its outputs never freed (except at library unload time, and arguably we should not support library unload).

nicowilliams:
We're going to close this and instead make sure that `gss_release_oid()` does nothing.  Heimdal doesn't offer `gss_str_to_oid()`, so that's good.

yasuoka:
> Did you actually observe a leak?
Yes, a leak actually happened.

Fixing the problem by #447 will make sense.  Thank for fixing the issue.

PR #434, opened 2018-10-06
Remove an extra line which uses a deleted command.

yasuoka:
There is an extra line in tests/gss/check-context.in.  It uses a deleted command.

### DIFF
--- a/tests/gss/check-context.in
+++ b/tests/gss/check-context.in
@@ -54,7 +54,6 @@ cache="FILE:krb5ccfile"
 
 kinit="${TESTS_ENVIRONMENT} ../../kuser/kinit -c $cache ${afs_no_afslog}"
 klist="${TESTS_ENVIRONMENT} ../../kuser/heimtools klist -c $cache"
-klist="${TESTS_ENVIRONMENT} ../../kuser/klist -c $cache"
 kgetcred="${TESTS_ENVIRONMENT} ../../kuser/kgetcred -c $cache"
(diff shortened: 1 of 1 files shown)


jaltman:
If you can please reference in the commit message the change that removed the command.

yasuoka:
Updated the comment.  I hope it becomes better.

jaltman:
@yasuoka, Thank you for updating the commit message. 

However, I would like to understand when in the repository history resulted in "klist" no longer being available.   "klist" is supposed to be an alias for "heimtools klist".   

@nicowilliams, was the addition of "heimtools klist -c $cache" in 95a2ba6aea supposed to replace the "klist -c $cache" definition?

If so, I would like to see the commit message for this change say

   Remove extraneous definition for "klist" that was supposed to be removed by 95a2ba6aea.

yasuoka:
92a827da9bddaf769724922b04b7f633d2eb0e67 does "klist" => "kcc" and then 2b365b2327e74f9d46de357c1724a878798903f7 does "kcc" => "heimtool".  "klist" still exists in ${bindir}, but it doesn't seem to exist in ${srcdir}/kuser any longer.

I noticed this problem since my first commit of #433 didn't pass the check-context.
In check-context.log:
```
./check-context[140]: ../../kuser/klist: not found
```
It actually caused an error.  Note that this doesn't cause a check failure because
```
    139 # gss_acquire_cred_with_password() must not have side-effects
    140 ${klist} && { eval "$testfailed"; }
```
the test assumes that "klist" command will fail.

PR #439, opened 2018-11-12
kdc: allow checksum of PA-FOR-USER to be HMAC_MD5

iboukris:
even if tgt used an enctype with a different checksum.

Per [MS-SFU] 2.2.1 PA-FOR-USER the checksum is always
HMAC_MD5, and that's what Windows and MIT clients send.

In heimdal both the client and kdc use instead the
checksum of the tgt, and therefore work with each other
but windows and MIT clients fail against heimdal KDC.

Both Windows and MIT KDC would allow any keyed checksum
to be used so Heimdal client work fine against it.

Change Heimdal KDC to allow HMAC_MD5 even for non RC4
based tgt in order to support per-spec clients.

### DIFF
--- a/kdc/krb5tgs.c
+++ b/kdc/krb5tgs.c
@@ -2052,12 +2052,34 @@ server_lookup:
 		goto out;
 	    }
 
-	    ret = krb5_verify_checksum(context,
-				       crypto,
-				       KRB5_KU_OTHER_CKSUM,
-				       datack.data,
-				       datack.length,
-				       &self.cksum);
+	    /* Allow HMAC_MD5 checksum with any key type */
+	    if (self.cksum.cksumtype == CKSUMTYPE_HMAC_MD5) {
+		struct krb5_crypto_iov iov;
+		unsigned char csdata[16];
+		Checksum cs;
+
+		cs.checksum.length = sizeof(csdata);
+		cs.checksum.data = &csdata;
+
+		iov.data.data = datack.data;
+		iov.data.length = datack.length;
+		iov.flags = KRB5_CRYPTO_TYPE_DATA;
+
+		ret = _krb5_HMAC_MD5_checksum(context, NULL, &crypto->key,
+					      KRB5_KU_OTHER_CKSUM, &iov, 1,
+					      &cs);
+		if (ret == 0 &&
+		    krb5_data_ct_cmp(&cs.checksum, &self.cksum.checksum) != 0)
+		    ret = KRB5KRB_AP_ERR_BAD_INTEGRITY;
+	    }
+	    else {
+		ret = krb5_verify_checksum(context,
+					   crypto,
+					   KRB5_KU_OTHER_CKSUM,
+					   datack.data,
+					   datack.length,
+					   &self.cksum);
+	    }
 	    krb5_data_free(&datack);
 	    krb5_crypto_destroy(context, crypto);
 	    if (ret) {
--- a/lib/krb5/libkrb5-exports.def.in
+++ b/lib/krb5/libkrb5-exports.def.in
@@ -781,6 +781,7 @@ EXPORTS
 	_krb5_principalname2krb5_principal
 	_krb5_put_int
 	_krb5_s4u2self_to_checksumdata
+	_krb5_HMAC_MD5_checksum
 	_krb5_expand_path_tokens	;!
 
         ; kinit helper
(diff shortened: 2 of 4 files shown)


kaduk:
Note that RFC 8429 has deprecated HMAC_MD5 for kerberos.
The windows 7 behavior is less interesting in this regard than the behavior of a current windows environment.
(Do you know what version of the Windows KDC you checked the behavior of?)

iboukris:
The Windows KDC version is Windows 2008, I'll try to test on more modern version but the spec hasn't changed so I guess it would behave the same, see:
https://msdn.microsoft.com/en-us/library/cc246089.aspx

MIT client behavior is rather odd, it always sends hmac_md5 checksum, see:
https://github.com/krb5/krb5/blob/0b872730081f01d0d2b6f78675bb5c74604cef62/src/lib/krb5/krb/s4u_creds.c#L175

However when it gets the response, in case of AES enctype, it would ignore it and insist on finding PA_S4U_X509_USER which is an alternative padata to perform protocol transition (see verify_s4u2self_reply()).

Presumably, if we implement support for PA_S4U_X509_USER method in heimdal KDC both MIT and windows clients will practically work, using that padata.

iboukris:
I found why the MIT client expects a PA_S4U_X509_USER, quote from [MS-SFU] 2.2.2 PA_S4U_X509_USER:
https://msdn.microsoft.com/en-us/library/cc246091.aspx

checksum: This is the Kerberos checksum (as defined in [RFC3961]) computed over the DER encoding of the ASN.1 type S4UUserID contained in the user-id field that immediately precedes this field. The key used is the session key of the TGT used in the TGS request (note that the same key is used in the TGS request and reply when this padata is used in both the request and the reply); the checksum operation is the required checksum for the encryption type of that TGT session key per [RFC3961]; and the key usage is 26.

Because there is no required checksum type defined for the encryption type RC4_HMAC_NT (23), if the key's encryption type is RC4_HMAC_NT (23) the checksum type is rsa-md4 (2) as defined in section 6.2.6 of [RFC3961]. If the encryption type is "not-newer" (note that the term "not-newer" is described in section 1 of [RFC4121]), a padata element of type 130 is included in the encrypted-pa-data field of the reply (note that the encrypted-pa-data field is described in appendix A of [Referrals]). The padata of type 130 in the encrypted-pa-data field contains the checksum value in the S4U request concatenated with the checksum value in the S4U reply. The checksum value of a Kerberos Checksum type here refers to the OCTET STRING of the Checksum field. The client when receiving this padata type in the encrypted-pa-data field MUST verify the checksum values match with the corresponding checksum values in the request and the reply.

iboukris:
I closed this back then due to #555 as we found that bug while working on this issue. I need to think this issue again. Just re-opening for now.

iboukris:
CC: @abartlet 

I tested an updated Windows KDC and it now rejects unkeyed checksum with KRB_AP_ERR_MODIFIED. However, it still does not require the checksum to be HMAC_MD5 as per spec, which is nice as we don't have to change Heimdal client, and they'll still work against updated Windows.

I've updated the commit and simplified it a little, please take a look.

kdc/krb5tgs.c
```diff
@@ -2052,12 +2052,34 @@ tgs_build_reply(krb5_context context,
...
+		cs.checksum.length = sizeof(csdata);
+		cs.checksum.data = &csdata;
+
+		iov.data.data = datack.data;
+		iov.data.length = datack.length;
+		iov.flags = KRB5_CRYPTO_TYPE_DATA;
+
+		ret = _krb5_HMAC_MD5_checksum(context, NULL, &crypto->key,
+					      KRB5_KU_OTHER_CKSUM, &iov, 1,
+					      &cs);
```
jaltman: For the benefit of reviewers, the only failure from _krb5_HMAC_MD5_checksum() would be ENOMEM.

iboukris:
I fixed the typo in the commit message.

iboukris:
Updated commit message to mention that MIT KDC would also allow any keyed checksum (per code).

lhoward:
> Updated commit message to mention that MIT KDC would also allow any keyed checksum (per code).

Yes, although git blame says that Sam committed that change, I think that comes from the Novell changes I made for XAD – I was careful about this particular issue ;)

jaltman:
@lhoward and I believe the change is correct but we wonder if this could be implemented without exporting _krb5_HMAC_MD5_checksum() but that can be implemented as a follow on commit.

jaltman:
This change cannot be cherry-picked to heimdal-7-1-branch as is because of the implementation changes to _krb5_HMAC_MD5_checksum() as a result of ca756f0f7fbfba9483c9a1a5ae58f34d96a5e7bb and e50faea7f01738f4638e21f820c12619d64f0ee8.

iboukris:
My initial patch was against the old code (for samba).
I've ported back the commit for 7.1 branch in #576 if that's okay.

PR #442, opened 2018-11-30
Fix assorted typos

ojwb:
There are some further uses of "inital"/"initalize" in changelog files and old NEWS entries which I left for now as some projects are reluctant to fix typos in such files.  Also some in `lib/sqlite/` but that seems to be an embedded code copy.


jaltman:
This change includes chunks that are whitespace only.   In order to make "git blame" be more useful we avoid whitespace only changes.

Issue #416, opened 2018-09-13
crossbuilding on some hosts does not find curses or ncurses


lucianm:
Hello,

when trying to crossbuild heimdal-7.5.0 on an up to date Gentoo-based distribution (Sabayon), I encountered the problem described in [this pull request at LibreELEC](https://github.com/LibreELEC/LibreELEC.tv/pull/2970) which I was able to solve by adding a dependency to host-ncurses, as suggested in an older report which is also mentioned in that conversation. Seems to be an easy fix for somebody familiar with heimdal's code and build system.

Greetings,
Lucian

dfranusic:
I had the same problem and it is due to **getent** being in **libtinfo.so** and not in **libcurses.so** or **libncurses.so**. I made a quick fix in **configure.ac** by adding **tinfo** to **AC_FIND_FUNC_NO_LIBS** macro that checks for **tgetent** 
```
AC_FIND_FUNC_NO_LIBS(tgetent, termcap ncurses curses tinfo,[
#ifdef HAVE_TERMCAP_H
#include <termcap.h>
#endif
#ifdef HAVE_CURSES_H
#include <curses.h>
#endif
],[0,0])
```

lhoward:
This appears to have been fixed @jaltman, can we close?

Issue #398, opened 2018-08-16
Missing make headers


timokau:
Since the 7.5.0 update, the [nix](https://github.com/nixos/nixpkgs) package applies a patch to heimdal that adds two headers to the makefile:

```
--- a/lib/hx509/Makefile.am 2018-03-21 15:41:38.622968809 +0100
+++ b/lib/hx509/Makefile.am 2018-03-21 15:41:32.655162197 +0100
@@ -9,6 +9,8 @@
	sel-gram.h			\
	$(gen_files_ocsp:.x=.c)		\
	$(gen_files_pkcs10:.x=.c)	\
+	ocsp_asn1.h			\
+	pkcs10_asn1.h			\
	hx509_err.c			\
	hx509_err.h
```

Without this patch, the build fails:

```
  #include <ocsp_asn1.h>
            ^~~~~~~~~~~~~
  compilation terminated.
  make[3]: *** [Makefile:1321: hxtool.o] Error 1
  make[3]: Leaving directory '/build/source/lib/hx509'
  make[2]: *** [Makefile:1099: all] Error 2
  make[2]: Leaving directory '/build/source/lib/hx509'
  make[1]: *** [Makefile:564: all-recursive] Error 1
  make[1]: Leaving directory '/build/source/lib'
  make: *** [Makefile:611: all-recursive] Error 1
```

jaltman:
It is unclear from your report.  Is the build of Heimdal 7.5.0 itself failing or of nix?

On which operating system is the failure?

timokau:
`nix` is a package manager that includes Heimdal. The build receipt includes that patch which I just noticed when trying to fix another error. Building without the patch fails. The OS is `NixOS`.

antekone:
Without this patch the build also fails on ArchLinux. Adding those two lines fixes the build.

lhoward:
Believe this may have been fixed by 572a6fd7, can you please test with master and reopen if necessary?

timokau:
I can confirm that this issue is fixed. Thanks!

lhoward:
Great! Thanks for checking!

PR #444, opened 2018-12-09
Build system bugfixes

### LINKED ISSUE #435: hcrypto/engine.h: No such file or directory
> The problem is that the makefile dependencies do not ensure that this 
>  happens before files requiring it in ./include/hcrypto/ get built.

See http://bugs.debian.org/906623 for full details

dfranusic:
Couple of build system fixes that should resolve some known issues on Gentoo and other distributions
* Fixes #435 #398 #416

### DIFF
--- a/lib/hx509/hx_locl.h
+++ b/lib/hx509/hx_locl.h
@@ -63,8 +63,8 @@
 #include <pkcs8_asn1.h>
 #include <pkcs9_asn1.h>
 #include <pkcs12_asn1.h>
-#include <ocsp_asn1.h>
-#include <pkcs10_asn1.h>
+#include <ocsp_asn1.hx>
+#include <pkcs10_asn1.hx>
 #include <asn1_err.h>
 #include <pkinit_asn1.h>
 
--- a/configure.ac
+++ b/configure.ac
@@ -453,7 +453,7 @@ AC_FIND_FUNC_NO_LIBS(openpty, util,[
 #endif
 ],[0,0,0,0,0])
 
-AC_FIND_FUNC_NO_LIBS(tgetent, termcap ncurses curses,[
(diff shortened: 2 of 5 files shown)


lhoward:
Ah yes I think I was running to this on Ubuntu

lib/hx509/hx_locl.h
```diff
@@ -63,8 +63,8 @@
 #include <pkcs8_asn1.h>
 #include <pkcs9_asn1.h>
 #include <pkcs12_asn1.h>
-#include <ocsp_asn1.h>
-#include <pkcs10_asn1.h>
+#include <ocsp_asn1.hx>
```
nicowilliams: Why is this needed?

nicowilliams: The rest LGTM.

lhoward: It's really a dependency issue, I think the fix is this

```
diff --git a/lib/hx509/Makefile.am b/lib/hx509/Makefile.am
index b58deb3e3..09643c43a 100644
--- a/lib/hx509/Makefile.am
+++ b/lib/hx509/Makefile.am
@@ -164,7 +164,7 @@ hxtool-commands.c hxtool-commands.h: hxtool-commands.in $(SLC)
 dist_hxtool_SOURCES = hxtool.c
 nodist_hxtool_SOURCES = hxtool-commands.c hxtool-commands.h
 
-$(hxtool_OBJECTS): hxtool-commands.h hx509_err.h
+$(hxtool_OBJECTS): hxtool-commands.h $(nodist_include_HEADERS)
 
 hxtool_LDADD = \
        libhx509.la \
```

lhoward:
I've cherry-picked the fixes you approved and will push shortly. Testing my other fix above now.

PR #449, opened 2018-12-16
PY3: fix "TabError: inconsistent use of tabs and spaces"

abartlet:
From Samba

Python2 is EOL soon and Samba builds with python3 by default now so it would be helpful to have this merged upstream.


jaltman:
@abartlet @noelpower  The python source file lib/wind/get-punycode-examples.py was not converted to Python3.  Could you please update it?

Thank you.

Issue #452, opened 2018-12-16
Request: Support aliases for client principal initial ticket acquisition


jaltman:
Heimdal permits aliases to be assigned to client principals but the alias name cannot be used for initial ticket acquisition.

jaltman:
@nicowilliams has suggested using PA-TGS in AS-REQs

lhoward:
Isn’t this a bug in the implementation of the referrals draft rather than a new feature? I may be missing context.

lhoward:
@jaltman do you want AS-REQ aliases (sans name canonicalisation) to be on by default, or do you want a knob?

jaltman:
I cannot think of a reason why this functionality should be off by default or otherwise.

nicowilliams:
Hard aliases == you use the alias in the `AS-REQ` and you get a TGT for the alias name.

Soft aliases == you use the alias in the `AS-REQ` and you get a TGT for the canonical name.

?

If so, I agree.

In the hard alias case you'd have to get the non-default salt in the `ETYPE-INFO`.  In the soft alias case you get to use the default salt.

There's no referrals codepaths in the AS-REQ case, thank goodness.  That's all TGS stuff and is pretty complex in both, Heimdal and MIT :(

I do like the idea that you can grant one principal the ability to get TGTs for another using `PA-TGS` in an `AS-REQ`.

jaltman:
@nicowilliams using your terminology, the case I'm concerned with is the hard alias case.   I believe the soft alias case requires that the client explicitly request canonicalization.  Today's clients will reject any response with a name other than the AS-REQ specified name unless canonicalization has been requested.

I'm trying to reduce the work for an organization that needs to rename their realm from FOO.COM to BAR.COM .   By adding user@BAR.COM as an alias to the user@FOO.COM client principal entry, the end user should be able to switch to authenticating as user@BAR.COM without requiring any rekeying.

Its then up to the application service administrators to configure their systems to accept one or both of user@FOO.COM and user@BAR.COM at various stages of the transition.

nicowilliams:
@jaltman It might be interesting to have a realm alias feature for that use case -- it would be easier to deploy for your specific use case.  AD has it, sort of, by supporting all case combinations of a realm as hard aliases of the realm.

jaltman:
Something I noticed today is that Heimdal kinit squashes the --canonicalize and --enterprise options together.  Is this correct?

[...]

PR #457, opened 2018-12-19
kdc: Support aliases for client principal initial ticket acquisition (#452)

lhoward:
This patch implements "hard" aliases where to the client an aliased principal is indistinguishable from an unaliased one: it's not required to request name canonicalization and the client is returned the name it asked for. (Note: you need to create a TGS alias, too.)

I've made this a runtime configurable option in this initial implementation: I'm being careful not to change the name canonicalization (NC) behavior as I don't want to break Samba (cc @abartlet). IIRC Windows always sets NC flag in the AS-REQ, so perhaps I'm being needlessly cautious, however there are a couple of edge cases:

- The MIT code always canonicalizes the client realm in the returned ticket (I recall implementing this based on observed Windows behavior). However, for hard aliases to work (where the client expects the client name not to change), we need to return the one the client requested.
- The MIT code always only canonicalizes the server name if the client is asking for a TGT and the canonicalize flag is set. (Again, I remember implementing this based on observed Windows behavior.) This path is required to support NetBIOS names (e.g. krbtgt/SAMBA@SAMBA). For hard aliases (or generally in the case where NC is not set) we should return the server name the client requested.

### DIFF
--- a/lib/hdb/common.c
+++ b/lib/hdb/common.c
@@ -119,6 +119,7 @@ _hdb_fetch_kvno(krb5_context context, HDB *db, krb5_const_principal principal,
 	if (ret)
 	    return ret;
 	principal = enterprise_principal;
+	flags |= HDB_F_CANON; /* enterprise implies canonicalization */
     }
 
     hdb_principal2key(context, principal, &key);
@@ -129,7 +130,7 @@ _hdb_fetch_kvno(krb5_context context, HDB *db, krb5_const_principal principal,
     if(ret)
 	return ret;
     ret = hdb_value2entry(context, &value, &entry->entry);
-    if (ret == ASN1_BAD_ID && (flags & HDB_F_CANON) == 0) {
+    if (ret == ASN1_BAD_ID && (flags & (HDB_F_CANON|HDB_F_FOR_AS_REQ)) == 0) {
 	krb5_data_free(&value);
 	return HDB_ERR_NOENTRY;
     } else if (ret == ASN1_BAD_ID) {
@@ -152,6 +153,19 @@ _hdb_fetch_kvno(krb5_context context, HDB *db, krb5_const_principal principal,
 	if (ret) {
 	    krb5_data_free(&value);
 	    return ret;
+	}
+
+	if ((flags & HDB_F_FOR_AS_REQ) && (flags & HDB_F_CANON) == 0) {
+	    krb5_principal tmp;
+
+	    /* "hard" alias: return the principal the client asked for */
+	    ret = krb5_copy_principal(context, principal, &tmp);
+	    if (ret) {
+		krb5_data_free(&value);
+		return ret;
+	    }
+	    krb5_free_principal(context, entry->entry.principal);
+	    entry->entry.principal = tmp;
 	}
     }
     krb5_data_free(&value);


lhoward:
OK, build/tests aren't passing, stay tuned.

nicowilliams:
Why is this dangerous to turn on without a command-line option for the KDC?

I mean, surely no hard aliases exist now, especially out-of-realm ones.

nicowilliams:
Regarding the `krbtgt/REALM@REALM` (we call this a "starting" TGT name)...  You could just create that for the alias realm, rather than alias the one for the canonical name.  Just sayin', that ought to work and then you don't need to add code for aliasing the realm.  It's just as well, since presumably during migration cross-realm TGTs will be needed between the two.

It's interesting, convenient, and necessary that they'll all be in the same HDB.  That's because kadmind doesn't really support multiple realms unless they are all in the same HDB.  IIRC.

lhoward:
> Why is this dangerous to turn on without a command-line option for the KDC?

It might be OK. Let me check [MS-KILE] and verify Windows always sets the NC flag.

lhoward:
OK, I think it is safe to always enable (see revised patch). @abartlet please check this doesn't regress Samba DCs.

lhoward:
> Regarding the `krbtgt/REALM@REALM` (we call this a "starting" TGT name)... You could just create that for the alias realm, rather than alias the one for the canonical name. Just sayin', that ought to work and then you don't need to add code for aliasing the realm. It's just as well, since presumably during migration cross-realm TGTs will be needed between the two.
> 
> It's interesting, convenient, and necessary that they'll all be in the same HDB. That's because kadmind doesn't really support multiple realms unless they are all in the same HDB. IIRC.

The realm aliasing code was to avoid regressing with Windows, actually; there's no issue with hard aliases (the client gets what it asked for if the NC flag is unset). However I checked [MS-KILE] and the behavior now matches that, even though it's different to what I implemented for MIT. (The effect is more or less the same as far as Windows clients are concerned. My implementation for MIT predated the availability of [MS-KILE] and was based on observation rather than specification.)

lhoward:
Not sure why check-kadmin is failing on one of the Travis builds. Will investigate.

lhoward:
One issue now that HDB_F_CANON is always set for the client lookup, is that a backend database may return KDC_ERR_WRONG_REALM when the name canonicalization flag is unset (instead of KDC_ERR_C_PRINCIPAL_UNKNOWN). This *may* confuse some clients. I haven't yet taken a look at the Samba code to see if it honors this expected behavior.

lhoward:
OK, a quick look at the Samba4 code suggests they don't check HDB_F_CANON anywhere. So they're not going to be impacted by this change in behavior. Phew! (I mean, that probably is a bug in Samba but... sounds like no one has run into it yet.)

nicowilliams:
Wrong realm can always get mapped to client principal unknown...

lhoward:
Right, that would actually fall out of omitting the wrong realm check if f.canonicalize == 0. But that’s ostensibly fixing a problem we don’t have... maybe nice to do to be correct?

lhoward:
cc @abartlet 

This patch breaks Samba's test suite, which tests for canonicalization behavior identical to Windows (i.e. the realm is always canonicalized irrespective of the setting of the flag in the AS-REQ).

So @jaltman there are a few possibilities here for implementing the original feature request:

- You upgrade your clients to send the canonicalization flag (i.e. soft canonicalization as described in the feature request)
- We push the functionality into the backend (but not Samba's, obviously). HDB_F_CANON would (as it does today) reflect the canonicalize flag in the client AS-REQ. The backend would always lookup aliases irrespective of the setting of this flag, and if HDB_F_CANON was *not* set would return the principal the KDC requested rather than the canonical principal. 
- Runtime knob on the KDC
- Merge this PR, breaking Samba's test suite (but in theory still being compliant with [MS-KILE] as it doesn't appear to require that the AS-REP client name be canonicalized if the flag is unset)

jaltman:
@lhoward I don't want option 1.  If @abartlet is willing to updated Samba's test suite then I would prefer option 4.  Otherwise option 2.

lhoward:
I suspect it's going to be option 2 then, and leave the KDC untouched. @abartlet you should take care of that kadmin/changepw issue I mentioned in your code then.

lhoward:
Option 2 implemented, please review.

nicowilliams:
Option 2, yes.

lhoward:
Arguably there are some nicer things we could do in the AS for canonicalization but, as long as Samba is handling returning whichever principal they need to by rewriting the returned HDB entry, it's probably fine to leave as-is.

lib/hdb/common.c
```diff
@@ -153,6 +154,23 @@ _hdb_fetch_kvno(krb5_context context, HDB *db, krb5_const_principal principal,
 	    krb5_data_free(&value);
 	    return ret;
 	}
+
+	if ((flags & HDB_F_FOR_AS_REQ) && (flags & HDB_F_CANON) == 0) {
+	    krb5_principal tmp;
+
+	    /*
+	     * Return the name the client asked for. Should be done in the KDC,
```
nicowilliams: I actually like this being done in `libhdb`, honestly.  We already have a bunch of KDC specials, such as if the caller requests a specific kvno then `libhdb` puts that in the `keys` field so the caller doesn't have to go looking through key history (OK, I did that), but that's good if you posit a very normalized backend schema and you don't want to have to pull in things the caller doesn't need.

lib/hdb/common.c
```diff
@@ -153,6 +154,23 @@ _hdb_fetch_kvno(krb5_context context, HDB *db, krb5_const_principal principal,
 	    krb5_data_free(&value);
 	    return ret;
 	}
+
+	if ((flags & HDB_F_FOR_AS_REQ) && (flags & HDB_F_CANON) == 0) {
+	    krb5_principal tmp;
```
nicowilliams: You don't need a `tmp` variable...

lib/hdb/common.c
lhoward: I'm with you 50/50. I kind of like it too. But, I also like the idea of normalizing the AS path so that it's identical to the TGS path: if you ask for a principal with name canonicalization disabled, the name in the reply is the one you asked for, and this is done by the KDC.

lib/hdb/common.c
```diff
@@ -153,6 +154,23 @@ _hdb_fetch_kvno(krb5_context context, HDB *db, krb5_const_principal principal,
...
+	     * Return the name the client asked for. Should be done in the KDC,
+	     * but the interactions with the Samba backend would be complex, so
+	     * do it here.
+	     */
+	    ret = krb5_copy_principal(context, principal, &tmp);
+	    if (ret) {
+		krb5_data_free(&value);
+		return ret;
+	    }
+	    krb5_free_principal(context, entry->entry.principal);
```
nicowilliams: Do the `krb5_free_principal(context, entry->entry.principal)` first then the `krb5_copy_principal(context, principal, &entry->entry.principal)`, then you can do without a `tmp` variable.

lib/hdb/common.c
lhoward: Wanted to be sure allocation succeeded before swapping in the new principal.

lib/hdb/common.c
lhoward: Yes, but if krb5_copy_principal fails, then we've killed the principal. Which I reckon smells funny, and may impact logging too I suppose.

nicowilliams:
LGTM to me.  I'd rather get rid of that `tmp` var as it will reduce the amount of vertical code and whitespace, but it's a very minor thing.

lhoward:
> LGTM to me. I'd rather get rid of that `tmp` var as it will reduce the amount of vertical code and whitespace, but it's a very minor thing.

OK, I'm happy to do this.

lhoward:
Or I could just make the comment smaller ;)

nicowilliams:
Porqué no los dos?

nicowilliams:
Oh, but if you remove the `tmp` var, don't forget to set `entry->entry.principal` to `NULL` if the copy fails.  If you'd rather keep the `tmp` var, that's fine.

abartlet:
Thanks @lhoward for reaching out to me on the implications for Samba.  

As background for @jaltman the reason we can't change the testsuite is that we also test against Windows, and while the behaviour might never be seen in the wild, our testsuite tries it anyway to be sure.

I'll re-test the final patch tomorrow, but I can't see how it would harm us. 

Thanks!

Issue #166, opened 2016-03-29
MIT Interop - Heimdal Kerberos does not work with the MIT KEYRING type of credentials cache


quanah:
Heimdal linked cyrus-sasl GSSAPI modules fail to work with many RHEL based MIT Kerberos systems due to the fact that RHEL defaults to using the KEYRING type credential cache (More at [http://web.mit.edu/kerberos/krb5-1.12/doc/basic/ccache_def.html](http://web.mit.edu/kerberos/krb5-1.12/doc/basic/ccache_def.html)).  This is a Linux specific method.  The lack of ability to interoperate is of course rather frustrating to admins.  It would be extremely useful for the next release of Heimdal to include support for KEYRING credential caches.

nicowilliams:
Sure, we should interop with the MIT `KEYRING` ccache type.  Does anyone have patches for this?

quanah:
I don't. :/  I just recently encountered this when trying to get a client compiled against MIT Krb (ldapsearch) to work with a server compiled with Heimdal Krb (OpenLDAP's slapd linked to cyrus-sasl that was linked to heimdal).

quanah:
Note: breaks interop with default RHEL configs

jaltman:
@quanah sure.  but someone needs to write the necessary code.   If there is an organization that wants this functionality, they can provide code or money.

quanah:
Yeah, I was just mainly trying to update the title of the bug so that people could find it easily if they hit the issue, but that change seems to have gotten lost. :/

quanah:
There, fixed that now. ;)

prmarino1:
FYI
there may be a workaround for this with MIT Kerberos 1.13 or higher.
the is from the release for MIT Kerberos 1.13 released in October of 2014
"Add client support for the Kerberos Cache Manager protocol. If the host is running a Heimdal kcm daemon, caches served by the daemon can be accessed with the KCM: cache type."
while it does not solve the issue it is a viable work around.

nicowilliams:
I happen to be working on an extension to the `FILE` ccache that stores a large hash table in a `ccconfig` entry.  The hash table is O(1) *and* lock-less for both, reading *and* writing.  The thought occurs that when I'm done with this, we can use that for the `KEYRING` ccache type...

quanah:
@nicowilliams that's awesome!

jaltman:
@lhoward is implementing KEYRING ccache support for Heimdal work that would interest you?

[...]

PR #459, opened 2018-12-21
krb5: port MIT Linux keyring credentials cache (#166)

lhoward:
Initial revision for review, needs tests.

### DIFF
--- a/lib/krb5/krcache.c
+++ b/lib/krb5/krcache.c
@@ -1,0 +1,2034 @@
+/*
+ * Copyright (c) 2006 The Regents of the University of Michigan.
+ * All rights reserved.
+ *
+ * Portions Copyright (c) 2018, AuriStor, Inc.
+ *
+ * Permission is granted to use, copy, create derivative works
+ * and redistribute this software and such derivative works
+ * for any purpose, so long as the name of The University of
+ * Michigan is not used in any advertising or publicity
+ * pertaining to the use of distribution of this software
+ * without specific, written prior authorization.  If the
+ * above copyright notice or any other identification of the
+ * University of Michigan is included in any copy of any
+ * portion of this software, then the disclaimer below must
+ * also be included.
+ *
+ * THIS SOFTWARE IS PROVIDED AS IS, WITHOUT REPRESENTATION
+ * FROM THE UNIVERSITY OF MICHIGAN AS TO ITS FITNESS FOR ANY
+ * PURPOSE, AND WITHOUT WARRANTY BY THE UNIVERSITY OF
+ * MICHIGAN OF ANY KIND, EITHER EXPRESS OR IMPLIED, INCLUDING
+ * WITHOUT LIMITATION THE IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE. THE
+ * REGENTS OF THE UNIVERSITY OF MICHIGAN SHALL NOT BE LIABLE
+ * FOR ANY DAMAGES, INCLUDING SPECIAL, INDIRECT, INCIDENTAL, OR
+ * CONSEQUENTIAL DAMAGES, WITH RESPECT TO ANY CLAIM ARISING
+ * OUT OF OR IN CONNECTION WITH THE USE OF THE SOFTWARE, EVEN
+ * IF IT HAS BEEN OR IS HEREAFTER ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGES.
+ */
+/*
+ * Copyright 1990,1991,1992,1993,1994,2000,2004 Massachusetts Institute of
+ * Technology.  All Rights Reserved.
+ *
+ * Original stdio support copyright 1995 by Cygnus Support.
+ *
+ * Export of this software from the United States of America may
+ *   require a specific license from the United States Government.
+ *   It is the responsibility of any person or organization contemplating
+ *   export to obtain such a license before exporting.
+ *
+ * WITHIN THAT CONSTRAINT, permission to use, copy, modify, and
+ * distribute this software and its documentation for any purpose and
+ * without fee is hereby granted, provided that the above copyright
+ * notice appear in all copies and that both that copyright notice and
+ * this permission notice appear in supporting documentation, and that
+ * the name of M.I.T. not be used in advertising or publicity pertaining
+ * to distribution of the software without specific, written prior
+ * permission.  Furthermore if you modify this software you must label
+ * your software as modified software and not distribute it in such a
+ * fashion that it might be confused with the original M.I.T. software.
+ * M.I.T. makes no representations about the suitability of
+ * this software for any purpose.  It is provided "as is" without express
+ * or implied warranty.
+ */
+
+/*
+ * This file implements a collection-enabled credential cache type where the
+ * credentials are stored in the Linux keyring facility.
+ *
+ * A residual of this type can have three forms:
+ *    anchor:collection:subsidiary
+ *    anchor:collection
+ *    collection
+ *
+ * The anchor name is "process", "thread", or "legacy" and determines where we
+ * search for keyring collections.  In the third form, the anchor name is
+ * presumed to be "legacy".  The anchor keyring for legacy caches is the
+ * session keyring.
+ *
+ * If the subsidiary name is present, the residual identifies a single cache
+ * within a collection.  Otherwise, the residual identifies the collection
+ * itself.  When a residual identifying a collection is resolved, the
+ * collection's primary key is looked up (or initialized, using the collection
+ * name as the subsidiary name), and the resulting cache's name will use the
+ * first name form and will identify the primary cache.
+ *
+ * Keyring collections are named "_krb_<collection>" and are linked from the
+ * anchor keyring.  The keys within a keyring collection are links to cache
+ * keyrings, plus a link to one user key named "krb_ccache:primary" which
+ * contains a serialized representation of the collection version (currently 1)
+ * and the primary name of the collection.
+ *
+ * Cache keyrings contain one user key per credential which contains a
+ * serialized representation of the credential.  There is also one user key
+ * named "__krb5_princ__" which contains a serialized representation of the
+ * cache's default principal.
+ *
+ * If the anchor name is "legacy", then the initial primary cache (the one
+ * named with the collection name) is also linked to the session keyring, and
+ * we look for a cache in that location when initializing the collection.  This
+ * extra link allows that cache to be visible to old versions of the KEYRING
+ * cache type, and allows us to see caches created by that code.
+ */
+
+#include "krb5_locl.h"
+
+#ifdef HAVE_KEYUTILS_H
+
+#include <keyutils.h>
+
+/*
+ * We try to use the big_key key type for credentials except in legacy caches.
+ * We fall back to the user key type if the kernel does not support big_key.
+ * If the library doesn't support keyctl_get_persistent(), we don't even try
+ * big_key since the two features were added at the same time.
+ */
+#ifdef HAVE_KEYCTL_GET_PERSISTENT
+#define KRCC_CRED_KEY_TYPE		"big_key"
+#else
+#define KRCC_CRED_KEY_TYPE		"user"
+#endif
+
+/*
+ * We use the "user" key type for collection primary names, for cache principal
+ * names, and for credentials in legacy caches.
+ */
+#define KRCC_KEY_TYPE_USER		"user"
+
+/*
+ * We create ccaches as separate keyrings
+ */
+#define KRCC_KEY_TYPE_KEYRING		"keyring"
+
+/*
+ * Special name of the key within a ccache keyring
+ * holding principal information
+ */
+#define KRCC_SPEC_PRINC_KEYNAME		"__krb5_princ__"
+
+/*
+ * Special name for the key to communicate the name(s)
+ * of credentials caches to be used for requests.
+ * This should currently contain a single name, but
+ * in the future may contain a list that may be
+ * intelligently chosen from.
+ */
+#define KRCC_SPEC_CCACHE_SET_KEYNAME	"__krb5_cc_set__"
+
+/*
+ * This name identifies the key containing the name of the current primary
+ * cache within a collection.
+ */
+#define KRCC_COLLECTION_PRIMARY		"krb_ccache:primary"
+
+/*
+ * If the library context does not specify a keyring collection, unique ccaches
+ * will be created within this collection.
+ */
+#define KRCC_DEFAULT_UNIQUE_COLLECTION	"session:__krb5_unique__"
+
+/*
+ * Collection keyring names begin with this prefix.  We use a prefix so that a
+ * cache keyring with the collection name itself can be linked directly into
+ * the anchor, for legacy session keyring compatibility.
+ */
+#define KRCC_CCCOL_PREFIX		"_krb_"
+
+/*
+ * For the "persistent" anchor type, we look up or create this fixed keyring
+ * name within the per-UID persistent keyring.
+ */
+#define KRCC_PERSISTENT_KEYRING_NAME	"_krb"
+
+/*
+ * Name of the key holding time offsets for the individual cache
+ */
+#define KRCC_TIME_OFFSETS		"__krb5_time_offsets__"
+
+/*
+ * Keyring name prefix and length of random name part
+ */
+#define KRCC_NAME_PREFIX		"krb_ccache_"
+#define KRCC_NAME_RAND_CHARS		8
+
+#define KRCC_COLLECTION_VERSION		1
+
+#define KRCC_PERSISTENT_ANCHOR		"persistent"
+#define KRCC_PROCESS_ANCHOR		"process"
+#define KRCC_THREAD_ANCHOR		"thread"
+#define KRCC_SESSION_ANCHOR		"session"
+#define KRCC_USER_ANCHOR		"user"
+#define KRCC_LEGACY_ANCHOR		"legacy"
+
+#if SIZEOF_KEY_SERIAL_T != 4
+/* lockless implementation assumes 32-bit key serials */
+#error only 32-bit key serial numbers supported by this version of keyring ccache
+#endif
+
+typedef union _krb5_krcache_and_princ_id {
+    uint64_t krcu_cache_and_princ_id;
+    struct {
+	key_serial_t cache_id;		/* keyring ID representing ccache */
+	key_serial_t princ_id;		/* key ID holding principal info */
+    } krcu_id;
+    #define krcu_cache_id		krcu_id.cache_id
+    #define krcu_princ_id		krcu_id.princ_id
+} krb5_krcache_and_princ_id;
+
+/*
+ * This represents a credentials cache "file" where cache_id is the keyring
+ * serial number for this credentials cache "file".  Each key in the keyring
+ * contains a separate key.
+ *
+ * Thread-safe as long as caches are not destroyed whilst other threads are
+ * using them.
+ */
+typedef struct _krb5_krcache {
+    char *krc_name;			/* Name for this credentials cache */
+    krb5_timestamp krc_changetime;	/* update time, does not decrease (mutable) */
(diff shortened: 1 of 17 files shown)


nicowilliams:
@lhoward how much did you change ` lib/krb5/krcache.c` relative to MIT's?  I've not reviewed that file yet, but the rest LGTM.

lib/krb5/krcache.c
```diff
@@ -0,0 +1,1994 @@
...
+#ifdef DEBUG
+#define KRCC_DEBUG	  1
+#endif
+
+#if KRCC_DEBUG
+void debug_print(char *fmt, ...);       /* prototype to silence warning */
+#include <syslog.h>
+#define DEBUG_PRINT(x) debug_print x
+void
+debug_print(char *fmt, ...)
```
nicowilliams: Heimdal has a `_krb5_debug()`.  Can we use that?

lhoward:
> @lhoward how much did you change ` lib/krb5/krcache.c` relative to MIT's? I've not reviewed that file yet, but the rest LGTM.

- stylistic (in progress, at the risk of making it diverge I want to clean up the "MIT-isms")
- use krb5_storage for serialization rather than hand-rolled
- implement cc_move() (was not implemented by MIT, needed by Heimdal kinit, unfortunately impossible to do atomically as it there's no API to rename a keyring)

lib/krb5/krcache.c
```diff
@@ -0,0 +1,1994 @@
...
+    vsyslog(LOG_ERR, fmt, ap);
+#endif
+    va_end(ap);
+}
+#else
+#define DEBUG_PRINT(x)
+#endif
+
+/*
+ * We try to use the big_key key type for credentials except in legacy caches.
```
nicowilliams: I guess we need the legacy stuff too :(

lhoward:
Read/write interop seems to be fine with MIT, including support for multiple ccaches.

lib/krb5/krcache.c
lhoward: Yeah, that's the default type.

lib/krb5/krcache.c
lhoward: fixed

lib/krb5/krcache.c
```diff
@@ -0,0 +1,1994 @@
...
+	 * keyring for that uid.  We link it to the process keyring to ensure
+	 * that we have possession rights on the collection key.
+	 */
+	if (*collection_name != '\0') {
+	    errno = 0;
+	    uidnum = strtol(collection_name, &cnend, 10);
+	    if (errno || *cnend != '\0')
+		return KRB5_KCC_INVALID_UID;
+	} else {
+	    uidnum = geteuid();
```
nicowilliams: In Heimdal (and MIT) we use `getuid()`, not `geteuid()`, to name default `FILE` ccaches, don't we?  Hmm, MIT supports ruid and euid in token expansion while Heimdal only supports ruid in token expansion, but both use the ruid for the compiled-in default `FILE` ccache name.

Should this be using `geteuid()`?  Strikes me as at least weird, if not outright wrong.

lhoward: Also, I was wondering if `get_persistent_fallback()` should be using `issuid()` (which nothing else appears to use?)

lib/krb5/krcache.c
```diff
@@ -0,0 +1,1994 @@
...
+
+/*
+ * GET_PERSISTENT(uid) acquires the persistent keyring for uid, or falls back
+ * to the user keyring if uid matches the current effective uid.
+ */
+
+static key_serial_t
+get_persistent_fallback(uid_t uid)
+{
+    return (uid == geteuid()) ? KEY_SPEC_USER_KEYRING : -1;
```
nicowilliams: I don't quite grok why `geteuid()` instead of `getuid()` here.

lib/krb5/krcache.c
lhoward: Both implementations use `FILE:/tmp/krb5cc_%{uid}` as the default ccache name. @greghudson do you have any thoughts on whether the default persistent keyring cache should use ruid or euid?

lib/krb5/krcache.c
```diff
@@ -0,0 +1,1994 @@
...
+static krb5_error_code
+set_primary_name(krb5_context context, key_serial_t collection_id,
+		 const char *subsidiary_name)
+{
+    krb5_error_code ret;
+    krb5_storage *sp;
+    krb5_data payload;
+    key_serial_t key;
+
+    sp = krb5_storage_emem();
```
nicowilliams: Thanks for doing this.

lib/krb5/krcache.c
lhoward: No idea (yet). Perhaps @greghudson can enlighten?

lib/krb5/krcache.c
nicowilliams: thx!

lib/krb5/krcache.c
nicowilliams: Oh?  :(

lib/krb5/krcache.c
nicowilliams: `rk_secure_getenv()` uses `issuid()`.

lhoward:
My understanding is that big_key support requires a kernel module which may or may not be present.

lib/krb5/krcache.c
nicowilliams: It's a tricky thing.  You might say that you don't want privileged processes even reading and parsing user files/data like Kerberos credentials caches, as any vulnerability in the parsing code could be exploited.  This is why Heimdal refuses to read user ccaches when running as root.  But also these libraries should not use the effective UID for finding default credentials -- finding credentials in a set-uid context should require explicit _intent_.  MIT can do what they do, but for us I think the appropriate thing would be to do for the `KEYRING` cache what we do for the `FILE` cache.

lib/krb5/krcache.c
```diff
@@ -0,0 +1,1972 @@
...
+
+cleanup:
+    free(payload);
+    free(subsidiary_name);
+
+    return ret;
+}
+
+/*
+ * Note: MIT keyring code uses krb5int_random_string() as if the second argument
```
nicowilliams: Hmm?  Near as I can tell the MIT keyring code passes `KRCC_NAME_RAND_CHARS` (which is 8) as the `slen` argument, and that's taken to be a buffer size in bytes, with the last being set to `'\0'`.

lib/krb5/krcache.c
```diff
@@ -0,0 +1,1972 @@
...
+    p = malloc(slen);
+    if (p == NULL)
+	return krb5_enomem(context);
+
+    krb5_generate_random_block(p, slen);
+
+    for (i = 0; i < slen; i++)
+	s[i] = chars[p[i] % (sizeof(chars) - 1)];
+
+    s[i] = '\0';
```
nicowilliams: This would be a buffer overflow by one byte here because the buffer passed in won't have room for this `NUL`.

```
     char uniquename[sizeof(KRCC_NAME_PREFIX) + KRCC_NAME_RAND_CHARS];
...
    memcpy(uniquename, KRCC_NAME_PREFIX, sizeof(KRCC_NAME_PREFIX));
...
    tries = 100;
    while (tries-- > 0) {
	ret = generate_random_string(context, uniquename + prefixlen,
				     KRCC_NAME_RAND_CHARS);
```

That's why MIT's `krb5int_random_string()` starts like this:

```
    krb5_error_code
    krb5int_random_string(krb5_context context, char *string, unsigned int length)
    {
        static const unsigned char charlist[] =
            "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ";
        krb5_error_code err = 0;
        unsigned char *bytes = NULL;
------->unsigned int bytecount = length - 1;
```

lib/krb5/krcache.c
lhoward: But the containing variable (from memory) is sizeof(prefix) + KRCC_NAME_RAND_CHARS which suggests to me that it’s a character count not a buffer size.

lib/krb5/krcache.c
```diff
@@ -0,0 +1,1972 @@
...
+    *subsidiary_out = NULL;
+    *cache_id_out = 0;
+
+    memcpy(uniquename, KRCC_NAME_PREFIX, sizeof(KRCC_NAME_PREFIX));
+
+    HEIMDAL_MUTEX_lock(&krcc_mutex);
+
+    /* Loop until we successfully create a new ccache keyring with
+     * a unique name, or we get an error. Limit to 100 tries. */
+    tries = 100;
```
nicowilliams: 100!  Fine.  Whatever.  I assume this is a pretty fast operation anyways.

lib/krb5/krcache.c
nicowilliams: I'm pretty sure they mean "CHARS" as in "char", not as in "characters".  There's no string yet whose Unicode (or whatever) characters to count -- it can only be bytes.

nicowilliams: BTW, thanks for this comment, it made me look more closely.

lib/krb5/krcache.c
lhoward: OK, my bad. I’ll take a look tomorrow.

lib/krb5/krcache.c
```diff
@@ -0,0 +1,1972 @@
...
+    char uniquename[sizeof(KRCC_NAME_PREFIX) + KRCC_NAME_RAND_CHARS];
+    int prefixlen = sizeof(KRCC_NAME_PREFIX) - 1;
+    int tries;
+
+    *subsidiary_out = NULL;
+    *cache_id_out = 0;
+
+    memcpy(uniquename, KRCC_NAME_PREFIX, sizeof(KRCC_NAME_PREFIX));
+
+    HEIMDAL_MUTEX_lock(&krcc_mutex);
```
nicowilliams: What does this lock protect?  I know, it's in the original from MIT.  But no global or shared state is being touched, and the `add_key()` operation too involves non-shared state (we're creating a unique, randomized ccache name, after all -- it can't be shared!).

IMO we don't need this lock and neither does MIT.

@greghudson might have some insight?

lib/krb5/krcache.c
lhoward: Are you sure? Doesn’t size of prefix include room for terminating NUL? Afk so ability to diagnose limited.

lib/krb5/krcache.c
lhoward: Agreed, not a fan of magic numbers... but am a fan of code that’s been tested by other people!

lib/krb5/krcache.c
lhoward: Maybe they originally used a counter or something? You’re right, it doesn’t make sense. I’ll delete it unless Greg has some ideas.

lib/krb5/krcache.c
nicowilliams: @greghudson 100 is a bit much, no?  What are the chances that we'd conflict N times trying to create keyrings with 8 randomly selected printable non-space characters -- let's say 62 chars?  Yeah, minute.  N=5 is plenty good enough.

greghudson: The pre-collection MIT code loops indefinitely, and that's probably better than having a fixed number of tries, since as Nico notes it's very unlikely that we will see even one conflict during the race window.  I don't remember if it was Simo's idea or mine to add a fixed limit.  From my current perspective it seems inelegant but harmless.

lib/krb5/krcache.c
```diff
@@ -0,0 +1,1972 @@
...
+    ret = krb5_storage_to_data(sp, &buf);
+    if (ret) {
+	krb5_storage_free(sp);
+	return ret;
+    }
+
+    krb5_storage_free(sp);
+
+    /* Add new key into keyring */
+#ifdef KRCC_DEBUG
```
nicowilliams: You can un-ifdef this out now.

lib/krb5/krcache.c
```diff
@@ -0,0 +1,1972 @@
...
+    sp = krb5_storage_from_data(&payload);
+    if (sp == NULL) {
+	ret = KRB5_CC_IO;
+	goto fail;
+    }
+
+    krb5_storage_set_byteorder(sp, KRB5_STORAGE_BYTEORDER_BE);
+
+    ret = krb5_ret_int32(sp, sec_offset);
+    ret |= krb5_ret_int32(sp, usec_offset);
```
nicowilliams: Don't use `|=` here.  Just say `if (ret == 0) ret = krb5_ret_int32(sp, usec_offset);`.

lib/krb5/krcache.c
lhoward: Otherwise uniquename wastes a byte always, no?

lib/krb5/krcache.c
lhoward: Happy to change to loop indefinitely.

lib/krb5/krcache.c
lhoward: Noted, was being lazy!

lib/krb5/krcache.c
nicowilliams: I'd rather have a definite N here than indefinite.

lib/krb5/krcache.c
nicowilliams: I mean, it needs a byte for a `NUL` because it has to be a C string.  Call it a waste if you like :)  Or maybe I didn't get the comment.

nicowilliams:
All my review comments are in.

lib/krb5/krcache.c
lhoward: char uniquename[sizeof(KRCC_NAME_PREFIX) + KRCC_NAME_RAND_CHARS];
...

My reading is that we copy PREFIX and then have RAND_CHARS + 1 bytes left?

lhoward: So we can copy RAND_CHARS in and terminate at s[prefixlen - 1 + RAND_CHARS]?

nicowilliams: Oy, I missed that.  Resolved (never mind).

lhoward:
Thanks Nico, will address them tomorrow.

lib/krb5/krcache.c
nicowilliams: @vdukhovni says we should use `euid` everywhere, not `ruid`, both because we don't want to read user ccaches with privs, and because `euid != ruid` is in fact how you express intent to use privilege (and privileged credentials) on Unix.

nicowilliams: So I think we leave this as-is, but maybe we should start a discussion of whether we should be using `euid` for the default `FILE` ccache.

lib/krb5/krcache.c
nicowilliams: Thanks @greghudson.  @lhoward do please remove these.

@greghudson I am suspicious of all the mutex usage here.  Does MIT actually allow multiple threads to be active in a shared `ccache`?  Is there reference counting to protect against racing to use and racing to free?

greghudson: Note that if multiple threads can't safely access a ccache simultaneously, they probably can't also access a GSS cred id simultaneously.

lib/krb5/krcache.c
```diff
@@ -0,0 +1,1972 @@
...
+update_keyring_expiration(krb5_context context, krb5_ccache id)
+{
+    krb5_krcache *data = KRCACHE(id);
+    krb5_cc_cursor cursor;
+    krb5_creds creds;
+    krb5_timestamp now, endtime = 0;
+    unsigned int timeout;
+
+    /*
+     * We have no way to know what is the actual timeout set on the keyring.
```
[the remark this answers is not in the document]
lhoward: How about updating the keyring expiration iff the credential being stored is a start TGT and it is later than the expiration recorded in the keyring?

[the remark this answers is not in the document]
lhoward: NB: there is no syscall to get the timeout on a keyring, so the only option is to enumerate the creds in one.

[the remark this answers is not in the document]
lhoward: So we would change the behavior of update_keyring_expiration() to ignore non-start-TGTs

nicowilliams:
You can stop at the first start TGT.  We always re-initialize to store a
fresh TGT.

nicowilliams:
And you know whether you're storing a start TGT or something else, so no
need to iterate at all.  IMO.  @greghudson might know better though.

nicowilliams:
In principle you can have multiple start TGTs.  I'm practice it doesn't
happen.  We could always make a ccconfig entry to store the expiration time
so we only have to look it up (and maybe update it).  Whatever.

jaltman:
Start TGT or ticket with the INITIAL flag set?

lhoward:
> Start TGT or ticket with the INITIAL flag set?

Good point. How about

```
    if (creds->times.endtime > now) {
        krb5_timestamp timeout;

        timeout = creds->times.endtime > now ? creds->times.endtime - now : 1;

        /* Set timeout on credential key */
        (void) keyctl_set_timeout(cred_key, timeout);

        /* Iff a start TGT, set timeout on credentials cache */
        if (creds->flags.b.initial &&
            krb5_principal_is_root_krbtgt(context, creds->server))
            (void) keyctl_set_timeout(data->cache_id, timeout);
    }
```

nicowilliams:
kinit -R stores a non-INITIAL TGT in a re-initialized ccache.  That's why i didn't suggest using the end time of INITIAL tickets.

Let's leave this optimization for some other time.

greghudson:
If you require the server to be a TGT principal, you won't set an expiration on caches obtained with kinit -S.

nicowilliams:
I mean, you can use the end time of INITIAL tickets and also the end time of start TGTs.

lhoward:
On another note, it seems that changetime and princ_id are the only mutable state in a krb5_krcache (which are integers). So we could use compare and swap rather than a mutex?

jaltman:
Although I've wanted the ability to renew non-start TGTs forever there is no support for it even though many KDCs issue non-start TGT service tickets with the renewable flag set and a renew lifetime specified.

lhoward:
Cleaned up, added tests, etc. I'll give it a final once-over next week and merge.

lhoward:
Is there a way to get keyutils installed on the Travis build machines?

greghudson:
In .travis.yml, in the "sudo apt-get install" line (under before_install), add libkeyutils-dev.  Also add keyutils if you need to run keyctl for automated tests.

lib/krb5/test_cc.c
```diff
@@ -691,6 +695,9 @@ main(int argc, char **argv)
 #endif
     test_init_vs_destroy(context, krb5_cc_type_scc);
     test_init_vs_destroy(context, krb5_cc_type_dcc);
+#ifdef HAVE_KEYUTILS_H
+    //test_init_vs_destroy(context, krb5_cc_type_keyring);
```
nicowilliams: Is this not passing yet?

lhoward: Did I write that? 🤔 I’ll fix now.

nicowilliams:
Also, you're not updating the Travis `.yml` either, so it's not getting tested in Travis.

lhoward:
Didn’t know how to do that until Greg pointed it out earlier, it’s on my list to fix.

lib/krb5/test_cc.c
lhoward: Actually this test does fail. Investigating.

lhoward: @nicowilliams, this could either be a bug in the test suite, or in the keyring cache. `test_init_vs_destroy()` creates a unique ccache, resolves it into a second, and destroys the first before initializing the second. This leaves the second with a dangling keyring reference to the first.

We could workaround it by catching the error and re-resolving the ccache in `krcc_initialize()`. Or, the tset suite could destroy the first ccache before resolving the second. Thoughts?

lhoward: Interestingly the error I get when trying to access the invalid keyring handle depends on whether I'm running under the debugger or not...

nicowilliams: This doesn't have to work....

lhoward: I fixed it, if the underlying keyring is destroyed between resolve- and initialize-time, we re-resolve (and thus recreate) it.

Issue #322, opened 2017-08-12
krb5_get_init_creds_opt_set_change_password_prompt incomplete


rra:
In 7422cd1f6 `krb5_get_init_creds_opt_set_change_password_prompt` was added. Unfortunately, it doesn't work (causing the pam-krb5 test suite to break) because it doesn't catch all of the places where the password change prompt is shown. It modified the fallback code at the end of `krb5_get_init_creds_password` to call `change_password`, but the normal password change is kicked off here:

https://github.com/heimdal/heimdal/blob/master/lib/krb5/init_creds_pw.c#L2381

The same check needs to be done here, but unfortunately it doesn't look like the init_creds options are passed down this far.  Or perhaps this password change logic should be deleted in favor of always using the code that handles the return value of `krb5_init_creds_get`?  I can submit a patch, but I'm not sure the correct fix.

(cc @spaceone)

lhoward:
Removing the password change code from `krb5_init_creds_step()` looks like the right solution to me. MIT handles password expiry at a higher level, in `krb5_get_init_creds_password()` (as Heimdal does also). It could be an issue if we ever implement IAKERB but MIT doesn't handle password changes within IAKERB either, as far as I can tell. Patch coming up.

lhoward:
@rra please let me know if this fixes your issue

lhoward:
Hmm, this may break some tests. Investigating.

lhoward:
Need to revist: kinit uses `krb5_init_creds_get()` directly so removing this code breaks it. Will fix tomorrow.

lhoward:
OK, try new patch.

lhoward:
Merged in 070d0cf

PR #460, opened 2018-12-22
krb5: remove key expired handler from krb5_init_creds_step() (#322)

lhoward:
krb5: krb5_get_init_creds_opt_set_change_password_prompt incomplete (#322)
    
krb5_get_init_creds_opt_set_change_password_prompt() was being ignored by krb5_init_creds_step() which broke pam_krb5 tests. MIT doesn't handle password expiration within krb5_init_creds_step(), instead deferring to higher level functions such as krb5_get_init_creds_password(). However, Heimdal kinit uses krb5_init_creds_step() directly and thus requires this behaviour to be implemented to pass its own tests.

### DIFF
--- a/lib/krb5/init_creds_pw.c
+++ b/lib/krb5/init_creds_pw.c
@@ -66,6 +66,7 @@ typedef struct krb5_get_init_creds_ctx {
 
     struct {
 	unsigned change_password:1;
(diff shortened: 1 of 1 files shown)


nicowilliams:
LGTM.  Is it ready?

lhoward:
Yup

Issue #360, opened 2018-02-15
kadmin: pw-expiration-time cannot bet set when adding new principal with random key


klada:
When creating new principals with a random key the `pw-expiration-time` argument is ignored. The value is always set to the default of 1 year, even if the principal expiration time is shorter.

Example:
```
kadmin -l add \
  --random-key \
  --max-ticket-life=30h \
  --max-renewable-life=30d \
  --expiration-time=2018-12-31 \
  --pw-expiration-time=2018-12-31 \
  --use-defaults \
  foobar
```

Result:

```
kadmin -l get foobar
            Principal: foobar@REALM.LAN
    Principal expires: 2018-12-31 23:59:59 UTC
     Password expires: 2019-02-14 15:03:08 UTC
 Last password change: 2018-02-14 15:03:08 UTC
```

lhoward:
@klada please try fix in pull request above

lhoward:
Merged in def6e4fc

lhoward:
Great. I can’t even remember this fix now, it’s clearly been a busy two weeks!

PR #461, opened 2018-12-23
kadmin: honour pw-expiration-time when adding principal (#360)

lhoward:
Adding a principal with a random key or password did not respect non-default password expiration times, because the act of setting the key or password would clobber it with the default. As we update the principal anyway after setting the keys, use this opportunity to restore the requested password expiration time. (There are other ways to solve this, but this is the least intrusive.)


nicowilliams:
LGTM.  Waiting for Travis.

Issue #397, opened 2018-08-08
Merge Kerberos account synchronization plugin patch


macrotex:
Merge into the Heimdal kadmin libraries code that supports the krb5-sync plugin that synchronizes Kerberos passwords and account enabling/disabling with an Active Directory.

From the patch documentation:

> The patches in this directory apply to Heimdal and add plugin support to the kadmin libraries for performing actions before and after a password change is committed to the KDC database and after a change is made to the attributes of a principal (specifically, a change to DISALLOW_ALL_TIX).

The source code and documentation is here: https://git.eyrie.org/git/kerberos/krb5-sync.git

Note that this issue is asking _only_ to integrate the plugin code into the Heimdal kadmin code. Patches for previous versions of Heimdal, as well as documentation, is in the `patches/` directory of the Git repository referenced above.

jaltman:
The upstream repository is forked as https://github.com/heimdal/krb5-sync.   The master branch was updated for Heimdal 7.4.

lhoward:
The patch doesn't implement the delete, randkey, or rename hooks. Should these be implemented before merging?

Also, it has some bugs (which I will fix); postcommit hooks shouldn't trample on the return code.

lhoward:
Is it OK to change the plugin ABI? I'm happy to leave as-is, but it seems like if the post-commit hooks are going to get called irrespective of update success, then they should also be passed the return value of kadm5_log_modify()

macrotex:
> 
> 
> The patch doesn't implement the delete, randkey, or rename hooks. Should these be implemented before merging?
> 
> Also, it has some bugs (which I will fix); postcommit hooks shouldn't trample on the return code.

I don't mind supporting hooks for the additional actions, but all I really need are the add and change password hooks. It's up to you.

jaltman:
@lhoward for completeness please implement delete, randkey and renamke hooks.

lhoward:
Fixed in master

PR #462, opened 2018-12-23
kadm5: pre/post-commit plugin hook for kadm5 update operations (#397)

lhoward:
This change adds plugin support to the kadmin libraries for performing actions before and after a password change is committed to the KDC database and after a change is made to the attributes of a principal (specifically, a change to DISALLOW_ALL_TIX).

This change adds a hook_libraries configuration option to the [kadmin] section of krb5.conf (or kdc.conf if you use that file) that must be set to load the module. That configuration option is in the form:

```
[kadmin]
  hook_libraries = /usr/local/lib/krb5/plugins/kadm5_hook/krb5_sync.so
```

where the value is the full path to the plugin that you want to load. If this option is not present, kadmind will not load a plugin and the changes from the patch will be inactive. If this option is given and the plugin cannot be loaded, kadmind startup will abort with a (hopefully useful) error message in syslog.

Any plugin used with this patch must expose a public struct named kadm5_hook. That struct must contain the following:

```
typedef struct kadm5_hook {
    const char *name;
    int version;
    const char *vendor;

    krb5_error_code (*init)(krb5_context, void **);
    void (*fini)(krb5_context, void *);

    krb5_error_code (*chpass)(krb5_context, void *, enum kadm5_hook_stage,
                              krb5_principal, const char *);
    krb5_error_code (*create)(krb5_context, void *, enum kadm5_hook_stage,
                              kadm5_principal_ent_t, uint32_t mask,
                              const char *password);
    krb5_error_code (*modify)(krb5_context, void *, enum kadm5_hook_stage,
                              kadm5_principal_ent_t, uint32_t mask);
    ...
} kadm5_hook;
```

where enum kadm5_hook_stage is:

```
enum kadm5_hook_stage {
    KADM5_HOOK_STAGE_PRECOMMIT,
    KADM5_HOOK_STAGE_POSTCOMMIT
};
```

init creates a hook context that is passed into all subsequent calls. chpass is called for password changes, create is called for principal creation (with the newly-created principal in the kadm5_principal_ent_t argument), and modify is called when a principal is modified.

These functions should follow the normal Kerberos calling convention of returning 0 on success and a Kerberos error code on failure, setting the Kerberos error message in the provided context.

This change is submitted under the following license

Copyright 2012, 2013
The Board of Trustees of the Leland Stanford Junior University

Copying and distribution of this file, with or without modification, are permitted in any medium without royalty provided the copyright notice and this notice are preserved. This file is offered as-is, without any warranty.


jaltman:
`/*
 * libkadm5srv expects a symbol named kadm5_hook_v0 exported by the dynamicaly
 * loaded module and of type kadm5_hook.  version must be
 * KADM5_HOOK_VERSION_V0.  Any or all of the function pointers may be NULL, in
 * which case that hook will not be called.
 */`

I don't like the idea of exporting a data symbol as it doesn't work well on Windows. I would prefer the shared library export a function that passes the data structure as an output parameter. 

Since the plugin functions accept the krb5_context it is critical that we detect mismatches due to modifications of the krb5_context structure.   Perhaps export a function

`  int init(void **kadm_hook);`

and add a `size_t sizeof_krb5_context` parameter to the structure in addition to the kadm_hook version.

lhoward:
I've amended the patch to export a function instead. I'd prefer to use a version number than sizeof(krb5_context) as the hook won't have access to krb5_locl.h. If KADM5_STRUCT_VERSION changes then we can up KADM5_HOOK_VERSION.

lhoward:
I also have a patch for krb5-sync which compiles although I don't have the infrastructure to test it.

lhoward:
TOOD: use krb5_const_principal in plugin API

nicowilliams:
@lhoward given your TODO comment, is this ready for review?

lhoward:
I’ll fix it this morning.

lhoward:
Fixed, ready for review

lib/kadm5/kadm5-hook.h
```diff
@@ -0,0 +1,126 @@
...
+/*
+ * libkadm5srv expects a symbol named kadm5_hook_init that must be a function
+ * of type kadm5_hook_init_t. The function will be called with the maximum
+ * version of the hook API supported by libkadm5; the plugin may return an
+ * earlier version.
+ */
+typedef struct kadm5_hook {
+    const char *name;
+    uint32_t version;
+    const char *vendor;
```
nicowilliams: Unused.  What's it for?  Documentation?

lib/kadm5/chpass_s.c
```diff
@@ -67,6 +68,22 @@ change(void *server_handle,
     if (ret)
 	goto out2;
 
+    for (i = 0; i < context->num_hooks; i++) {
```
nicowilliams: The calling of the hooks could be abstracted out.

lib/kadm5/server_hooks.c
```diff
@@ -0,0 +1,175 @@
...
+    krb5_context context = ctx->context;
+    char **libraries;
+    size_t i;
+    void *handle = NULL;
+    struct kadm5_hook_context *hook_context = NULL;
+    struct kadm5_hook_context **tmp;
+    kadm5_ret_t ret = KADM5_BAD_SERVER_HOOK;
+
+    libraries = krb5_config_get_strings(context, NULL,
+					"kadmin", "hook_libraries", NULL);
```
nicowilliams: s/hook_libraries/hooks/

lib/kadm5/kadm5-hook.h
lhoward: Appears to be.

lib/kadm5/kadm5-hook.h
```diff
@@ -0,0 +1,126 @@
...
+    void (KRB5_CALLCONV *fini)(krb5_context, void *data);
```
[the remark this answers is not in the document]
lhoward: I agree that passing a table of function pointers *in* is better, but as you mentioned, given we're passing `kadm5_principal_ent_t` in perhaps it's fairly tightly coupled already.

lib/kadm5/kadm5-hook.h
```diff
@@ -0,0 +1,126 @@
...
+    /*
+     * Hook functions; NULL functions are ignored. code is only valid on
+     * post-commit hooks and represents the result of the commit. Post-
+     * commit hooks are not called if a pre-commit hook aborted the call.
+     */
+    krb5_error_code (KRB5_CALLCONV *chpass)(krb5_context context,
+					    void *data,
+					    enum kadm5_hook_stage stage,
+					    krb5_error_code code,
+					    krb5_const_principal princ,
```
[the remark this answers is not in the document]
lhoward: I feel if we're going down that path, then the hook API should send JSON or something...

lib/kadm5/kadm5-hook.h
```diff
@@ -0,0 +1,126 @@
...
+					    const char *password);
+
+    krb5_error_code (KRB5_CALLCONV *create)(krb5_context context,
+					    void *data,
+					    enum kadm5_hook_stage stage,
+					    krb5_error_code code,
+					    kadm5_principal_ent_t ent,
+					    uint32_t mask,
+					    const char *password);
```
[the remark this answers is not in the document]
lhoward: Well, also for creating with keys. But perhaps that should be a separate API, I was just going on the Stanford patch.

lib/kadm5/kadm5-hook.h
```diff
@@ -0,0 +1,126 @@
...
+					    enum kadm5_hook_stage stage,
+					    krb5_error_code code,
+					    krb5_const_principal princ,
+					    const char *password);
+
+    krb5_error_code (KRB5_CALLCONV *create)(krb5_context context,
+					    void *data,
+					    enum kadm5_hook_stage stage,
+					    krb5_error_code code,
+					    kadm5_principal_ent_t ent,
```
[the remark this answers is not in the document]
lhoward: Right.

lib/kadm5/server_hooks.c
```diff
@@ -0,0 +1,175 @@
...
+	handle = dlopen(library, RTLD_NOW | RTLD_LOCAL | RTLD_GROUP);
+	if (handle == NULL) {
+	    krb5_warnx(context, "failed to open `%s': %s", library, dlerror());
+	    ret = KADM5_SERVER_HOOK_NOT_FOUND;
+	    goto fail;
+	}
+
+	hook_init = dlsym(handle, "kadm5_hook_init");
+	if (hook_init == NULL) {
+	    krb5_warnx(context, "didn't find `kadm5_hook_init' symbol in `%s':"
```
nicowilliams: I forget.  Do these go to stderr if we're running `kadmin -l`?  I fear for usability if not.

lib/kadm5/chpass_s.c
lhoward: Ack.

lib/kadm5/kadm5-hook.h
```diff
@@ -0,0 +1,126 @@
...
+					      krb5_boolean keepold,
+					      int n_ks_tuple,
+					      krb5_key_salt_tuple *ks_tuple,
+					      int n_keys,
+					      krb5_keyblock *keyblocks);
+
+} kadm5_hook;
+
+typedef krb5_error_code
+(KRB5_CALLCONV *kadm5_hook_init_t)(krb5_context context,
```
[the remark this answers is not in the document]
lhoward: So, we could get the *plugin* to pass us back krb5_init_context(), that we plugins can't get around it...

lib/kadm5/server_hooks.c
lhoward: Think it depends on `[logging]`

lib/kadm5/server_hooks.c
```diff
@@ -0,0 +1,175 @@
...
+    char **libraries;
+    size_t i;
+    void *handle = NULL;
+    struct kadm5_hook_context *hook_context = NULL;
+    struct kadm5_hook_context **tmp;
+    kadm5_ret_t ret = KADM5_BAD_SERVER_HOOK;
+
+    libraries = krb5_config_get_strings(context, NULL,
+					"kadmin", "hook_libraries", NULL);
+    if (libraries == NULL)
```
nicowilliams: Oy, we can't tell if `krb5_config_get_string()` failed due to `ENOMEM`.  In principle we could check if `errno == ENOMEM` here.  I believe we don't check anywhere for that, so don't.

Oh, and you need to `krb5_config_free_strings(libraries)`, FYI.

lib/kadm5/server_hooks.c
```diff
@@ -0,0 +1,175 @@
...
+
+    return 0;
+
+fail:
+    _kadm5_s_free_hooks(ctx);
+    if (hook_context != NULL)
+	free(hook_context);
+    if (handle != NULL)
+	dlclose(handle);
+
```
nicowilliams:
```
    if (libraries != NULL)
        krb5_config_free_strings(libraries);
````

File: lib/kadm5/kadm5-hook.h (hunk 5 above)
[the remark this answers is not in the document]
Comment (nicowilliams) [reply lib/kadm5/kadm5-hook.h]: Someday, when I finish my PG backend, we could use JSON-serializations of HDB entries.

File: lib/kadm5/kadm5-hook.h (hunk 9 above)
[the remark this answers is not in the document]
Comment (nicowilliams) [reply lib/kadm5/kadm5-hook.h]: That works for me.

File: lib/kadm5/server_hooks.c (hunk 8 above)
Comment (nicowilliams) [reply lib/kadm5/server_hooks.c]: OK.  We should really make `kadmin -l` force logging to also go to `stderr`, but not today.

File: lib/kadm5/kadm5-hook.h (hunk 9 above)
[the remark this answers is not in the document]
Comment (lhoward) [reply lib/kadm5/kadm5-hook.h]: Is this incompatible with loading with RTLD_GROUP? Maybe we shouldn't be specifying that.

**Comment by nicowilliams** [comment]
All my comments are in.

Changes requested:

 - abstract calling the hooks somewhat
 - don't leak `libraries`
 - s/`"hook_libraries"`/`"hooks"`/
 - have the plugin output a pointer to `krb5_init_context()` so `libkadm5` can check that the plugin is using the same `libkrb5`

**Comment by lhoward** [comment]
Fixed, also added a not-really-related commit that moves password quality checks out of `kadmindd` and `kpasswdd` and into `libkadm5`, with a view to eventually refactoring password quality checks as pre-commit hooks.

**Comment by nicowilliams** [comment]
Well, the least we could do is retry on realm create if the random password is "weak".

File: lib/kadm5/server_hooks.c [hunk 12]
```diff
@@ -0,0 +1,183 @@
...
+	    goto fail;
+	}
+
+	if (hook->version < KADM5_HOOK_VERSION_V1)
+	    ret = KADM5_OLD_SERVER_HOOK_VERSION;
+	else if (hook->version > KADM5_HOOK_VERSION_V1)
+	    ret = KADM5_NEW_SERVER_HOOK_VERSION;
+	if (ret) {
+	    krb5_warnx(context, "version of loaded hook `%s' is %u"
+		       " (supported versions %u to %u)", hookpath, hook->version,
```
jaltman: Should include hook vendor in the output;

lib/kadm5/server_hooks.c
```diff
@@ -0,0 +1,183 @@
...
+		       KADM5_HOOK_VERSION_V1, KADM5_HOOK_VERSION_V1);
+	    hook->fini(context, data);
+	    goto fail;
+	}
+
+	if (hook->init_context != krb5_init_context) {
+	    krb5_warnx(context, "loaded hook `%s' is not linked against "
```
jaltman: should include hook version and vendor in the output.

lib/kadm5/server_hooks.c
```diff
@@ -0,0 +1,183 @@
...
+	    ret = krb5_enomem(context);
+	    hook->fini(context, data);
+	    goto fail;
+	}
+	ctx->hooks = tmp;
+	ctx->hooks[ctx->num_hooks] = hook_context;
+	hook_context = NULL;
+	ctx->num_hooks++;
+    }
+
```
jaltman: I think we should krb5_warnx() the hook name, version and vendor that was successfully loaded.

lib/kadm5/server_hooks.c
```diff
@@ -0,0 +1,183 @@
...
+    ctx->hooks = NULL;
+    ctx->num_hooks = 0;
+}
+
+# else /* !HAVE_DLOPEN */
+
+kadm5_ret_t
+_kadm5_s_init_hooks(kadm5_server_context *ctx)
+{
+    return 0;
```
jaltman: In the no DLOPEN case we should log a hook load failure and return an error.

nicowilliams: Only if hooks are defined in krb5.conf though.

nicowilliams:
BTW, @lhoward, thanks for noticing and using `krb5_prepend_error_message()`.

lhoward:
The lack of hyphen did trouble me @jaltman :)

I'll fix the other issues today.

lhoward:
Additional review comments were addressed in f62b00e3

lhoward:
> Well, the least we could do is retry on realm create if the random password is "weak".

Something like this?

```udiff
diff --git a/kadmin/init.c b/kadmin/init.c
index 20ed93216..abf57d6cc 100644
--- a/kadmin/init.c
+++ b/kadmin/init.c
@@ -50,11 +50,12 @@ create_random_entry(krb5_principal princ,
     kadm5_ret_t ret;
     int mask = 0;
     krb5_keyblock *keys;
-    int n_keys, i;
+    int n_keys, i, n_password_tries = 1;
     char *name;
     const char *password;
     char pwbuf[512];
 
+retry_password:
     random_password(pwbuf, sizeof(pwbuf));
     password = pwbuf;
 
@@ -83,6 +84,10 @@ create_random_entry(krb5_principal princ,
     if(ret) {
        if (ret == KADM5_DUP && (flags & CRE_DUP_OK))
            goto out;
+       if (ret == KADM5_PASS_Q_DICT && n_password_tries < 5) {
+           n_password_tries++;
+           goto retry_password;
+       }
        krb5_warn(context, ret, "create_random_entry(%s): randkey failed",
                  name);
        goto out;
```

nicowilliams:
Yeah, something like that.  Or disable password quality checks at init time, or special-case the `K/M` principal.

lhoward:
Realm initialization rewrites the random passwords with real random keys, so perhaps the real solution is to create a principal with no password and no keys, before calling generating a random key.

lhoward:
OK, we can do this, cleaner, tests pass:

```udiff
diff --git a/kadmin/init.c b/kadmin/init.c
index 20ed93216..47201424e 100644
--- a/kadmin/init.c
+++ b/kadmin/init.c
@@ -52,11 +52,6 @@ create_random_entry(krb5_principal princ,
     krb5_keyblock *keys;
     int n_keys, i;
     char *name;
-    const char *password;
-    char pwbuf[512];
-
-    random_password(pwbuf, sizeof(pwbuf));
-    password = pwbuf;
 
     ret = krb5_unparse_name(context, princ, &name);
     if (ret) {
@@ -76,14 +71,14 @@ create_random_entry(krb5_principal princ,
        mask |= KADM5_MAX_RLIFE;
     }
     ent.attributes |= attributes | KRB5_KDB_DISALLOW_ALL_TIX;
-    mask |= KADM5_ATTRIBUTES;
+    mask |= KADM5_ATTRIBUTES | KADM5_KEY_DATA;
 
-    /* Create the entry with a random password */
-    ret = kadm5_create_principal(kadm_handle, &ent, mask, password);
+    /* Create the entry with no keys or password */
+    ret = kadm5_s_create_principal_with_key(kadm_handle, &ent, mask);
     if(ret) {
        if (ret == KADM5_DUP && (flags & CRE_DUP_OK))
            goto out;
-       krb5_warn(context, ret, "create_random_entry(%s): randkey failed",
+       krb5_warn(context, ret, "create_random_entry(%s): create failed",
                  name);
        goto out;
     }
```

PR #466, opened 2018-12-24
Samba Heimdal upgrade

iboukris:
Recently I got involved in @abartlet's efforts to upgrade Heimdal in Samba, and have come up with some patches to help with that.

Note about the last commit, enterprise vs canonicalize; this is now needed in order to be able to test and exercise flows where enterprise name type is used without canonicalize kdc-req-option, and confirm the name is not being canonicalized by the KDC. It used to be done in Samba torture tests by hooking via send_to_kdc and removing the canonicalize flag but this now fails due to PA-REQ-ENC-PA-REP protection.

I noticed a related discussion in issue #452, if we prefer to keep the alignment of the two flags for convenience then perhaps we can move both to kinit code so that other consumer of the API could chose whether or not the canonicalize flag is set (also, this will avoid changing the tests).

### DIFF
--- a/kdc/kerberos5.c
+++ b/kdc/kerberos5.c
@@ -1221,6 +1221,30 @@ get_pa_etype_info2(krb5_context context,
     return 0;
 }
 
+static krb5_error_code
+get_pa_etype_info_both(krb5_context context,
+		       krb5_kdc_configuration *config,
+		       METHOD_DATA *md, Key *ckey)
+{
+    krb5_error_code ret;
+
+    /*
+     * RFC4120 requires:
+     * - If the client only knows about old enctypes, then send
+     *   both info replies (we send 'info' first in the list).
+     * - If the client is 'modern', because it knows about 'new'
+     *   enctype types, then only send the 'info2' reply.
+     */
+
+    if (older_enctype(ckey->key.keytype)) {
+	ret = get_pa_etype_info(context, config, md, ckey);
+	if (ret)
+	    return ret;
+    }
+
+    return get_pa_etype_info2(context, config, md, ckey);
+}
+
 /*
  *
  */
@@ -1816,6 +1840,21 @@ _kdc_as_rep(kdc_request_t r,
 	    if (pa) {
 		ret = pat[n].validate(r, pa);
 		if (ret != 0) {
+		    krb5_error_code  ret2;
+		    Key *ckey = NULL;
+		    /*
+		     * If there is a client key, send ETYPE_INFO{,2}
+		     */
+		    ret2 = _kdc_find_etype(context,
+					   config->preauth_use_strongest_session_key,
+					   TRUE, r->client, b->etype.val,
+					   b->etype.len, NULL, &ckey);
+		    if (ret2 == 0) {
+			ret2 = get_pa_etype_info_both(context, config,
+						      &error_method, ckey);
+			if (ret2 != 0)
+			    ret = ret2;
+		    }
 		    goto out;
 		}
 		kdc_log(context, config, 0,
@@ -1847,27 +1886,7 @@ _kdc_as_rep(kdc_request_t r,
 			      config->preauth_use_strongest_session_key, TRUE,
 			      r->client, b->etype.val, b->etype.len, NULL, &ckey);
 	if (ret == 0) {
-
-	    /*
-	     * RFC4120 requires:
-	     * - If the client only knows about old enctypes, then send
-	     *   both info replies (we send 'info' first in the list).
-	     * - If the client is 'modern', because it knows about 'new'
-	     *   enctype types, then only send the 'info2' reply.
-	     *
-	     * Before we send the full list of etype-info data, we pick
-	     * the client key we would have used anyway below, just pick
-	     * that instead.
-	     */
-
-	    if (older_enctype(ckey->key.keytype)) {
-		ret = get_pa_etype_info(context, config,
-					&error_method, ckey);
-		if (ret)
-		    goto out;
-	    }
-	    ret = get_pa_etype_info2(context, config,
-				     &error_method, ckey);
+	    ret = get_pa_etype_info_both(context, config, &error_method, ckey);
 	    if (ret)
 		goto out;
 	}
--- a/kdc/krb5tgs.c
+++ b/kdc/krb5tgs.c
@@ -807,7 +807,7 @@ tgs_make_reply(krb5_context context,
     if(ret)
 	goto out;
 
-    ret = copy_Realm(&server_principal->realm, &rep.ticket.realm);
+    ret = copy_Realm(&server->entry.principal->realm, &rep.ticket.realm);
     if (ret)
 	goto out;
     _krb5_principal2principalname(&rep.ticket.sname, server_principal);
--- a/lib/krb5/get_cred.c
+++ b/lib/krb5/get_cred.c
@@ -940,6 +940,12 @@ get_cred_kdc_capath(krb5_context context,
     return ret;
 }
 
+static krb5_boolean skip_referrals(krb5_principal server,
+				   krb5_kdc_flags *flags)
+{
+    return server->name.name_string.len < 2 && !flags->b.canonicalize;
+}
+
 /*
  * Get a service ticket from a KDC by chasing referrals from a start realm.
  *
@@ -966,7 +972,7 @@ get_cred_kdc_referral(krb5_context context,
     int want_tgt;
     size_t i;
 
-    if (in_creds->server->name.name_string.len < 2 && !flags.b.canonicalize) {
+    if (skip_referrals(in_creds->server, &flags)) {
 	krb5_set_error_message(context, KRB5KDC_ERR_PATH_NOT_ACCEPTED,
 			       N_("Name too short to do referals, skipping", ""));
 	return KRB5KDC_ERR_PATH_NOT_ACCEPTED;
@@ -1213,7 +1219,7 @@ _krb5_get_cred_kdc_any(krb5_context context,
                                   second_ticket,
                                   out_creds,
                                   ret_tgts);
-        if (ret == 0)
+        if (ret == 0 || skip_referrals(in_creds->server, &flags))
             return ret;
     }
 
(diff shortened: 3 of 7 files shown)


nicowilliams:
LGTM.  I'd like @lhoward to review [cc1a957](https://github.com/heimdal/heimdal/pull/466/commits/cc1a957219fe23f55357253a894403da82b0fb2a), but I do believe it's correct.

The subject of [b1e46b4](https://github.com/heimdal/heimdal/pull/466/commits/b1e46b40532b708de39b2e3e4f423298e4bb047a) should be something more like

```
KDC: Allow hdb to set the issued ticket's realm
```

That's all I got.  I'll wait for @lhoward's review.

kdc/kerberos5.c
```diff
@@ -1816,6 +1840,21 @@ _kdc_as_rep(kdc_request_t r,
...
+		    krb5_error_code  ret2;
+		    Key *ckey = NULL;
+		    /*
+		     * If there is a client key, send ETYPE_INFO{,2}
+		     */
+		    ret2 = _kdc_find_etype(context,
+					   config->preauth_use_strongest_session_key,
+					   TRUE, r->client, b->etype.val,
+					   b->etype.len, NULL, &ckey);
+		    if (ret2 == 0) {
```
iboukris: I think I shouldn't ignore error from _kdc_find_etype(), I'll change to:
```suggestion
		    if (ret2) {
		        ret = ret2;
		        goto out;
		    }
		    ret2 = get_pa_etype_info_both(0) ...
```

iboukris:
My last comment was wrong, please ignore.
I've updated the subject of commit b1e46b4.

kdc/krb5tgs.c
```diff
@@ -807,7 +807,7 @@ tgs_make_reply(krb5_context context,
     if(ret)
 	goto out;
 
-    ret = copy_Realm(&server_principal->realm, &rep.ticket.realm);
+    ret = copy_Realm(&server->entry.principal->realm, &rep.ticket.realm);
```
lhoward: Hmm, is this just needed to handle referrals? If I remember right, as of W2K3 the canonicalize flag is ignored for TGS service ticket requests, unless you're returning a referral. How about this?

```
diff --git a/kdc/krb5tgs.c b/kdc/krb5tgs.c
index 51f0c9208..1fcdd5e6f 100644
--- a/kdc/krb5tgs.c
+++ b/kdc/krb5tgs.c
@@ -755,7 +755,7 @@ tgs_make_reply(krb5_context context,
     EncTicketPart et;
     KDCOptions f = b->kdc_options;
     krb5_error_code ret;
-    int is_weak = 0;
+    int is_weak = 0, tmp;
 
     memset(&rep, 0, sizeof(rep));
     memset(&et, 0, sizeof(et));
@@ -807,7 +807,11 @@ tgs_make_reply(krb5_context context,
     if(ret)
        goto out;
 
-    ret = copy_Realm(&server_principal->realm, &rep.ticket.realm);
+    if (krb5_find_padata(enc_pa_data->val, enc_pa_data->len,
+                        KRB5_PADATA_SERVER_REFERRAL, &tmp))
+       ret = copy_Realm(&server->entry.principal->realm, &rep.ticket.realm);
+    else
+       ret = copy_Realm(&server_principal->realm, &rep.ticket.realm);
     if (ret)
        goto out;
     _krb5_principal2principalname(&rep.ticket.sname, server_principal);
@@ -1108,7 +1112,8 @@ need_referral(krb5_context context, krb5_kdc_configuration *config,
 {
     const char *name;
 
-    if(!options->canonicalize && server->name.name_type != KRB5_NT_SRV_INST)
+    if(!options->canonicalize && !options->enc_tkt_in_skey &&
+       server->name.name_type != KRB5_NT_SRV_INST)
        return FALSE;
 
     if (server->name.name_string.len == 1)
```

Checking `enc_tkt_in_skey` is unset mirrors the MIT code. If this breaks any of your tests, let me know more detail.

lib/krb5/get_cred.c
```diff
@@ -1193,6 +1193,8 @@ _krb5_get_cred_kdc_any(krb5_context context,
 {
     krb5_error_code ret;
     krb5_deltat offset;
+    krb5_boolean no_ref = in_creds->server->name.name_string.len < 2 &&
+			  !flags.b.canonicalize;
```
lhoward: `need_referral()` in the KDC allows referrals to be sent for one-part names. Can you describe the desired behavior in more detail? Will Windows issue a ticket when the SPN is specified as `sAMAccountName$@REALM`?

kdc/krb5tgs.c
nicowilliams: So, yeah, in principle the `Ticket`'s `sname` and `realm` can be anything at all because the `Ticket` is opaque.  And it's wrong for _clients_ to check if the `KDC-REP`'s `Ticket`'s `sname` and `realm` matches the one requested.  So I'm OK with this change.

@lhoward PA-SVR-REFERRAL-DATA is old.  Clients aren't supposed to use it.

Or, actually, what's this change _really_ about?  I'm kinda tired, so I'm probably missing something.

lib/krb5/get_cred.c
nicowilliams: This looks like an optimization: if the capaths path didn't work, why would the referrals path work if the KDC wouldn't issue a referral?  And right now Heimdal (and MIT) only do referrals in TGS when the service principal is something like a host-based service name.  But I don't think we should preclude referrals for other name types, so I'm inclined to reject this part of this change.

lib/krb5/init_creds_pw.c
```diff
@@ -426,9 +426,7 @@ get_init_creds_common(krb5_context context,
     if (ctx->keyproc == NULL)
 	ctx->keyproc = default_s2k_func;
 
-    /* Enterprise name implicitly turns on canonicalize */
-    if ((ctx->ic_flags & KRB5_INIT_CREDS_CANONICALIZE) ||
-	krb5_principal_get_type(context, client) == KRB5_NT_ENTERPRISE_PRINCIPAL)
```
nicowilliams: Why do we want to lose this?

nicowilliams: Oh, right, yeah, that's explained in the PR's description.  Sorry for the noise.

kdc/krb5tgs.c
lhoward: PA-SVR-REFERRAL-DATA check was just to see whether referrals were being returned, without adding yet another argument to tgs_make_reply(). could add a flag instead.

nicowilliams: Ah, ok.  But do you agree that it almost doesn't matter?  Or do we need to worry about clients that do inspect the issued `Ticket`?

lhoward: @iboukris / @abartlet does Windows always canonicalize the realm on a TGS-REQ even when the canonicalize flag is unset?

lib/krb5/get_cred.c
lhoward: Agreed

kdc/krb5tgs.c
iboukris: As far as I can tell, realm canonicalization such of netbios and lowercase-realm is independent of the canonicalize flag, and is *always* done by Windows.

The current case is when a ticket for netbios name is requested, which can be done with:
$ ./kuser/kgetcred HTTP/apache.acme.com@ACME

The capath function would think it's a foreign realm and get a cross realm ticket first, but at the end the KDC returned ticket realm is the canonical one ACME.COM.

lhoward: I recall that Windows always canonicalizes the realm in an AS-REQ, but I wasn't sure about TGS-REQ in the non-referral case. Can you try ./kuser/kgetcred HTTP/apache.acme.com@acme.com and see what you get back?

lhoward: And also what is actually in the server ticket if possible.

lhoward: Assuming you have authenticated as a principal in ACME.COM that is.

lib/krb5/get_cred.c
iboukris: Yes, that's right, wouldn't call it an optimization but to preserve the error (such client not found) if we know it will certainly fail.

The background as mentioned, is that the change of order between referral and capath functions broke samba torture suite miserably. I was working on adapting the tests (while keeping the actual check of the KDC behavior) and noticed that sometimes KDC error from capath get masked by the error in the referral fallback (matching cred not found) and thought we could improve that.

kdc/krb5tgs.c
lhoward: OK, I guess this is fine then. MIT and Heimdal have different philosophies on whether name canonicalization should be done by the front- or back-end and I suppose there is not a right answer. In the AS-REQ path MIT will return the requested client name if canonicalize is not requested, but is has a special case for canonicalizing the realm for dealing with Windows; in the TGS-REQ path, it only canonicalizes when issuing referrals. Heimdal always returns the canonical principal, that is, the principal the backend returned, and it's up to the backend to *not* do that if (for example) HDB_F_CANON is not requested.

@nicowilliams do you have any further thoughts? I guess we at least should be consistent between AS and TGS paths, which this patch does. Myself, I don't really mind putting Windows-specific stuff in the KDC front-end as there is already so much anyway, but perhaps there's an argument that it's nice to avoid.

lib/krb5/get_cred.c
iboukris: Ah, sorry @nicowilliams I think I misread your comment previously. I agree that we might want to allow referrals for other name types but right now the referral function would error immediately if the name is short and canonicalize was not requested which hides an actual error from the KDC.
Perhaps I should move this check to a function which would be called from both places.

nicowilliams: Ohhh, I get it now.  The problem is that `get_cred_kdc_referral()` does this very same check and returns an error that is useless because it clobbers the more useful error from `get_cred_kdc_capath()`.

jaltman:
ab7685a  and 4bb9e44 look good to me.

kdc/krb5tgs.c
jaltman: I am fine with this change as-is.

nicowilliams: I have no further thoughts as to this patch.  I do think we should implement AD's case-insensitivity as to realms.  Case-insensitivity for other things (such as hostname and domainname components of host-based and domain-based service names) might be nice, but much harder to implement.

PR #343, opened 2017-10-11
Do not call getpwuid() if possible; use getpwuid_r() instead

nicowilliams:

PR #468, opened 2018-12-26
Do not call getpwuid() if possible; use getpwuid_r() instead

lhoward:
Revive @nicowilliams' patch in #343, rebased against master.


lhoward:
fix appl/su breakage

nicowilliams:
Thank you for rebasing that!

PR #476, opened 2018-12-26
Fix gss_add_cred() crash when output_cred_handle is NULL (fix #413)

### LINKED ISSUE #413: gss_add_cred() crashes on null output_cred_handle
RFC 2744 says about gss_add_cred():
```
   This routine can be used to either compose a new credential
   containing all credential-elements of the original in addition to the
   newly-acquire credential-element, or to add the new credential-
   element to an existing credential. If NULL is specified for the
   output_cred_handle parameter argument, the new credential-element
   will be added to the credential identified by input_cred_handle; if a
   valid pointer is specified for the output_cred_handle parameter, a
   new credential handle will be created.
```
Heimdal's mechglue gss_add_cred() does not expect output_cred_handle to be NULL; it dereferences it unconditionally at the beginning and end of the function, and only contains logic to create a new credential (with or without a copy of mechanism creds from an old credential), not to mutate an existing one.

(I discovered this through code inspection while looking into bugs in MIT krb5's version of the same function.  I have not tried to reproduce it, and I don't believe gss_add_cred() is used much in the wild.)

nicowilliams:
First pass.  I'm thinking of removing the `target_mc` stuff -- if you try to add an element for a mech you already have an element for then we should return `GSS_S_COMPLETE` and be done.

### DIFF
--- a/lib/gssapi/mech/gss_add_cred.c
+++ b/lib/gssapi/mech/gss_add_cred.c
@@ -1,5 +1,7 @@
 /*-
  * Copyright (c) 2005 Doug Rabson
+ * Copyright (c) 2018 Kungliga Tekniska Högskolan
+ * (Royal Institute of Technology, Stockholm, Sweden).
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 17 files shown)


nicowilliams:
Heimdal's interpretation of GSS is a bit funky:

- We allow `gss_name_t` instances to have multiple mech names (MNs).  I suspect this is about SPNEGO, specifically about inquiring a SPNEGO credential's name, but it really shouldn't be necessary.

- We seem to allow `gss_cred_id_t` instances to have multiple `{mech, name}` elements.  This can be good for acceptors, but it's pretty pointless for initiators.  Acceptors should just always use the default credential and then inquire the context for the name used by the initiator and possibly reject (at the application layer) if that name was not acceptable.

lhoward:
did the mcache fix accidentally slip in here?

lib/gssapi/mech/gss_add_cred.c
```diff
@@ -93,94 +93,132 @@ gss_add_cred(OM_uint32 *minor_status,
...
+            }
+            return GSS_S_COMPLETE;
+        }
+
+
+        if (output_cred_handle) {
+            new_cred = malloc(sizeof(struct _gss_cred));
+            if (!new_cred) {
+                *minor_status = ENOMEM;
+                return (GSS_S_FAILURE);
```
lhoward: return GSS_S_FAILURE

lhoward: picky I know

nicowilliams: Oh, I agree.

nicowilliams:
Which mcache fix?  The one for the deadlock?  I pushed it.

lib/gssapi/mech/gss_add_cred.c
nicowilliams: Pushed a separate commit to restyle this file.

nicowilliams:
The Travis OS X build evinced a `check-ldap` failure.  I don't know why that would be.  I've restarted that build in case it's spurious.

nicowilliams:
The `check-ldap` failure is due to a race condition in the test that I'll fix separately.

nicowilliams:
Needs more work.

lib/gssapi/mech/gss_add_cred.c
```diff
@@ -83,104 +84,120 @@ gss_add_cred(OM_uint32 *minor_status,
...
+        return GSS_S_BAD_MECH;
+    if (input_cred_handle == GSS_C_NO_CREDENTIAL &&
+        output_cred_handle == NULL) {
+        return GSS_S_CALL_INACCESSIBLE_WRITE;
+    }
+
+    /* Setup mut_cred to be the credential we mutate */
+    if (input_cred_handle != GSS_C_NO_CREDENTIAL &&
+        output_cred_handle != NULL) {
+        gss_cred_id_t new_cred = GSS_C_NO_CREDENTIAL;
```
nicowilliams: Initialization not needed here.

lib/gssapi/mech/gss_add_cred.c
```diff
@@ -83,104 +84,120 @@ gss_add_cred(OM_uint32 *minor_status,
...
+                                        (gss_const_cred_id_t)mut_cred, NULL,
+                                        NULL, NULL, actual_mechs);
+        if (major_status != GSS_S_COMPLETE)
+            _gss_mg_error(m, major_status, *minor_status);
+    }
+    if (major_status == GSS_S_COMPLETE)
+        *output_cred_handle = (gss_cred_id_t)mut_cred;
+    else
+        gss_release_cred(&junk, &release_cred);
+    free(mc);
```
nicowilliams: s/mc/new_mc/

lib/gssapi/netlogon/duplicate_cred.c
```diff
@@ -0,0 +1,54 @@
+/*
+ * Copyright (c) 2010 Kungliga Tekniska Högskolan
```
nicowilliams: Wrong copyright.

lib/gssapi/ntlm/duplicate_cred.c
```diff
@@ -0,0 +1,66 @@
...
+    new_cred->key.data = malloc(cred->key.length);
+    if (new_cred->username == NULL || new_cred->domain == NULL ||
+        new_cred->key.data == NULL) {
+        *output_cred_handle = (gss_cred_id_t) new_cred;
+        _gss_ntlm_release_cred(&junk, output_cred_handle);
+        *minor_status = ENOMEM;
+        return GSS_S_FAILURE;
+    }
+
+    new_cred->key.length = cred->key.length;
```
nicowilliams: Also copy the key contents!

lib/gssapi/ntlm/duplicate_cred.c
```diff
@@ -0,0 +1,66 @@
...
+    OM_uint32 junk;
+
+    *output_cred_handle = GSS_C_NO_CREDENTIAL;
+    if ((new_cred = calloc(1, sizeof(*new_cred))) == NULL) {
+        *minor_status = ENOMEM;
+        return GSS_S_FAILURE;
+    }
+    
+    new_cred->usage = cred->usage;
+    new_cred->username = strdup(cred->username);
```
nicowilliams: What if the input cred is `GSS_C_NO_CREDENTIAL`?

PR #478, opened 2018-12-27
kadmin: do not assign passwords at realm initialization

lhoward:
WIP – tests are failing on my end

Since c6bf100b password quality checks have been moved out of kadmindd and into libkadm5. This means that all password changes are subject to quality checks, if enforce_on_admin_set is true (the default). In rare instances it could be possible for realm initialization to fail because the randomly generated passwords do not pass the password quality test. Fix this by creating principals with no password or key, rather than with a random password.

Random *keys* continue to be set immediately after the principal is created, and before DISALLOW_ALL_TIX is unset, so there should be no functionality or security implications from this change. It is safe to call a server-side API such as kadm5_s_create_principal_with_key() as local_flag is asserted to be true.

### DIFF
--- a/kadmin/init.c
+++ b/kadmin/init.c
@@ -52,11 +52,6 @@ create_random_entry(krb5_principal princ,
     krb5_keyblock *keys;
     int n_keys, i;
     char *name;
-    const char *password;
-    char pwbuf[512];
-
-    random_password(pwbuf, sizeof(pwbuf));
-    password = pwbuf;
 
     ret = krb5_unparse_name(context, princ, &name);
     if (ret) {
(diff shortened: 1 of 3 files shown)


lhoward:
Ah, my tests were failing owing to a local configuration issue. (The previous commit to master was failing because I forgot to export a symbol.)

lhoward:
Will merge once tests here pass.

kadmin/init.c
```diff
@@ -76,14 +71,14 @@ create_random_entry(krb5_principal princ,
 	mask |= KADM5_MAX_RLIFE;
     }
     ent.attributes |= attributes | KRB5_KDB_DISALLOW_ALL_TIX;
-    mask |= KADM5_ATTRIBUTES;
+    mask |= KADM5_ATTRIBUTES | KADM5_KEY_DATA;
 
-    /* Create the entry with a random password */
-    ret = kadm5_create_principal(kadm_handle, &ent, mask, password);
+    /* Create the entry with no keys or password */
+    ret = kadm5_s_create_principal_with_key(kadm_handle, &ent, mask);
```
nicowilliams: Does that actually randomize keys?  Somehow I can't tell from code inspection!

lhoward: `kadm5_randkey_principal() is called afterwards, before `KRB5_KDB_DISALLOW_ALL_TIX` is removed.

Issue #469, opened 2018-12-26
kpasswd silently fails if passwords mismatch


lhoward:
`UI_UTIL_read_pw_string()` failing to match the new passwords results in nothing being printed to the user. This is confusing. Most `passwd` programs will print an error.

```
% kpasswd
foobar@LUKKTONE.COM's Password: 
New password: 
Verify password - New password: 
% 
```

lhoward:
Exit code is 1 though

lhoward:
`UI_UTIL_read_pw_string()` may need a distinct error code for password mismatch

lhoward:
Looking at the OpenSSL code that `UI_UTIL_read_pw_string()` purports to emulate, the verify failure message should be reported by the function itself rather than the caller.

lhoward:
merged

lhoward:
merged in 65ed504d

lhoward:
This fix now causes kadmin to double report password mismatches

```
kadmin> cpw lukeh
lukeh@LUKKTONE.COM's Password: 
Verify password - lukeh@LUKKTONE.COM's Password: 
Verify failure
kadmin: cpw lukeh: Password mismatch
```

nicowilliams:
OpenSSL did add the third argument, and they call it `verify`.  I don't mind the extension.

lhoward:
Fixed again in 014f1688

PR #479, opened 2018-12-27
hcrypto: print failure on password mismatch (#469)

lhoward:
UI_UTIL_read_pw_string(), an interface borrowed from OpenSSL, should report password verification failure to stderr.


nicowilliams:
Strange that we call `UI_UTIL_read_pw_string()` from kinit directly...

PR #483, opened 2018-12-27
Fix check-ldap race condition (fix #482)

### LINKED ISSUE #482: check-ldap is racy

nicowilliams:
From the commit:

```
We start slapd in the foreground (-d0) but backgrounded in the shell,
then we wait 4 seconds.  This causes a race condition however.  This
commit makes the slapd-init script more robust and limits the wait to
however many seconds (up to 30) that slapd needs to start service.
```

The fix is to loop up to N times checking that `slapd` is responsive, sleeping a second in between.


nicowilliams:
Ah, yes, we'll need to also install `ldap-utils`, otherwise the Travis builds won't be able to run `check-ldap`.

PR #485, opened 2018-12-28
Fix VC warnings

nicowilliams:
These commits fix a bunch of warnings from VC.

I'd like to figure out how to declare printf-like functions to VC as well.  So far I've found how to tell VC that an argument is a format string, but not where the remaining arguments start, or that the arguments are in a `va_list` passed in.


nicowilliams:
Well, VC in Appveyor did not like `__declspec(noreturn)`.  @jaltman, please help.

lib/krb5/test_acl.c
```diff
@@ -37,7 +37,7 @@
 	do { if (r != e) krb5_errx(c, 1, "%s", s); } while (0)
 #define STRINGMATCH(c, s, _s1, _s2) \
 	do {							\
-		if (_s1 == NULL || _s2 == NULL) 		\
+		if (_s1 == NULL)				\
```
nicowilliams: Exclude this fix.

jaltman:
3503507 is fixing a bug.  its not simply a warning suppression.  I'd prefer it if the commit message described the bug fix.

lib/krb5/test_acl.c
jaltman: ok, but what was it supposed to be fixing?

nicowilliams:
@jaltman I'm not sure how that bug manifests...  I don't understand why gcc/clang didn't complain about that, or how it's just a warning for VC -- it should have been an error!

lib/krb5/test_acl.c
nicowilliams: This macro is only ever used with a constant string literal as `_s2`, so VC complains that that comparison to `NULL` is pointless.

kadmin/check.c
```diff
@@ -90,7 +90,7 @@ do_check_entry(krb5_principal principal, void *data)
 	    krb5_warnx(context,
 		       "Principal %s enctype %d, wrong length: %lu\n",
 		       name, princ.key_data[i].key_data_type[0],
-		       (unsigned long)princ.key_data[i].key_data_length);
+		       (unsigned long)princ.key_data[i].key_data_length[0]);
```
jaltman: could you remove the cast to `unsigned long` and change the format string to use `%hu` instead of `%lu` as the type is `int16_t`?   The cast was masking the prior use of the pointer value as an integer.

nicowilliams: I'll change the `%lu` to `%d`, since `princ.key_data[i].key_data_length[0]` is an `int16_t` that will get promoted to `int` anyways.

lib/krb5/test_acl.c
jaltman: Thanks for the explanation

nicowilliams:
Should be fixed now.

nicowilliams:
It is fixed now.  Tests will pass; please merge when done.

PR #488, opened 2018-12-29
SPNEGO: Use error tokens (init. side) (fix #486)

### LINKED ISSUE #486: SPNEGO does not allow mechanism error tokens in reject responses
Presuming I have read RFC4178 correctly and it is OK to return a responseToken with a reject response –

A SPNEGO negState of reject causes context initialization to abort before the mechanism has a chance to process any context error tokens.

Related to – https://github.com/krb5/krb5/pull/885

nicowilliams:
This is just the initiator side, and only about consuming error tokens.  I've not looked at the acceptor side nor whether we have error token non-sending issues as well.

### DIFF
--- a/lib/gssapi/spnego/accept_sec_context.c
+++ b/lib/gssapi/spnego/accept_sec_context.c
@@ -35,10 +35,12 @@
 
 static OM_uint32
 send_reject (OM_uint32 *minor_status,
+	     gss_buffer_t mech_token,
 	     gss_buffer_t output_token)
 {
     NegotiationToken nt;
     size_t size;
+    heim_octet_string responseToken;
 
     nt.element = choice_NegotiationToken_negTokenResp;
 
@@ -49,7 +51,13 @@ send_reject (OM_uint32 *minor_status,
     }
     *(nt.u.negTokenResp.negResult)  = reject;
     nt.u.negTokenResp.supportedMech = NULL;
-    nt.u.negTokenResp.responseToken = NULL;
+
+    if (mech_token != GSS_C_NO_BUFFER) {
+	responseToken.length = mech_token->length;
+	responseToken.data   = mech_token->value;
+	nt.u.negTokenResp.responseToken = &responseToken;
+    } else
+	nt.u.negTokenResp.responseToken = NULL;
     nt.u.negTokenResp.mechListMIC   = NULL;
 
     ASN1_MALLOC_ENCODE(NegotiationToken,
@@ -454,7 +462,7 @@ acceptor_complete(OM_uint32 * minor_status,
 	    ret = verify_mechlist_mic(minor_status, ctx, mech_buf, mic);
 	    if (ret) {
 		if (*get_mic)
-		    send_reject (minor_status, output_token);
+		    send_reject (minor_status, GSS_C_NO_BUFFER, output_token);
 		return ret;
 	    }
 	    ctx->verified_mic = 1;
@@ -774,7 +782,7 @@ acceptor_continue
 	    if (ret != GSS_S_COMPLETE && ret != GSS_S_CONTINUE_NEEDED) {
 		free_NegotiationToken(&nt);
 		gss_mg_collect_error(ctx->negotiated_mech_type, ret, minor);
-		send_reject (minor_status, output_token);
+		send_reject (minor_status, mech_output_token, output_token);
 		HEIMDAL_MUTEX_unlock(&ctx->ctx_id_mutex);
 		return ret;
 	    }
--- a/lib/gssapi/spnego/init_sec_context.c
+++ b/lib/gssapi/spnego/init_sec_context.c
@@ -501,7 +501,6 @@ spnego_reply
 				   &ctx->mech_time_rec);
 	if (GSS_ERROR(ret)) {
 	    HEIMDAL_MUTEX_unlock(&ctx->ctx_id_mutex);
-	    free_NegotiationToken(&resp);
             /*
              * If the acceptor rejected, preserve the minor status information
              * from the mechanism, but indicate GSS_S_BAD_MECH.
@@ -509,6 +508,7 @@ spnego_reply
             if (resp.u.negTokenResp.negResult != NULL &&
                 *resp.u.negTokenResp.negResult == reject)
                 ret = GSS_S_BAD_MECH;
+	    free_NegotiationToken(&resp);
 	    gss_mg_collect_error(&mech, ret, minor);
 	    *minor_status = minor;
 	    return ret;
@@ -517,8 +517,12 @@ spnego_reply
          * If the acceptor rejected, we're out even if the inner context is
          * now complete.  Note that the rejection is not integrity-protected.
          */
-        if (resp.u.negTokenResp.negResult != NULL &&
-            *resp.u.negTokenResp.negResult == reject) {
+        if (resp.u.negTokenResp.negResult == NULL) {
+            HEIMDAL_MUTEX_unlock(&ctx->ctx_id_mutex);
+            free_NegotiationToken(&resp);
+            return GSS_S_DEFECTIVE_TOKEN;
+        }
+        if (*resp.u.negTokenResp.negResult == reject) {
             HEIMDAL_MUTEX_unlock(&ctx->ctx_id_mutex);
             free_NegotiationToken(&resp);
             return GSS_S_BAD_MECH;
--- a/lib/gssapi/spnego/init_sec_context.c
+++ b/lib/gssapi/spnego/init_sec_context.c
@@ -415,18 +415,10 @@ spnego_reply
     if (ret)
       return ret;
 
+    /* The SPNEGO token must be a negTokenResp */
     if (resp.element != choice_NegotiationToken_negTokenResp) {
 	free_NegotiationToken(&resp);
 	*minor_status = 0;
-	return GSS_S_BAD_MECH;
-    }
-
-    if (resp.u.negTokenResp.negResult == NULL
-	|| *(resp.u.negTokenResp.negResult) == reject
-	/* || resp.u.negTokenResp.supportedMech == NULL */
-	)
-    {
-	free_NegotiationToken(&resp);
 	return GSS_S_BAD_MECH;
     }
 
@@ -475,7 +467,7 @@ spnego_reply
 	return GSS_S_BAD_MECH;
     }
 
-    /* if a token (of non zero length), or no context, pass to underlaying mech */
+    /* if a token (of non zero length) pass to underlaying mech */
     if ((resp.u.negTokenResp.responseToken != NULL && resp.u.negTokenResp.responseToken->length) ||
 	ctx->negotiated_ctx_id == GSS_C_NO_CONTEXT) {
 	gss_buffer_desc mech_input_token;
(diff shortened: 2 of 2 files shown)


lib/gssapi/spnego/init_sec_context.c
```diff
@@ -510,14 +502,37 @@ spnego_reply
 	if (GSS_ERROR(ret)) {
 	    HEIMDAL_MUTEX_unlock(&ctx->ctx_id_mutex);
 	    free_NegotiationToken(&resp);
+            /*
+             * If the acceptor rejected, preserve the minor status information
+             * from the mechanism, but indicate GSS_S_BAD_MECH.
+             */
+            if (resp.u.negTokenResp.negResult != NULL &&
+                *resp.u.negTokenResp.negResult == reject)
+                ret = GSS_S_BAD_MECH;
```
lhoward: Clobbering ret in this case gives different behavior to had the application invoked the mechanism directly. Suggest honoring mechanism return code.

lhoward: Also, you need to move free_NegotiationToken() after this

lib/gssapi/spnego/init_sec_context.c
```diff
@@ -415,21 +415,13 @@ spnego_reply
...
       return ret;
 
+    /* The SPNEGO token must be a negTokenResp */
     if (resp.element != choice_NegotiationToken_negTokenResp) {
 	free_NegotiationToken(&resp);
 	*minor_status = 0;
 	return GSS_S_BAD_MECH;
     }
 
-    if (resp.u.negTokenResp.negResult == NULL
```
lhoward: We still need to handle `resp.u.negTokenResp.negResult == NULL` somewhere

lib/gssapi/spnego/init_sec_context.c
```diff
@@ -510,14 +502,37 @@ spnego_reply
...
 	    *minor_status = minor;
 	    return ret;
 	}
+        /*
+         * If the acceptor rejected, we're out even if the inner context is
+         * now complete.  Note that the rejection is not integrity-protected.
+         */
+        if (resp.u.negTokenResp.negResult != NULL &&
+            *resp.u.negTokenResp.negResult == reject) {
+            HEIMDAL_MUTEX_unlock(&ctx->ctx_id_mutex);
```
lhoward: maybe `resp.u.negTokenResp.negResult == NULL || *resp.u.negTokenResp.negResult == reject`

lib/gssapi/spnego/init_sec_context.c
nicowilliams: But the acceptor at the SPNEGO layer said `reject`, meaning they rejected the _mechanism_.  This would only happen on the first round trip.  Any more legs and I think it's not sensible to reject the choice of mechanism -- you've already committed to it by then.

nicowilliams: Oy, right about the free.

lib/gssapi/spnego/init_sec_context.c
nicowilliams: Yes.  What's it mean?  :)

lib/gssapi/spnego/init_sec_context.c
nicowilliams: Sure.

lib/gssapi/spnego/init_sec_context.c
nicowilliams: `GSS_S_DEFECTIVE_TOKEN`

lib/gssapi/spnego/init_sec_context.c
lhoward: If the acceptor is rejecting the *choice* then it should omit the responseToken.

lib/gssapi/spnego/init_sec_context.c
lhoward: reject :)

lib/gssapi/spnego/init_sec_context.c
nicowilliams: Pushed a fix.  I'm leaving it as `GSS_S_BAD_MECH`.  If we have a counter of legs I could make it do that only on the second.

lib/gssapi/spnego/init_sec_context.c
nicowilliams: I went with something slightly different.

lib/gssapi/spnego/init_sec_context.c
lhoward: I presume... BTW is that misnamed or from an earlier version? RFC 4178 says it's called negState. It's also only required on the first reply from the acceptor. So, hmm, maybe I need to look at this more carefully – if it is absent we should still process a context token that may be an error token.

lib/gssapi/spnego/accept_sec_context.c
```diff
@@ -454,7 +462,7 @@ acceptor_complete(OM_uint32 * minor_status,
 	    ret = verify_mechlist_mic(minor_status, ctx, mech_buf, mic);
 	    if (ret) {
 		if (*get_mic)
-		    send_reject (minor_status, output_token);
+		    send_reject (minor_status, GSS_C_NO_BUFFER, output_token);
```
nicowilliams: Why `GSS_C_NO_BUFFER` in this case?

nicowilliams: Ah, because we don't have one yet.

nicowilliams: No, we do have one, but I guess we don't want to complete the inner context.  Good.

nicowilliams:
@lhoward your commit LGTM.

lib/gssapi/spnego/init_sec_context.c
lhoward: the local negState variable may be overkill, just looked cleaner to me. Presumably the ctx->open and else branches can't be reached, unless you can restart negotiation (which I don't think you can do with SPNEGO?)

lhoward:
Deleted my diff. I need to look at it when I'm less tired.

lhoward:
```c
diff --git a/lib/gssapi/spnego/init_sec_context.c b/lib/gssapi/spnego/init_sec_context.c
index 006f3ef59..41ed64b76 100644
--- a/lib/gssapi/spnego/init_sec_context.c
+++ b/lib/gssapi/spnego/init_sec_context.c
@@ -396,6 +396,7 @@ spnego_reply
     gss_buffer_desc mic_buf, mech_buf;
     gss_buffer_desc mech_output_token;
     gssspnego_ctx ctx;
+    int negState;
 
     *minor_status = 0;
 
@@ -415,21 +416,13 @@ spnego_reply
     if (ret)
       return ret;
 
+    /* The SPNEGO token must be a negTokenResp */
     if (resp.element != choice_NegotiationToken_negTokenResp) {
 	free_NegotiationToken(&resp);
 	*minor_status = 0;
 	return GSS_S_BAD_MECH;
     }
 
-    if (resp.u.negTokenResp.negResult == NULL
-	|| *(resp.u.negTokenResp.negResult) == reject
-	/* || resp.u.negTokenResp.supportedMech == NULL */
-	)
-    {
-	free_NegotiationToken(&resp);
-	return GSS_S_BAD_MECH;
-    }
-
     /*
      * Pick up the mechanism that the acceptor selected, only allow it
      * to be sent in packet.
@@ -437,8 +430,15 @@ spnego_reply
 
-    if (resp.u.negTokenResp.supportedMech) {
+    /* negResult is required on the first response only, else inferred */
+    if (resp.u.negTokenResp.negResult != NULL)
+	negState = *resp.u.negTokenResp.negResult;
+    else if (ctx->open)
+	negState = accept_completed;
+    else
+	negState = accept_incomplete;
 
+    if (resp.u.negTokenResp.supportedMech) {
 	if (ctx->oidlen) {
 	    free_NegotiationToken(&resp);
 	    HEIMDAL_MUTEX_unlock(&ctx->ctx_id_mutex);
@@ -509,20 +509,39 @@ spnego_reply
 				   &ctx->mech_time_rec);
 	if (GSS_ERROR(ret)) {
 	    HEIMDAL_MUTEX_unlock(&ctx->ctx_id_mutex);
+            /*
+             * If the acceptor rejected, preserve the minor status information
+             * from the mechanism, but indicate GSS_S_BAD_MECH.
+             */
 	    free_NegotiationToken(&resp);
 	    gss_mg_collect_error(&mech, ret, minor);
 	    *minor_status = minor;
 	    return ret;
 	}
+        /*
+         * If the acceptor rejected, we're out even if the inner context is
+         * now complete.  Note that the rejection is not integrity-protected.
+         */
+	if (negState == reject) {
+            HEIMDAL_MUTEX_unlock(&ctx->ctx_id_mutex);
+            free_NegotiationToken(&resp);
+            return GSS_S_BAD_MECH;
+        }
+
 	if (ret == GSS_S_COMPLETE) {
 	    ctx->open = 1;
 	}
-    } else if (*(resp.u.negTokenResp.negResult) == accept_completed) {
+    } else if (negState == accept_completed) {
+        /*
+         * Note that the accept_completed isn't integrity-protected, but
+         * ctx->maybe_open can only be true if the inner context is fully
+         * established.
+         */
 	if (ctx->maybe_open)
 	    ctx->open = 1;
     }
 
-    if (*(resp.u.negTokenResp.negResult) == request_mic) {
+    if (negState == request_mic) {
 	ctx->require_mic = 1;
     }
 
```

lhoward:
Take 3.

```udiff
diff --git a/lib/gssapi/spnego/init_sec_context.c b/lib/gssapi/spnego/init_sec_context.c
index 006f3ef59..c4bbfd4b2 100644
--- a/lib/gssapi/spnego/init_sec_context.c
+++ b/lib/gssapi/spnego/init_sec_context.c
@@ -396,6 +396,7 @@ spnego_reply
     gss_buffer_desc mic_buf, mech_buf;
     gss_buffer_desc mech_output_token;
     gssspnego_ctx ctx;
+    int negState;
 
     *minor_status = 0;
 
@@ -415,21 +416,13 @@ spnego_reply
     if (ret)
       return ret;
 
+    /* The SPNEGO token must be a negTokenResp */
     if (resp.element != choice_NegotiationToken_negTokenResp) {
 	free_NegotiationToken(&resp);
 	*minor_status = 0;
 	return GSS_S_BAD_MECH;
     }
 
-    if (resp.u.negTokenResp.negResult == NULL
-	|| *(resp.u.negTokenResp.negResult) == reject
-	/* || resp.u.negTokenResp.supportedMech == NULL */
-	)
-    {
-	free_NegotiationToken(&resp);
-	return GSS_S_BAD_MECH;
-    }
-
     /*
      * Pick up the mechanism that the acceptor selected, only allow it
      * to be sent in packet.
@@ -437,8 +430,15 @@ spnego_reply
 
-    if (resp.u.negTokenResp.supportedMech) {
+    /* negResult is required on the first response only, else inferred */
+    if (resp.u.negTokenResp.negResult != NULL)
+	negState = *resp.u.negTokenResp.negResult;
+    else if (ctx->open)
+	negState = accept_completed;
+    else
+	negState = accept_incomplete;
 
+    if (resp.u.negTokenResp.supportedMech) {
 	if (ctx->oidlen) {
 	    free_NegotiationToken(&resp);
 	    HEIMDAL_MUTEX_unlock(&ctx->ctx_id_mutex);
@@ -514,15 +514,30 @@ spnego_reply
 	    *minor_status = minor;
 	    return ret;
 	}
+        /*
+         * If the acceptor rejected, we're out even if the inner context is
+         * now complete. Note that the rejection is not integrity-protected.
+         */
+	if (negState == reject) {
+            HEIMDAL_MUTEX_unlock(&ctx->ctx_id_mutex);
+            free_NegotiationToken(&resp);
+            return GSS_S_BAD_MECH;
+        }
+
 	if (ret == GSS_S_COMPLETE) {
 	    ctx->open = 1;
 	}
-    } else if (*(resp.u.negTokenResp.negResult) == accept_completed) {
+    } else if (negState == accept_completed) {
+        /*
+         * Note that the accept_completed isn't integrity-protected, but
+         * ctx->maybe_open can only be true if the inner context is fully
+         * established.
+         */
 	if (ctx->maybe_open)
 	    ctx->open = 1;
     }
 
-    if (*(resp.u.negTokenResp.negResult) == request_mic) {
+    if (negState == request_mic) {
 	ctx->require_mic = 1;
     }
 
```

lhoward:
Also `ctx->open` shouldn't be a valid state entering `spnego_reply()`.

lhoward:
So maybe instead of setting `negState = accept_completed` we should return an error there.

nicowilliams:
Can you push your commits to my branch?

lhoward:
Have I permission?

lhoward:
OK, I pushed your branch + my diff to heimdal:nicowilliams/fix_486

lhoward:
Just was reminded of this. I can take a look vs current code.

lhoward:
Fixed in master.

PR #489, opened 2018-12-29
kadm5: refactor krb5 plugin framework

lhoward:
WIP

### DIFF
--- a/kdc/windc.c
+++ b/kdc/windc.c
@@ -39,6 +39,21 @@ static int have_plugin = 0;
  * Pick the first WINDC module that we find.
  */
 
+static const char *windc_plugin_deps[] = {
+    "kdc",
+    "krb5",
+    "hdb",
+    NULL
+};
+
+static struct krb5_plugin_data windc_plugin_data = {
+    "krb5",
+    "windc",
+    KRB5_WINDC_PLUGIN_MINOR,
+    windc_plugin_deps,
+    kdc_get_instance
+};
+
 static krb5_error_code KRB5_LIB_CALL
 load(krb5_context context, const void *plug, void *plugctx, void *userctx)
 {
@@ -49,8 +64,8 @@ load(krb5_context context, const void *plug, void *plugctx, void *userctx)
 krb5_error_code
 krb5_kdc_windc_init(krb5_context context)
 {
-    (void)_krb5_plugin_run_f(context, "krb5", "windc",
-			     KRB5_WINDC_PLUGIN_MINOR, 0, NULL, load);
+    (void)_krb5_plugin_run_f(context, &windc_plugin_data, 0, NULL, load);
+
     return 0;
 }
 
@@ -84,8 +99,8 @@ _kdc_pac_generate(krb5_context context,
     uc.client = client;
     uc.pac = pac;
 
-    (void)_krb5_plugin_run_f(context, "krb5", "windc",
-			     KRB5_WINDC_PLUGIN_MINOR, 0, &uc, generate);
+    (void)_krb5_plugin_run_f(context, &windc_plugin_data,
+			     0, &uc, generate);
     return 0;
 }
 
@@ -141,8 +156,8 @@ _kdc_pac_verify(krb5_context context,
     uc.pac = pac;
     uc.verified = verified;
 
-    (void)_krb5_plugin_run_f(context, "krb5", "windc",
-			     KRB5_WINDC_PLUGIN_MINOR, 0, &uc, verify);
+    (void)_krb5_plugin_run_f(context, &windc_plugin_data,
+			     0, &uc, verify);
     return 0;
 }
 
@@ -191,8 +206,8 @@ _kdc_check_access(krb5_context context,
         uc.req = req;
         uc.method_data = method_data;
 
-        ret = _krb5_plugin_run_f(context, "krb5", "windc",
-                                 KRB5_WINDC_PLUGIN_MINOR, 0, &uc, check);
+        ret = _krb5_plugin_run_f(context, &windc_plugin_data,
+                                 0, &uc, check);
     }
 
     if (ret == KRB5_PLUGIN_NO_HANDLE)
@@ -201,4 +216,19 @@ _kdc_check_access(krb5_context context,
 			       server_ex, server_name,
 			       req->msg_type == krb_as_req);
     return ret;
+}
+
+uintptr_t
+kdc_get_instance(const char *libname)
+{
+    static const char *instance = "libkdc";
+
+    if (strcmp(libname, "kdc") == 0)
+        return (uintptr_t)instance;
+    else if (strcmp(libname, "hdb") == 0)
+	return hdb_get_instance(libname);
+    else if (strcmp(libname, "krb5") == 0)
+        return krb5_get_instance(libname);
+
+    return 0;
 }
--- a/lib/gssapi/mech/gss_mech_switch.c
+++ b/lib/gssapi/mech/gss_mech_switch.c
@@ -314,14 +314,6 @@ _gss_load_mech(void)
 		if (found)
 			continue;
 
-#ifndef RTLD_LOCAL
-#define RTLD_LOCAL 0
-#endif
-
-#ifndef RTLD_GROUP
-#define RTLD_GROUP 0
-#endif
-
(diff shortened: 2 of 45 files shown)


lhoward:
@nicowilliams, @jaltman what about the library that consumes the plugin exporting `XXX_get_instance()`, the result of which is returned by the plugin's initialization function. So for example for the hook/password quality functions:

```c
uintptr_t
kdb5_get_instance(void)
{
    static char x;
    /* for all deps that expose their ABI to the plugin */
    return (uintptr_t)&x ^ hdb_get_instance() ^ krb5_get_instance();
}
```

Are we relying on the dynamic linker to deal with ABIs changing otherwise, e.g. every time we change a structure size are we bumping the SONAME? I haven't quite got my head around whether we would need to mix in a compile-time constant too.

lhoward:
BTW, I'd kind of prefer for the plugin consumer to validate the instance IDs, so that we don't have to trust that the plugin remembers to do it.

lhoward:
Do we want to do this for new plugins only or to the libkrb5 plugin API? Do you want to bump the version and break things? Are any of these plugins API compatible with MIT?

lhoward:
I'm thinking possibly about separating out plugin `load` and `init`; new plugins would export a `load` function which would return a consumer instance value and a dispatch table, and `init` would move back into the dispatch table. Doesn't make much difference for kadm5 hooks, but it would mean that we could support both old and new plugin types easily in libkrb5 simply by looking for different symbol names (e.g. `foo` vs `foo_plugin_v1_load` or something).

lhoward:
The principal advantage of this is minimal intrusiveness to both libkrb5 and plugins, plugins can support both old and new mechanisms, we move to an exported function rather than data symbol for registration.

lhoward:
Having the plugin load function take a plugin version as input is extraneous if we require plugins to be built against the consuming version of Heimdal, because there should be a compile time error if they implement a plugin API that is not supported. (The original intention was forwards compatibility, i.e. to allow plugins built against newer libraries to be loaded by older ones by returning a different dispatch table depending on the input version number.)

I will remove this.

Separating the load and init functions also allows a plugin to return multiple dispatch tables, which could be useful, so I'll add this functionality.

nicowilliams:
Why the XOR?

lhoward:
> Why the XOR?

I was looking for a poor man's `RTLD_GROUP`, an attempt to indicate a dependency chain in a single uintptr_t. But I didn't think about it too deeply.

lhoward:
Can we kill `_krb5_plugin_find()` and `_krb5_plugin_free()`? Nothing uses it, they are in fact useless because the structure they return is private to their compilation unit and the accessor for it (`_krb5_plugin_find()`) has been removed.

lhoward:
I've also refactored the kadm5 server hook subsystem to also use the libkrb5 plugin API.

nicowilliams:
Wow!  This is a large one.  It will take a few days to review.

lhoward:
> Wow! This is a large one. It will take a few days to review.

Thanks @nicowilliams, it might be easier to look at lib/krb5/plugin.c fresh rather than as a diff.

kdc/windc.c
```diff
@@ -202,3 +206,14 @@ _kdc_check_access(krb5_context context,
 			       req->msg_type == krb_as_req);
     return ret;
 }
+
+/*
+ * Instance value should capture all libraries that export data structures
+ * that are used by the plugin so, in this case, libkdc, libhdb and libkrb5.
+ */
+uintptr_t
+kdc_get_instance(void)
```
nicowilliams: I would prefer this take a `const char *libname` and return the corresponding library's "instance".

lib/gssapi/mech/gss_mech_switch.c
```diff
@@ -314,14 +314,6 @@ _gss_load_mech(void)
 		if (found)
 			continue;
 
-#ifndef RTLD_LOCAL
```
nicowilliams: You should keep these.  Just because we don't always get them doesn't mean we shouldn't try to.

nicowilliams: Oh, never mind.

kdc/windc.c
lhoward: How would you envisage this working with a common plugin load/initialize API between subsystems, i.e. where the plugin can only return a single instance to the consuming API?

Also, where should that API live? If it's in libkrb5, then it can't link against libkdc or libkadm5 (for example), in order to return those instances.

lhoward: Goals were:

* a single plugin load/initialize API for all (well, most) plugins – i.e. libkrb5, libkdc, libhdb, libkadm5
* instance validation done by the consuming library so plugins cannot bypass it (i.e. plugin returns instance value at load time)
* no change in existing library dependencies
* instance value (attempts to) capture instances of dependent libraries, where dependent is defined as being all libraries that expose structures to the plugin (as opposed to the actual library dependency chain). Hence XOR.

lhoward: I suppose we could have a _krb5_register_instance() API that other APIs could use. But I kind of like the idea that the plugin only has to lookup one cookie to return to the loader.

lhoward: OK, I have some ideas...

nicowilliams: Either the caller, or the plugin, should give the other a function _pointer_ of a get-instance function (not some value returned by such a function).  Then the one receiving it can call it and check that the instances match.

lhoward: OK, the plugin load function now looks like this –

```c
typedef krb5_error_code
(KRB5_CALLCONV *krb5_plugin_load_t)(krb5_context context,
                                    krb5_get_instance_func *func,
                                    size_t *n_deps,
                                    const char *const **deps,
                                    size_t *n_plugins,
                                    const common_plugin_ftable *const **plugins);
```

It's a bit unwieldy, but it works. Other options would be to just have the caller (consumer) provide a get_instance function and have the plugin do the validation itself; I mean, it can avoid any validation by omitting an instance list (well, almost, we will always check krb5).

lhoward: But I think providing a list of deps is less error prone than copying boilerplate code to check the instances.

lhoward: Fixed

```c
KRB5_LIB_FUNCTION uintptr_t KRB5_LIB_CALL
krb5_get_instance(krb5_context context, const char *module);
```

nicowilliams:
3947025 looks good.

lib/krb5/plugin.c
```diff
@@ -33,34 +35,165 @@
...
+ * Some existing plugin consumers outside libkrb5 use the "krb5" module
+ * namespace, but going forward the module should match the consumer library
+ * name (e.g. libhdb should use the "hdb" module rather than "krb5").
+ */
+
+/* global module use, use copy_modules() accessor to access */
+static heim_dict_t __modules;
+
+/* for updating __modules; NB retain/release is atomic and does not need lock */
+static HEIMDAL_MUTEX modules_mutex = HEIMDAL_MUTEX_INITIALIZER;
```
nicowilliams: N.B., this doesn't work.  There's a race condition between the read of the global and the taking of a reference to the object.  See https://github.com/nicowilliams/ctp

IIUC we already had this problem, so I might allow it for now and immediately do something about it.

lib/krb5/plugin.c
```diff
@@ -33,34 +35,165 @@
...
+
+/* returns global modules list, refcount +1 */
+static heim_dict_t
+copy_modules(void)
+{
+    if (__modules == NULL)
+	__modules = heim_dict_create(11);
```
nicowilliams: This should be done with `heim_once`.

lib/krb5/plugin.c
```diff
@@ -33,34 +35,165 @@
...
+
+    heim_assert(__modules, "plugin modules array allocation failure");
+
+    return heim_retain(__modules);
```
nicowilliams: And here's the race.

lib/krb5/plugin.c
nicowilliams: One thing we could do for now is always take the lock even to read (well, to retain the reference).

nicowilliams: Or, if we never replace the `__modules`, then just use a `heim_once` to initialize it.

nicowilliams: Ah, you only ever write `__modules` once.

Issue #383, opened 2018-05-14
MIT has /var/kerberos/krb5/user/${EUID}/client.keytab


jaltman:
According to a presentation at HEPiX, MIT Kerberos will silently perform a `kinit -k` if there is no valid TGT and a keytab is available for the user at `/var/kerberos/krb5/user/${EUID}/client.keytab`

jcpunk:
Documented upstream at: https://web.mit.edu/kerberos/krb5-1.15/doc/mitK5defaults.html (see bottom of page)

kaduk:
https://web.mit.edu/kerberos/krb5-1.15/doc/basic/keytab_def.html#default-client-keytab has some (minimal) feature documentation that indicates the scope of the feature.

PR #491, opened 2018-12-31
gssapi: default client keytab (#383)

lhoward:
WIP for #383 – not sure if fix should be this simple?

### DIFF
--- a/lib/gssapi/krb5/acquire_cred.c
+++ b/lib/gssapi/krb5/acquire_cred.c
@@ -58,7 +58,7 @@ __gsskrb5_ccache_lifetime(OM_uint32 *minor_status,
 
 
 static krb5_error_code
-get_keytab(krb5_context context, krb5_keytab *keytab)
+get_system_keytab(krb5_context context, krb5_keytab *keytab)
 {
     krb5_error_code kret;
 
@@ -78,6 +78,33 @@ get_keytab(krb5_context context, krb5_keytab *keytab)
     HEIMDAL_MUTEX_unlock(&gssapi_keytab_mutex);
 
     return (kret);
+}
+
+static krb5_error_code
+get_client_keytab(krb5_context context,
+		  krb5_const_principal principal,
+		  krb5_keytab *keytab)
+{
+    krb5_error_code ret;
+    char *name = NULL;
+
+    ret = _krb5_kt_client_default_name(context, &name);
+    if (ret == 0)
+	ret = krb5_kt_resolve(context, name, keytab);
+    if (ret == 0 && principal) {
+	krb5_keytab_entry entry;
+
+	ret = krb5_kt_get_entry(context, *keytab, principal,
+				 0, 0, &entry);
+	if (ret == 0)
+	    krb5_kt_free_entry(context, &entry);
+    }
+    krb5_xfree(name);
+
+    if (ret)
+	ret = get_system_keytab(context, keytab);
+
+    return ret;
 }
 
 /*
@@ -292,7 +319,7 @@ try_keytab:
         if (kret)
             goto end;
     }
-    kret = get_keytab(context, &keytab);
+    kret = get_client_keytab(context, handle->principal, &keytab);
     if (kret)
         goto end;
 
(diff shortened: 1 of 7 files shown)


lib/gssapi/krb5/acquire_cred.c
```diff
@@ -80,6 +80,34 @@ get_keytab(krb5_context context, krb5_keytab *keytab)
...
+static krb5_error_code
+get_client_keytab(krb5_context context,
+		  krb5_const_principal principal,
+		  krb5_keytab *keytab)
+{
+    krb5_error_code kret;
+    krb5_keytab_entry entry;
+    char *name = NULL;
+
+    kret = _krb5_kt_client_default_name(context, &name);
```
nicowilliams: Not a required change, but you can deindent a lot like so:

```
    kret = _krb5_kt_client_default_name(context, &name);
    if (kret == 0)
        kret = krb5_kt_resolve(context, name, keytab);
    if (kret == 0 && principal != NULL)
	kret = krb5_kt_get_entry(context, *keytab, principal,
				 0, 0, &entry);
    if (kret == 0)
	krb5_kt_free_entry(context, &entry);
    krb5_xfree(name);
    if (kret == 0)
        return 0;

    return get_keytab(context, keytab); /* fall back on regular keytab */
```

lib/gssapi/krb5/acquire_cred.c
```diff
@@ -80,6 +80,34 @@ get_keytab(krb5_context context, krb5_keytab *keytab)
...
+    if (kret == 0 && principal != NULL) {
+	kret = krb5_kt_get_entry(context, *keytab, principal,
+				 0, 0, &entry);
+	if (kret == 0)
+	    krb5_kt_free_entry(context, &entry);
+    }
+    krb5_xfree(name);
+
+    if (kret != 0)
+	kret = get_keytab(context, keytab);
```
nicowilliams: If a principal name was given, do we not want to check if this keytab has entries for it?

lib/gssapi/krb5/acquire_cred.c
lhoward: Yeah, thought about that, wasn't convinced it was necessarily more readable. But happy to.

lib/gssapi/krb5/acquire_cred.c
lhoward: Probably but – didn't want to change current behavior. Hmm, does Heimdal do the MIT thing where rd_req tries all the keys in the keytab? We'd have to be careful we don't break that (if it does that).

lhoward: Don't think it'll make much difference though for the client, it'd just fail later on.

lhoward: Hmm I guess we don’t handle the case where we don’t know the client principal name. Todo.

lhoward:
@nicowilliams the only thing with this patch (well, with the existing Heimdal code), is that if you don't specify a principal when acquiring a GSS credential, it uses `krb5_get_default_principal()`. Arguably if there's a client keytab, we should use the first principal we find there. Or perhaps not. Anyway, if you think this is a worthwhile change I can make it.

lhoward:
Excerpt below, see with `get_client_keytab()` I've just (lazily!) hooked into the existing keytab path.

```c
try_keytab:
    if (handle->principal == NULL) {
        /* We need to know what client principal to use */
        kret = krb5_get_default_principal(context, &handle->principal);
        if (kret)
            goto end;
    }
    kret = get_client_keytab(context, handle->principal, &keytab);
    if (kret)
        goto end;
```

lhoward:
But there's no canonical principal for a keytab so...

nicowilliams:
@lhoward It's very tricky...

kinit with a keytab (or `gss_acquire_cred()` with a keytab) and no other source of a principal name can do things like:

- try `${USER}` as the principal name
- try the first/last keytab entry for a principal that has one component
- try the first/last keytab entry

I say "last" mostly because keytabs get appended to, and rarely get re-sorted.  This is a very obnoxious thing about keytabs...

Ideally things like kadmin's `ext_keytab` and ktutil should re-sort the keytab to put the newest keytab entries first, which helps performance of `krb5_rd_req*()` and `gss_accept_sec_context()`!  But since the order in which we try things is meaningful (see above), this strikes me as a notable behavior change, perhaps even a backwards-incompatible behavior change.

One way to pick default keytab entries that is less dependent on keytab entry order is to pick those with the highest set time (fortunately keytab entries record the time at which they were created).  Here keytab entry order matters only when you have multiple entries with the same newest set-time, and that's unlikely to happen for the same principal name and kvno.

What a pain.

Issue #451, opened 2018-12-16
Request: gss_acquire_cred_from() and gss_store_cred_into()


jaltman:
[See MIT Credential Store Extensions](https://k5wiki.kerberos.org/wiki/Projects/Credential_Store_extensions#gss_acquire_cred_from.28.29)

nicowilliams:
Luke implemented this in e0bb9c1.

PR #492, opened 2019-01-01
 gssapi: credential store extensions (#451)

lhoward:
Implement `gss_acquire_cred_from()`, `gss_add_cred_from()`, and `gss_store_cred_into()`.

Import mechanism glue bugfixes from Apple's Heimdal-520.


lhoward:
@nicowilliams I've imported the Apple Heimdal-520 bugfixes (not new features) into this branch, but with each fix as a separate commit. So you may find it easiest to review the commits in this PR individually.

nicowilliams:
f33fecd looks good.

nicowilliams:
4dc4719 looks good.

lib/gssapi/mech/gss_names.c
```diff
@@ -38,6 +40,10 @@ _gss_find_mn(OM_uint32 *minor_status, struct _gss_name *name, gss_OID mech,
 
 	*output_mn = NULL;
 
+	/* null names are ok, some mechs might not have names */
```
nicowilliams: I can see that we might support `GSS_C_NO_NAME` as a `targ_name` in `gss_init_sec_context()`, but is it wise to make this choice so globally?  Might we paper over bugs this way?

lhoward:
> [f33fecd](https://github.com/heimdal/heimdal/commit/f33fecd7468a3a9cf19cf400c1416a91f8366ca6) looks good.
> [4dc4719](https://github.com/heimdal/heimdal/commit/4dc47198503ba81f03bf24b90b9ebdd423e4acc4) looks good.
> [a332161](https://github.com/heimdal/heimdal/commit/a332161f5da578dbd642328286792d1e5a3db1ea) looks... I've a question: do we want to accept `GSS_C_NO_NAME` in as many places as the change to `_gss_find_mn()` implies? I think so: the API basically accepts `GSS_C_NO_NAME` everywhere except `gss_init_sec_context()`, and this change seems to be in part about allowing it there. Not sure if I missed anything.

How about this

```udiff
+       /*
+        * XXX do we want to handle NULL names here or explicitly elsewhere?
+        * in the Apple code we return GSS_S_COMPLETE if have a NULL name.
+        */
+#if 0
+       if (name == NULL)
+               return GSS_S_COMPLETE;
+#else
+       heim_assert(name != NULL, "_gss_find_mn passed null name");
+#endif
+

**Comment by nicowilliams** [comment]
@lhoward nah, I'm convinced.

File: lib/gssapi/krb5/acquire_cred.c [hunk 2]
```diff
@@ -33,6 +33,42 @@
...
+/*
+ * Find an element in a cred store. Returns GSS_S_COMPLETE if the cred store
+ * is absent or well formed, irrespective of whether the element exists.
+ */
+OM_uint32
+__gsskrb5_cred_store_find(OM_uint32 *minor_status,
+			  gss_const_key_value_set_t cred_store,
+			  const char *key,
+			  const char **value,
+			  int internal)
```
nicowilliams: `internal` is no longer needed since you removed `__gss_cred_store_password_key`.

lib/gssapi/krb5/store_cred.c
```diff
@@ -89,38 +91,41 @@ _gsskrb5_store_cred(OM_uint32         *minor_status,
...
-            if (ret) {
-                HEIMDAL_MUTEX_unlock(&cred->cred_id_mutex);
-                *minor_status = ret;
-                return GSS_S_FAILURE;
-            }
-            overwrite_cred = 1;
-        }
+    if (cs_ccache_name)
+	ret = krb5_cc_resolve(context, cs_ccache_name, &id);
+    else {
```
nicowilliams: This feels like a bug.  What if the app provided a non-empty cred store but didn't specify a ccache name?  We shouldn't then fallback on using the default ccache!  I think providing a non-empty cred store that doesn't specify a ccache name might be a bug.  I suppose we should check what MIT does and copy that.

Also, is there a difference between specifying a ccache key with a `NULL` name value and not specifying the key at all?

lhoward:
I cherry-picked to master the things that applied cleanly, the rest I'll wait for your review to complete.

lib/gssapi/krb5/store_cred.c
lhoward: New error `GSS_KRB5_S_G_UNKNOWN_CRED_STORE_ELEMENT`.

Re: specifying a key with a NULL value and not specifying it at all, good question. I'm going to say no, as there is not in MIT:

```c
OM_uint32
kg_value_from_cred_store(gss_const_key_value_set_t cred_store,
                         const char *type, const char **value)
{
    OM_uint32 i;

    if (value == NULL)
        return GSS_S_CALL_INACCESSIBLE_WRITE;
    *value = NULL;

    if (cred_store == GSS_C_NO_CRED_STORE)
        return GSS_S_COMPLETE;

    for (i = 0; i < cred_store->count; i++) {
        if (strcmp(cred_store->elements[i].key, type) == 0) {
            if (*value != NULL)
                return GSS_S_DUPLICATE_ELEMENT;
            *value = cred_store->elements[i].value;
        }
    }

    return GSS_S_COMPLETE;
}
```

lib/gssapi/mech/gss_add_cred_from.c
```diff
@@ -0,0 +1,227 @@
...
+				      mn ? mn->gmn_name : GSS_C_NO_NAME,
+				      &m->gm_mech_oid,
+				      cred_usage,
+				      initiator_time_req,
+				      acceptor_time_req,
+				      new_mc ? &new_mc->gmc_cred : NULL,
+				      NULL,
+				      initiator_time_rec,
+				      acceptor_time_rec);
+    } else
```
nicowilliams: I guess we don't have any mechs with an `gmc_add_cred_with_password()` method, so we don't need the same else branch here that we have in `gss_acquire_cred_from()`'s helper function `acquire_mech_cred()`?

lib/gssapi/mech/gss_duplicate_cred.c
```diff
@@ -31,52 +31,84 @@
...
+	    new_mc->gmc_mech_oid = mc->gmc_mech_oid;
+	    new_mc->gmc_cred = dup_cred;
+
+	    *out = new_mc;
+	}
+
+	return major_status;
+    }
+
+    mn.gmn_mech = m;
```
nicowilliams: This is dead code.

lib/gssapi/mech/gss_duplicate_cred.c
```diff
@@ -31,52 +31,84 @@
...
-    /* This path won't work for ephemeral creds */
+	major_status = m->gm_export_cred(minor_status, mc->gmc_cred, &export);
+	if (major_status == GSS_S_COMPLETE) {
+	    major_status = m->gm_import_cred(minor_status, &export, &dup_cred);
+	    gss_release_buffer(&tmp, &export);
+	}
+    } else
+	major_status = GSS_S_UNAVAILABLE;
+
+    if (major_status != GSS_S_UNAVAILABLE) {
```
nicowilliams: I would make this read:

```
     if (major_status != GSS_S_COMPLETE)
        return major_status;

     if (dup_cred != GSS_C_NO_CREDENTIAL) {
	new_mc = calloc(1, sizeof(*new_mc));
	if (new_mc == NULL) {
            *minor_status = ENOMEM;
	    m->gm_release_cred(&tmp, &dup_cred);
	    return GSS_S_FAILURE;
	}

	new_mc->gmc_mech = m;
	new_mc->gmc_mech_oid = mc->gmc_mech_oid;
	new_mc->gmc_cred = dup_cred;
        *out = new_mc;
	return major_status;
    }

    mn.gmn_mech = m;
    mn.gmn_mech_oid = mc->gmc_mech_oid;
    mn.gmn_name = GSS_C_NO_NAME;

    ...
```

lib/gssapi/mech/gss_duplicate_cred.c
nicowilliams: See above.

lib/gssapi/mech/gss_duplicate_cred.c
```diff
@@ -31,52 +31,84 @@
...
+	}
+
+	return major_status;
+    }
+
+    mn.gmn_mech = m;
+    mn.gmn_mech_oid = mc->gmc_mech_oid;
+    mn.gmn_name = GSS_C_NO_NAME;
+
+    /* This path won't work for ephemeral creds or cred stores */
```
nicowilliams: This comment was in the original.  Why doesn't this work for ephemeral creds or cred stores?  I think it does.

nicowilliams: Ah, right, it is correct in the export/import case, but not in the duplicate case.  Just move the comment back up.

lib/gssapi/mech/gss_duplicate_cred.c
```diff
@@ -31,52 +31,84 @@
...
-    if (m->gm_duplicate_cred)
-        return m->gm_duplicate_cred(minor_status, (gss_const_cred_id_t)mc,
-                                    (gss_cred_id_t *)out);
+    if (m->gm_duplicate_cred) {
+        major_status = m->gm_duplicate_cred(minor_status,
+					    mc->gmc_cred, &dup_cred);
+    } else if (m->gm_import_cred && m->gm_export_cred) {
+	gss_buffer_desc export;
 
-    /* This path won't work for ephemeral creds */
```
nicowilliams: This comment is correct here, incorrect below.

nicowilliams: Never mind.  I misunderstood.

lib/gssapi/mech/gss_duplicate_cred.c
nicowilliams: Oy, never mind.

lib/gssapi/mech/gss_add_cred_from.c
lhoward: I don’t think that SPI exists in Heimdal? Maybe that’s a bug?

nicowilliams: Maybe.  Let's leave it for later.

nicowilliams:
Thanks!  This was massive!

PR #494, opened 2019-01-02
WIP: add new --keep* and --enctypes options to kadmin and ktutil commands

nicowilliams:
WIP -- builds but not tested

 - Add --keepold/keepallold/pruneall options to various kadmin/ktutil
   commands.  Default behavior to "prune old keys".

 - When setting keys for a service, we need to specify enctypes for it:

    - Always use kadm5_randkey_principal_3() instead of the older
      kadm5_randkey_principal().

    - Add --enctypes options to various kadmin/ktutil commands.

    - Add [libdefaults] supported_enctypes param with enctype[:salttype]
      list.

      Default to aes128-cts-hmac-sha1-96:normal.

TODO:

 - Add tests
 - Add docs


nicowilliams:
Currently leaking the parsed key/salt type tuples...

nicowilliams:
The new `--keepold` behavior is rather difficult to test.  It would help to have an `LD_PRELOAD`able interposer for `gettimeofday()`/`time()` so we could create keys in the past....

nicowilliams:
Once the current builds complete, this should be ready to review.

kadmin/ank.c
```diff
@@ -252,30 +256,46 @@ add_new_key(struct add_options *opt, int argc, char **argv)
 	return 1;
     }
 
+    enctypes = opt->enctypes_string;
+    if (enctypes == NULL || enctypes[0] == '\0')
+        enctypes = krb5_config_get_string(context, NULL, "libdefaults",
+                                          "supported_enctypes", NULL);
+    if (enctypes == NULL || enctypes[0] == '\0')
+        enctypes = "aes128-cts-hmac-sha1-96";
```
lhoward: I suppose we should have a #define given it might change in the future...

nicowilliams: Good point.

nicowilliams: Not sure where to push such a macro...  And.. oof:

```
krb5_error_code
hdb_generate_key_set(krb5_context context, krb5_principal principal,
                     krb5_key_salt_tuple *ks_tuple, int n_ks_tuple,
                     Key **ret_key_set, size_t *nkeyset, int no_salt)
{
    char **ktypes = NULL;
    char **kp;
    krb5_error_code ret;
    Key *k, *key_set;
    size_t i, j;
    char **ks_tuple_strs;
    char **config_ktypes = NULL;
    static const char *default_keytypes[] = {
        "aes256-cts-hmac-sha1-96:pw-salt",
        "des3-cbc-sha1:pw-salt",
        "arcfour-hmac-md5:pw-salt",
        NULL
    };
```
I can just not default this in `kadmind`, but the above seems like a bad default.

A proper default, on the client-side, has to come from `krb5` or `kadm5`.

I'm inclined to leave it alone...  :)

PR #500, opened 2019-01-02
kadmin: kadmin_dispatch out of memory handling

jaltman:
Coverity determined that a NULL pointer segmentation fault could occur
if krb5_storage_emem() fails.  Use krb5_enomem() to set an error message
on the context and do not perform further krb5_storage operations.

### DIFF
--- a/kadmin/server.c
+++ b/kadmin/server.c
@@ -47,7 +47,7 @@ kadmind_dispatch(void *kadm_handlep, krb5_boolean initial,
     kadm5_server_context *contextp = kadm_handlep;
     char client[128], name[128], name2[128];
     const char *op = "";
-    krb5_principal princ, princ2;
+    krb5_principal princ = NULL, princ2 = NULL;
     kadm5_principal_ent_rec ent, ent_prev;
     char *password = NULL, *expression;
     krb5_keyblock *new_keys;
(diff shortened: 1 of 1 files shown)


kadmin/server.c
```diff
@@ -125,6 +127,10 @@ kadmind_dispatch(void *kadm_handlep, krb5_boolean initial,
 	ret = kadm5_get_principal(kadm_handlep, princ, &ent, mask);
 	krb5_storage_free(sp);
 	sp = krb5_storage_emem();
+        if (sp == NULL) {
+            ret = krb5_enomem(contextp->context);
```
nicowilliams: Also `krb5_free_principal(contextp->context, princ);`.  But I've been meaning to move that to the `fail` label, and initialize `princ` to `NULL` at the top.

nicowilliams:
LGTM.  You `goto fail`  on `ENOMEM` without having freed everything in some cases, but we always `krb5_err()` anyways.  I'd like to move freeing the principal to the `fail` label, but we can do that later.

kadmin/server.c
```diff
@@ -47,7 +47,7 @@ kadmind_dispatch(void *kadm_handlep, krb5_boolean initial,
     kadm5_server_context *contextp = kadm_handlep;
     char client[128], name[128], name2[128];
     const char *op = "";
-    krb5_principal princ, princ2;
+    krb5_principal princ = NULL, princ2;
```
nicowilliams: `princ2` also needs to be initialized to `NULL` now then.

PR #502, opened 2019-01-03
Travis: cat failure logs, show core backtraces (fix #499)

### LINKED ISSUE #499: Travis build should cat failing tests' log files
Also, we should set the core dump limit to unlimited, find core dumps, and print stack traces.

nicowilliams:


nicowilliams:
I should add a commit to make a test fail so we can see if this works...

nicowilliams:
Merged as 1aedba7.

PR #507, opened 2019-01-04
Bump roken SONAME version (fix #279)

### LINKED ISSUE #279: SONAME bump for libroken
Hello,

In the fix for #107, two symbols were renamed in libroken.

However the SONAME was never incremented for the next release.

This causes problems for distributions such as Debian in getting upgrades to work smoothly.

Can you please increment the SONAME for libroken?

Thanks

nicowilliams:


nicowilliams:
libtool's [`-version-info`](https://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html) is extremely confusing.

To go from soversion major 18 to 19 I had to change the `-version-info` from `19:0:1` to `20:0:1` or `19:0:0`.

libtool subtracts the third version-info number ("age") from the first one ("current").  I don't care for the complexity of this.  I only care to bump the major number as actually built, and switching to `19:0:0` has that effect but will be confusing to future developers, so `20:0:1` it is.

PR #510, opened 2019-01-05
hdb: dereference principal aliases in all KDC lookups (#452)

lhoward:
e11abf41 added support in libhdb for always dereferencing principal aliases during an AS-REQ (where dereferencing refers to enabling alias lookups, and rewriting the returned entry with the alias name unless canonicalization was enabled).

Due to the KDC setting `HDB_F_FOR_AS_REQ` for all lookups from the AS, this allowed aliases on the TGS itself to be dereferenced during an AS-REQ; however, on presenting the TGT, the TGS would fail to resolve. Creating an explicit TGS principal for the aliased realm would work (at least prior to c555ed6a), but  this could be confusing to deploy.

This commit changes enables alias dereferencing when `HDB_F_GET_ANY` is set, which essentially means dereference whenever the request is coming from the KDC (as opposed to, say, kadmin).

We also backout c555ed6a, which changed the TGS to always canonicalize the server realm, as this breaks serving multiple realms from a single KDC, where server principals in different realms share a single canonical entry. `HDB_F_CANON `is now passed to the backend as a hint only, and per RFC 6806 the principal name is never changed in TGS replies. (However, for Samba interop, backends can override this by setting the `force-canonicalize` HDB flag.)

### DIFF
--- a/kdc/krb5tgs.c
+++ b/kdc/krb5tgs.c
@@ -1559,10 +1559,14 @@ tgs_build_reply(krb5_context context,
     s = b->sname;
     r = b->realm;
 
-    /* 
-     * Always to do CANON, see comment below about returned server principal (rsp).
+    /*
+     * The canonicalize KDC option is passed as a hint to the backend, but
+     * can typically be ignored. Per RFC 6806, names are not canonicalized
+     * in response to a TGS request (although we make an exception, see
+     * force-canonicalize below).
      */
-    flags |= HDB_F_CANON;
+    if (b->kdc_options.canonicalize)
+	flags |= HDB_F_CANON;
 
     if(b->kdc_options.enc_tkt_in_skey){
 	Ticket *t;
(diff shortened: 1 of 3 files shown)


jaltman:
I am suspicious of c555ed6a1f7ddb1cb391326140a0c30a68a9b700.   Consider the case of a service principal  `host/abc.foo.com@FOO.COM` that has an alias of `host/xyz..bar.com@BAR.COM`.   If the client requested `host/xyz.bar.com@BAR.COM` without setting the canonicalize flag won't c555ed6a1f7ddb1cb391326140a0c30a68a9b700 result in a server principal in the reply of `host/xyz.bar.com@FOO.COM`?

lhoward:
Right, I was also suspicious of it (which I think I noted at the time). Now worked around in libhdb by this commit. But also happy to back out Samba patch.

lhoward:
Backing out the Samba patch will remain (to my knowledge) compliant with [MS-KILE] but break their test suite, which always expects realm canonicalization irrespective of the canonicalize KDC option. cc @abartlet @iboukris

lhoward:
> @abartlet @iboukris Why does [c555ed6](https://github.com/heimdal/heimdal/commit/c555ed6a1f7ddb1cb391326140a0c30a68a9b700) only canonicalize the server realm and not the entire service principal?

I can answer that. Windows, since W2K3 as far as I remember, changed the TGS behavior to never canonicalize (irrespective of the setting of the canonicalize flag), but it does unconditionally canonicalize the realm to deal with NetBIOS realm aliases.

lhoward:
@abartlet @iboukris new behavior required from your backend – to support NetBIOS realm aliases in TGS responses, you should set the new HDBFlag force-canonicalize in the returned entry, which will cause the KDC to return both the canonical principal name and realm. (Your implementation should take care to only canonicalise the realm in the returned entry, if that is all that is required.)

PR #511, opened 2019-01-05
kdc: perform AS-REQ canonicalization in kdc

lhoward:
Mirroring the logic recently introduced in the TGS, this patch modifies the KDC to perform client and server canonicalization itself rather than relying on the backend to do so. Per RFC 6806, the behavior is slightly different for the AS in that the setting of the canonicalize flag in the AS-REQ _does_ impact the returned names in the ticket. In order to support realm canonicalization or other custom behavior, we allow the backend to force the KDC to canonicalize by setting the `force-canonicalize` flag in the returned client or server entries.

Rationale: name canonicalization behavior should be as consistent as possible between the AS and TGS; naming logic does not belong in the HDB. (I'd also like to move `KRB5_NT_ENTERPRISE_PRINCIPAL` handling from HDB to the KDC, but I've left that for now.)

### DIFF
--- a/kdc/kerberos5.c
+++ b/kdc/kerberos5.c
@@ -1654,7 +1654,7 @@ _kdc_as_rep(kdc_request_t r,
     krb5_error_code ret = 0;
     Key *skey;
     int found_pa = 0;
-    int i, flags = HDB_F_FOR_AS_REQ;
+    int i, cflags, sflags;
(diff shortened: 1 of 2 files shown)


lhoward:
cc @abartlet @iboukris

jaltman:
I like it.

Issue #509, opened 2019-01-05
SPNEGO does not reset NTLM RC4 state


lhoward:
MIT and Apple Heimdal's SPNEGO implementations reset the NTLM key state after processing the mechlist MIC. Apple implements this by calling `gss_inquire_sec_context_by_oid()`, MIT uses `gss_set_sec_context_option()` (which is better). Heimdal should do this to support Windows interop.

[MS-SPNG]

> 3.2.5.1 NTLM RC4 Key State for MechListMIC and First Signed Message
When NTLM is negotiated, the SPNG server MUST set OriginalHandle to ServerHandle before generating the mechListMIC, then set ServerHandle to OriginalHandle after generating the mechListMIC. This results in the RC4 key state being the same for the mechListMIC and for the first message signed by the application.
Because the RC4 key state is the same for the mechListMIC and for the first message signed by the application, the SPNEGO Extension server MUST set OriginalHandle to ClientHandle before validating the mechListMIC and then set ClientHandle to OriginalHandle after validating the mechListMIC.

lhoward:
cc @abartlet

lhoward:
MIT implementation:

```
/*
 * MS-SPNG section 3.3.5.1 warns that the NTLM mechanism requires special
 * handling of the crypto state to interop with Windows.  If the mechanism for
 * sc is SPNEGO, invoke a mechanism-specific operation on the context to reset
 * the RC4 state after producing or verifying a MIC.  Ignore a result of
 * GSS_S_UNAVAILABLE for compatibility with older versions of the mechanism
 * that do not support this functionality.
 */
static OM_uint32
ntlmssp_reset_crypto_state(OM_uint32 *minor_status, spnego_gss_ctx_id_t sc,
                           OM_uint32 verify)
{
        OM_uint32 major, minor;
        gss_buffer_desc value;

        if (!g_OID_equal(sc->internal_mech, &gss_mech_ntlmssp_oid))
                return GSS_S_COMPLETE;

        value.length = sizeof(verify);
        value.value = &verify;
        major = gss_set_sec_context_option(&minor, &sc->ctx_handle,
                                           (gss_OID)&ntlmssp_reset_crypto_oid,
                                           &value);
        if (major == GSS_S_UNAVAILABLE)
                return GSS_S_COMPLETE;
        *minor_status = minor;
        return major;
}
```

lhoward:
Obviously needs to be handled in NTLM too.

lhoward:
[...]

PR #512, opened 2019-01-05
gssapi: SPNEGO does not reset NTLM RC4 state (#509)

lhoward:
Untested

### DIFF
--- a/lib/gssapi/Makefile.am
+++ b/lib/gssapi/Makefile.am
@@ -202,6 +202,7 @@ ntlmsrc = \
 	ntlm/process_context_token.c \
 	ntlm/release_cred.c \
 	ntlm/release_name.c \
+	ntlm/set_sec_context_option.c \
 	ntlm/kdc.c
 
 $(srcdir)/ntlm/ntlm-private.h: $(ntlmsrc)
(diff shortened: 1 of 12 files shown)


lhoward:
cc @simo5

lhoward:
Really we need to import Apple's NTLM implementation (or maybe Samba has one?)

nicowilliams:
@lhoward do we also need to restore sequence numbers?

lhoward:
@nicowilliams that is done by `_gss_ntlm_set_key()`

nicowilliams:
Ah, I'd missed that.

simo5:
@lhoward I think you should just support/use https://github.com/simo5/gss-ntlmssp

nicowilliams:
Could this have been implemented by having SPNEGO export the NTLM sec context prior to making the MIC, then re-import it at the end?

simo5:
> Could this have been implemented by having SPNEGO export the NTLM sec
context prior to making the MIC, then re-import it at the end?

I do not recall all details, but I remember considering that option
when I implemented gss-ntlmssp and the required fixes to MIT SPNEGO
implementation, and didn't go that way. So there may well be some
detail that won't work with a full export/import.

nicowilliams:
@simo5 that sounds familiar.  ISTR talking about this before.  I also don't recall the details, so I'm going to assume that this is the right fix.

lhoward:
I think in terms of coding style and licensing, ingesting the Apple changes is probably the best way forward.

simo5:
I was more concerned about correctness.
I have basically only one isue left with gss-ntlmssp, but I corrected ton of stuff that is broken in many implementations, including Heimdal one.
But up to you anyway

PR #513, opened 2019-01-07
KDC salt fixes

lhoward:
* generate default salts for entries missing them
* omit default salt from etype info

### DIFF
--- a/lib/hdb/common.c
+++ b/lib/hdb/common.c
@@ -31,6 +31,7 @@
  * SUCH DAMAGE.
  */
 
+#include "krb5_locl.h"
 #include "hdb_locl.h"
 
 int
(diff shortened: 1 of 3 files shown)


nicowilliams:
LGTM.

nicowilliams:
LGTM.

lib/hdb/common.c
```diff
@@ -98,6 +99,49 @@ hdb_value2entry_alias(krb5_context context, krb5_data *value,
...
+    krb5_salt pwsalt;
+
+    ret = krb5_get_pw_salt(context, entry->principal, &pwsalt);
+    if (ret)
+	return ret;
+
+    for (i = 0; i < entry->keys.len; i++) {
+	Key *key = &entry->keys.val[i];
+
+	if (key->salt != NULL ||
```
jaltman: It might be worth noting that if an explicit key was set instead of deriving via `string_to_key()`, then `key->salt == NULL`.  However, the default salt would still be the same.  Its just that the salt would be ignored by the client when a keytab is used.  Generating the default salt and adding it to the key doesn't alter any behavior.

PR #515, opened 2019-01-07
kdc: get_pa_etype_info_both comply with RFC4120

jaltman:
Heimdal's current behavior regarding the generation of PA-ETYPE-INFO2
and PA-ETYPE-INFO violates RFC4120 in two ways:

 1. when generating responding both PA-ETYPE-INFO2 and PA-ETYPE-INFO
    the hints returned in the inverse order: INFO then INFO2 instead
    of INFO2 then INFO.

 2. the determination that both PA-ETYPE-INFO2 and PA-ETYPE-INFO is
    currently based upon the KDC selected enctype when it should be
    determine based upon examining the entire enctype list specified
    by the requesting client.

This change corrects the behavior to follow the RFC4120 guidance.

### DIFF
--- a/kdc/kerberos5.c
+++ b/kdc/kerberos5.c
@@ -1221,28 +1221,55 @@ get_pa_etype_info2(krb5_context context,
     return 0;
 }
 
+static int
+newer_enctype_present(struct KDC_REQ_BODY_etype *etype_list)
+{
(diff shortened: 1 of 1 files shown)


nicowilliams:
LGTM.

PR #516, opened 2019-01-07
ASN.1: Support wider bit sets (fix #514)

### LINKED ISSUE #514: Heimdal ASN.1 compiler does not support BIT STRINGs with more than 32 members
The Heimdal ASN.1 compiler generates code using `unsigned` and `unsigned int` for dealing with `BIT STRING` types with named bit members.  This limits it to 32 members on most platforms.

This means, for example, that we cannot add new `HDBFlags` members without either re-purposing an old member or fixing the ASN.1 compiler to support `uint64_t` for bit sets.

Naturally, to go beyond 64 elements we'll need to redesign how the ASN.1 compiler handles bit sets.  Rather than encoding/decoding to integral types, it should encode/decode to fixed-sized `char` or `uint64_t` arrays.

nicowilliams:
Wait for builds and tests before reviewing.  Also: need to add a test.

The ASN.1 compiler now uses `uint64_t` for `BIT STRING` bitsets with between 33 and 64 members.

For `BIT STRING` bitsets with more than 64 members, no `<type>2int()` nor `int2<type>()` functions are generated, and the application can then only encode/decode.

### DIFF
--- a/lib/asn1/gen.c
+++ b/lib/asn1/gen.c
@@ -763,6 +763,27 @@ define_type (int level, const char *name, const char *basename, Type *t, int typ
 	Member *m;
 	Type i;
 	struct range range = { 0, UINT_MAX };
+        size_t max_memno = 0;
+        size_t bitset_size;
+
+        /*
+         * range.max implies the size of the base unsigned integer used for the
+         * bitfield members.  If it's less than or equal to UINT_MAX, then that
+         * will be unsigned int, otherwise it will be uint64_t.
+         *
+         * We could just use uint64_t, yes, but for now, and in case that any
+         * projects were exposing the BIT STRING types' C representations in
+         * ABIs prior to this compiler supporting BIT STRING with larger
+         * members, we stick to this.
+         */
+        ASN1_TAILQ_FOREACH(m, t->members, members) {
+            if (m->val > max_memno)
+                max_memno = m->val;
+        }
+        if (max_memno > 63)
+            range.max = INT64_MAX;
+        else
+            range.max = 1LU << max_memno;
 
 	i.type = TInteger;
 	i.range = &range;
@@ -780,10 +801,13 @@ define_type (int level, const char *name, const char *basename, Type *t, int typ
 	    ASN1_TAILQ_FOREACH(m, t->members, members) {
 		char *n = NULL;
 
-		/* pad unused */
+		/*
+                 * pad unused bits beween declared members (hopefully this
+                 * forces the compiler to give us an obvious layout)
+                 */
 		while (pos < m->val) {
 		    if (asprintf (&n, "_unused%d:1", pos) < 0 || n == NULL)
-			errx(1, "malloc");
+			err(1, "malloc");
 		    define_type (level + 1, n, newbasename, &i, FALSE, FALSE);
 		    free(n);
 		    pos++;
@@ -797,8 +821,13 @@ define_type (int level, const char *name, const char *basename, Type *t, int typ
 		n = NULL;
 		pos++;
 	    }
-	    /* pad to 32 elements */
-	    while (pos < 32) {
+	    /* pad unused tail (ditto) */
+            bitset_size = max_memno;
+            if (max_memno > 31)
+                bitset_size += 64 - (max_memno % 64);
+            else
+                bitset_size = 32;
+	    while (pos < bitset_size) {
 		char *n = NULL;
 		if (asprintf (&n, "_unused%d:1", pos) < 0 || n == NULL)
 		    errx(1, "malloc");
--- a/lib/asn1/gen.c
+++ b/lib/asn1/gen.c
@@ -288,15 +288,16 @@ close_generate (void)
 {
     fprintf (headerfile, "#endif /* __%s_h__ */\n", headerbase);
 
-    if (headerfile)
-        fclose (headerfile);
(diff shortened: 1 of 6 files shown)


lib/asn1/gen_glue.c
```diff
@@ -66,11 +66,11 @@ generate_int2 (const Type *t, const char *gen_name)
     Member *m;
 
     fprintf (headerfile,
-	     "%s int2%s(unsigned);\n",
+	     "%s int2%s(unsigned long);\n",
```
nicowilliams: Should be `unsigned long long`.

Alternatively we should use `uint64_t` and make sure we `#include <stdint.h>` anywhere we might be missing it.

lib/asn1/gen_glue.c
```diff
@@ -114,7 +114,7 @@ generate_units (const Type *t, const char *gen_name)
     if(t->members) {
 	ASN1_TAILQ_FOREACH_REVERSE(m, t->members, memhead, members) {
 	    fprintf (codefile,
-		     "\t{\"%s\",\t1U << %d},\n", m->name, m->val);
+		     "\t{\"%s\",\t1LU << %d},\n", m->name, m->val);
```
nicowilliams: Should be `1LLU`.

lib/asn1/gen_glue.c
```diff
@@ -53,7 +53,7 @@ generate_2int (const Type *t, const char *gen_name)
 	     gen_name, gen_name);
 
     ASN1_TAILQ_FOREACH(m, t->members, members) {
-	fprintf (codefile, "if(f.%s) r |= (1U << %d);\n",
+	fprintf (codefile, "if(f.%s) r |= (1LU << %d);\n",
```
nicowilliams: Should be `1LLU`.

lib/asn1/gen_glue.c
```diff
@@ -66,11 +66,11 @@ generate_int2 (const Type *t, const char *gen_name)
     Member *m;
 
     fprintf (headerfile,
-	     "%s int2%s(unsigned);\n",
+	     "%s int2%s(unsigned long);\n",
 	     gen_name, gen_name);
 
     fprintf (codefile,
-	     "%s int2%s(unsigned n)\n"
+	     "%s int2%s(unsigned long n)\n"
```
nicowilliams: Ditto.

lib/asn1/gen_glue.c
nicowilliams: Can't be `1LLU`: VC in Appveyor failed to support that.

nicowilliams:
Note that now `BIT STRING`s with members larger than 63 are supported, but the compiler does not generate `<type>2int()` and `int2<type>()` functions in that case (i.e., the app must always encode/decode to/from octet string).

lib/roken/parse_units.h
```diff
@@ -51,7 +51,7 @@
 
 struct units {
     const char *name;
-    unsigned mult;
+    unsigned long long mult;
```
vdukhovni: Given this, looks like units have a 64-bit or so ceiling?

lib/asn1/gen_glue.c
```diff
@@ -139,18 +139,26 @@ generate_units (const Type *t, const char *gen_name)
...
+        if (ASN1_TAILQ_EMPTY(t->members))
+            break;
+        ASN1_TAILQ_FOREACH(m, t->members, members) {
+            if (m->val > 63)
+                return;
+        }
+        generate_2int (t, gen_name);
+        generate_int2 (t, gen_name);
+        if (parse_units_flag)
+            generate_units (t, gen_name);
```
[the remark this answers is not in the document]
nicowilliams: No.  We need new parse_units support for these that's not based on integers.

lib/asn1/gen.c
```diff
@@ -797,8 +807,12 @@ define_type (int level, const char *name, const char *basename, Type *t, int typ
 		n = NULL;
 		pos++;
 	    }
-	    /* pad to 32 elements */
-	    while (pos < 32) {
+	    /* pad unused tail */
+            if (max_memno > 31)
+                max_memno += 64 - (max_memno % 64);
+            else
+                max_memno = 32;
```
vdukhovni: A brief comment explain why this is exactly right and not off by one could be cool.

nicowilliams:
FYI, the `lib/roken/parse_units.c` utilities are kinda broken.

First, they only support parsing / unparsing (formatting) `int` values, which means that we're limited to 31 flags, not 32, wherever we use this functionality.  Changing this to `int64_t` would increase the limit to 63, naturally.  We could also make `parse_flags()` and `unparse_flags()` deal in `uint64_t`, but that means splitting `{un,}parse_something()` into `{un,}parse_{signed,unsigned}()`.

Second, there is no way for these functions to signal errors (they return `-1`, but since they could legitimately return `-1`, there is no way to distinguish error cases from legitimate `-1` cases).  So, not only should we break the ABI for these functions as described above, but we should also break it by changing their signatures to allow for error reporting.

Third, for flags we ought to be able to deal with the bitfield structs that the ASN.1 compiler generates for `BIT STRING` types with named members.  In this case these functions should take a pointer and size, and read / write those bit fields.  This would allow code using `BIT STRING` types with named members larger than 63 to also use the `lib/roken` units code for user interfaces.

PR #517, opened 2019-01-07
WIP Explicit kvno change option (fix #480)

### LINKED ISSUE #480: Add kadmin cpw/randkey option to leave kvno unchanged
When rotating keys for services shared by multiple hosts it is necessary to first update all the keytabs, *then* change the keys on the KDC:

1. generate new keys;
2. update all the keytabs;
3. set those new keys for the principal in the HDB.

Heimdal's `kadm5` API does support writing new keysets to a principal _without_ changing its kvno, which then makes an alternative cluster key rotation procedure possible:

1. add new keys to the principal in the HDB _without_ changing the principal's current kvno;
2. update all the keytabs;
3. update the principal's kvno in the HDB.

However, we don't have support for this in the kadmin commands.  We need a `--keepkvno` (please suggest better names) option to the `cpw` and `randkey` commands.  The `modify` command already supports setting the kvno of a principal.

nicowilliams:
WIP.  Wait for builds and tests before reviewing.  Also: need to add a test.

### DIFF
--- a/admin/get.c
+++ b/admin/get.c
@@ -116,6 +116,130 @@ parse_enctypes(struct get_options *opt,
     return ret;
 }
 
+static krb5_error_code
+kt_max_kvno(krb5_keytab keytab,
+            krb5_const_principal princ,
+            krb5_kvno *max_kvno)
+{
+    krb5_error_code ret;
+    krb5_kt_cursor cursor;
+    krb5_keytab_entry entry;
+
+    *max_kvno = 0;
+    ret = krb5_kt_start_seq_get(context, keytab, &cursor);
+    if (ret)
+        return ret;
+    while ((ret = krb5_kt_next_entry(context, keytab, &entry, &cursor)) == 0) {
+        if (entry.vno > *max_kvno)
+            *max_kvno = entry.vno;
+    }
+    if (ret && ret != KRB5_KT_END)
+        (void) krb5_kt_end_seq_get(context, keytab, &cursor);
+    else
+        ret = krb5_kt_end_seq_get(context, keytab, &cursor);
+    return ret;
+
+}
+
+static krb5_error_code
+kt_write_keys(krb5_keytab keytab,
+              const char *pname,
+              kadm5_principal_ent_rec *princ,
+              krb5_kvno pre_randkey_kvno,
+              int n_randkeys,
+              krb5_keyblock *randkeys)
+{
+    krb5_keytab_entry entry;
+    krb5_error_code saved_ret = 0;
+    krb5_error_code ret;
+    krb5_kvno max_kt_kvno;
+    size_t i;
+
+    ret = kt_max_kvno(keytab, princ->principal, &max_kt_kvno);
+    if (ret)
+        return ret;
+
+    /*
+     * kadm5_randkey_principal_3() does not output a kvno for the new keyset.
+     *
+     * If the user has permission to get the principal's keys, then we just
+     * write those (we know their kvnos).  Otherwise we can only write the keys
+     * returned by kadm5_randkey_principal_3().
+     *
+     * In the latter case we can guess the new kvno because we do a get, a
+     * randkey, then a get, and if the kvno went up by the expected number,
+     * we're good, otherwise (e.g., we raced with another randkey) we are
+     * unsure.
+     *
+     * If KRB5_KDB_EXPLICIT_KVNO_CHANGE is set then the randkey should have
+     * failed if there were keys not yet brought into use, so we won't get here
+     * in that case.  Defendsively, we check anyways.
+     */
+
+    if (kadm5_all_keys_are_bogus(princ->n_key_data, princ->key_data)) {
+        /*
+         * We did a get, a randkey, then another get.  The used doesn't have
+         * permission to get the principal's keys, but does have permission to
+         * randkey the principal, and just did.  We have only the new keys, but
+         * don't know their kvno.
+         *
+         * Determine if the princ->kvno is meaningful here.
+         */
+        if (((princ->attributes & KRB5_KDB_EXPLICIT_KVNO_CHANGE) &&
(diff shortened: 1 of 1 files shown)


lib/hdb/print.c
```diff
@@ -341,6 +341,11 @@ entry2string_int (krb5_context context, krb5_storage *sp, hdb_entry *ent)
 #define KRB5_KDB_PWCHANGE_SERVICE       0x00002000
 #define KRB5_KDB_SUPPORT_DESMD5         0x00004000
 #define KRB5_KDB_NEW_PRINC              0x00008000
+#define KRB5_KDB_OK_AS_DELEGATE         0x00010000
```
nicowilliams: Put these in a separate commit.

lib/hdb/print.c
```diff
@@ -367,6 +372,14 @@ flags_to_attr(HDBFlags flags)
         a |= KRB5_KDB_DISALLOW_SVR;
     if (flags.change_pw)
         a |= KRB5_KDB_PWCHANGE_SERVICE;
+    if (flags.ok_as_delegate)
```
nicowilliams: Ditto.

lib/hdb/hdb.asn1
```diff
@@ -52,7 +52,9 @@ HDBFlags ::= BIT STRING {
 	force-canonicalize(30),		-- force the KDC to return the canonical
 					-- principal irrespective of the setting
 					-- of the canonicalize KDC option
-	do-not-store(31)		-- Not to be modified and stored in HDB
+	do-not-store(31),		-- Not to be modified and stored in HDB
+	explicit-kvno-changes-only(32)	-- New keys only become effective on
```
nicowilliams: We have a number of unused members between 19 and 29, inclusive.  Maybe use one of those?

nicowilliams:
We need a test that creates a principal, `randkey`s it, sets this flag, `randkey`s the principal again, checks the kvno hasn't changed, then `ext_keytab` to make sure we do get the new keyset, change the kvno explicitly, and check that the KDC sees the new keys.

lib/hdb/hdb.asn1
nicowilliams: And if we stick to 32, make sure load/dump preserves this new flag and that the LDAP HDB backend uses 64-bit integers for the principal attributes field or else does not support this flag.  (Also, do LDAP servers support 64-bit integers?)

hyc: Integers in LDAP are unbounded precision. Although it was common on 32bit servers to only implement them as 32bits, the ASN.1 specification does not limit their size at all. And 64bits is certainly safe on 64bit servers.

nicowilliams: Thanks @hyc.  I knew that LDAP supports large integers, but I didn't know if *servers* (or even client libraries) would.  In any case, not getting support for this flag in the event of truncation is not fatal, and the user can check for this.

nicowilliams: Argh, I forgot to check what the kadmin protocol does in Heimdal:

```
        krb5_store_int32(sp, princ->attributes);
```

nicowilliams: Well, I'll use `(20)`, as that's available and I'm already using `1<<20` for this elsewhere in this PR.

kadmin/util.c
```diff
@@ -65,6 +65,7 @@ struct units kdb_attrs[] = {
     { "disallow-tgt-based",	KRB5_KDB_DISALLOW_TGT_BASED },
     { "disallow-forwardable",	KRB5_KDB_DISALLOW_FORWARDABLE },
     { "disallow-postdated",	KRB5_KDB_DISALLOW_POSTDATED },
+    { "explicit-kvno-changes-only",	KRB5_KDB_EXPLICIT_KVNO_CHANGE },
```
nicowilliams: Order is important here.  Highest-numbered flags must come first.

lib/hdb/hdb.asn1
lhoward: You could use the higher bits for force-canonicalise and do-not-store, as they should never be stored.

nicowilliams:
So, `kadmin ext_keytab` does the right thing, but `ktutil get` does not yet.

One thing that's annoying is that `kadm5_randkey_principal*()` do not return a kvno for the new keys.  Applications do a `kadm5_get_principal()` after a randkey operation to get the principal's kvno.  But this is racy!  Suppose two clients race to randkey the same principal... one client can write the wrong kvno for the new keys in the keytab.

The only thing that works is to get the keys with `kadm5_get_principal(KAFM5_KEY_DATA)`.  But that presumes `get-keys` privilege.  Perhaps we should allow the client to get keys for decrypting potentially-extant tickets whenever the client can do a randkey operation...  Perhaps splitting `get-keys` from `get` was a mistake after all.  Perhaps what we needed all along for forward security was the new key pruning features.

nicowilliams:
`ktutil`'s `get`, `add`, and `change` are particularly difficult to test, as they prompt for a password.  For manual testing I link `ktutil` with `libkadm5srv` instead of `libkadm5clnt`.  It might be nice if `ktutil` could behave the same as `kadmin`, having a `-l` option.  I'm not sure if anyone is packaging `ktutil` such that if you install it you only get `libkadm5clnt`, but if you install `kadmin` then you also get `libkadm5srv`, though in Heimdal 8.0 we can surely break that.

nicowilliams:
We could also remove `ktutil`'s `get`, `add`, and `change` commands...  after all, `kadmin` has equivalent functionality.

nicowilliams:
Ah, MIT ran into this a few years back: http://krbdev.mit.edu/rt/Ticket/Display.html?id=7852

nicowilliams:
Given that MIT opted for fetching keys to work around these problems, I'll do the same.  I'll make a `kadm5_write_keytab()` function that writes a principal's keys to a keytab.

PR #523, opened 2019-01-15
Make lib/roken/parse_units.c more robust; fix #522

### LINKED ISSUE #522: `lib/roken/parse_units.c` utilities are kinda broken
The `lib/roken/parse_units.c` utilities are kinda broken.  These are utilities for CLIs and interactive UIs to parse and format/print integral values of various units (e.g., bytes, time), or flags bitset.

First, these utilities only support parsing / unparsing (formatting) `int` values, which means that we're limited to 31 flags, not 32, wherever we use this functionality. Changing this to `int64_t` would increase the limit to 63, naturally.  We could also make `{un,}parse_flags()` deal in `uint64_t`, but that means splitting `{un,}parse_something()` into `{un,}parse_{signed,unsigned}()`.  There is also the danger of signed integer overflow and aggressive optimizers applying undefined behaviors to this code or its callers.

Second, there is no way for these functions to signal errors (they return `-1`, but since they could legitimately return `-1`, there is no way to distinguish error cases from legitimate `-1` cases). So, not only should we break the ABI for these functions to support wider numbers, but we should also break it by changing their signatures to allow for error reporting.  Or at the very least we should change them to use `errno` to signal errors.

Third, for flags we ought to be able to deal with the bitfield structs that the ASN.1 compiler generates for `BIT STRING` types with named members. In this case these functions should take a pointer and size, and read / write those bit fields. This would allow code using `BIT STRING` types with named members larger than 63 to also use the `lib/roken` units code for user interfaces.

nicowilliams:


lib/roken/parse_bytes.c
```diff
@@ -33,10 +33,27 @@
...
+ * we should still be able to have declare entry for terabytes here and just
+ * not use it.
+ *
+ * The simplest way to do this is to skip as many entries as needed to reach a
+ * strictly-decreasing set of entries using casts to size_t below to get
+ * unsigned overflow semantics.  This allows us to have entries for
+ * terabyte which, if sizeof(ssize_t) == 4, get skipped over.
+ */
 static struct units bytes_units[] = {
+    { "terabyte", (size_t)1024 * 1024 * 1024 * 1024 },
```
nicowilliams: Actually, the second field is a `uint64_t` now, so there's no need to do this cast or check for overflow.

jaltman: I would prefer this as `0x10000000000LL` instead of the multiplication.

PR #527, opened 2019-02-21
Make common plugin api public

jaltman:
This series fixes build issues on Windows related to the new common plugin api and makes the common plugin interface public for consumption by third party plugins.

### DIFF
--- a/lib/krb5/plugin.c
+++ b/lib/krb5/plugin.c
@@ -151,7 +151,7 @@ copy_internal_dso(const char *name)
 }
 
 struct krb5_plugin {
-    krb5_plugin_common_ftable *ftable;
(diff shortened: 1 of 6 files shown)


lib/krb5/plugin.c
```diff
@@ -651,11 +646,6 @@ search_modules(heim_object_t key, heim_object_t value, void *ctx)
 					  s->caller,
 					  path,
 					  p->dsohandle);
-	if (plugins == NULL) {
-	    /* fallback to loading structure directly */
-	    plugins = add_dso_plugin_struct(s->context, path,
```
lhoward: is there a reason you moved this, is it because you don't regard the two mechanisms as peers, rather a fallback?

PR #537, opened 2019-03-21
Storage stdio

elric1:
Couple of optimisations of stdio krb5_storage and a couple of tests and increasing the consistency of truncate between backends a bit.


nicowilliams:
Some test fails: https://travis-ci.org/heimdal/heimdal/jobs/509217429

PR #546, opened 2019-04-06
GSS-API const-correctness

nicowilliams:
We still had a bunch of `const gss_...` arguments in function prototypes in `<gssapi/gssapi.h>`.  This fixes all of them.


lhoward:
This will likely be a source-breaking change for plugins that import Heimdal headers and implement the GSS-API, such as Moonshot.

nicowilliams:
I'm ok with that.

nicowilliams:
ping @jaltman

lhoward:
We could insulate by including a private copy of the header in mech_eap but, need to be careful as there are some platform-specific things like structure packing on macOS. I’d prefer not to do this but then mech_eap will he tied to a specific Heimdal version. (I don’t have any skin in the game, just prefer to move slow and not break things.)

nicowilliams:
Or we could have a macro that makes all the `gss_const_*` types not const.

lhoward:
> Or we could have a macro that makes all the gss_const_* types not const.
> 

I think I already did that in the current code. It’s dealing with mixed versions of Heimdal that have different degrees of const-ness which is tricky.

Cc @alejandro-perez

nicowilliams:
I mean, PuTTY and the OpenJDK include their own private copies of `<gssapi.h>`.

lhoward:
Yeah it’s probably the best option.

nicowilliams:
I think so.  OK, so now, to review this.  Of course, it's very mechanical, rote, large, boring.

alejandro-perez:
> I think I already did that in the current code. It’s dealing with mixed versions of Heimdal that have different degrees of const-ness which is tricky. Cc @alejandro-perez

This was done for the Mac client, wasn't it? In that case, I haven't had the time to check it out yet.

lhoward:
@nicowilliams wants to change it yet again to add more const-ness, which is correct but would be another source breaking change for building mech_eap against Heimdal (unless of course we synchronise and require the latest version).

alejandro-perez:
In principle, I'd be ok with that. But I'd check with @spaetow, who is the one managing mech_eap's builds for Mac at the moment.

lhoward:
It's not Mac-specific though, it's any platform that uses Heimdal (does Debian use it?). I think the right answer is to actually have a private gssapi.h in mech_eap. But this needs to be done carefully to be ABI compatible for both MIT and Heimdal.

alejandro-perez:
At the moment, we are using MIT for Debian, yet I guess we could potentially use Heimdal as well.
We might have a private gssapi.h in mech_eap that is only used when Heimdal is selected, whereas the system's default is used if MIT is chosen.

nicowilliams:
I'd also be OK with adding a macro you can define to get non-constified GSS funciton prototypes.

lhoward:
We already have one in mech_eap, problem is that to deal with different Heimdal versions we'd need to have two macros. Then it gets ugly quickly.

lhoward:
Actually that's not correct, we don't have a macro. Currently it is like:

```c
OM_uint32 GSSAPI_CALLCONV
gss_accept_sec_context(OM_uint32 *minor,
                       gss_ctx_id_t *context_handle,
#ifdef HAVE_HEIMDAL_VERSION
                       gss_const_cred_id_t cred,
#else
                       gss_cred_id_t cred,
#endif
                       gss_buffer_t input_token,
                       gss_channel_bindings_t input_chan_bindings,
                       gss_name_t *src_name,
                       gss_OID *mech_type,
                       gss_buffer_t output_token,
                       OM_uint32 *ret_flags,
                       OM_uint32 *time_rec,
                       gss_cred_id_t *delegated_cred_handle);
```

lhoward:
but yes, probably want to avoid something like `GSS_MAYBE_CONST`...

nicowilliams:
I was thinking of a macro that changes the `gss_const_*` typedefs to not actually be `const`...

Or, perhaps, add new typedefs like `gss_maybe_const_*`, use those in all the function prototypes where we now use `gss_const_*`, and then have a macro select whether `gss_maybe_const_X` is a typedef alias of `gss_X` or `gss_const_X`.

This way the header can be fairly clean.  I don't really want `#ifdef`s in the middle of function prototypes...

lhoward:
Above wasn't a prototype BTW (I just added the semicolon for correctness), it's actually the function body in mech_eap. Sorry if that was confusing!

lhoward:
Anyway, do what you need to do and I'll update mech_eap. Sounds like no one seems to be caring if it builds against Heimdal except maybe the SSP and macOS versions.

nicowilliams:
Hey, there was a semicolon...

lhoward:
My bad, should've been `{}`!

alejandro-perez:
> Anyway, do what you need to do and I'll update mech_eap. Sounds like no one seems to be caring if it builds against Heimdal except maybe the SSP and macOS versions.

Please, note the new upstream git repo for mech_eap is now https://github.com/janetuk/mech_eap

nicowilliams:
Ok, thanks.  Do you want to review this?  I can probably get @vdukhovni, but it's really fairly obvious, and produced fairly mechanically (I used lots of VIM macros to make it easy to do this quickly).

lhoward:
Thanks @alejandro-perez, I think I actually have a few fixes sitting in the old repository that aren't in this. I can look into this next week when I'm back in Australia. Also, is the SSP being maintained? Can we synchronise the two?

alejandro-perez:
I think the SSP is Painless work, I have no clue about its status.

nicowilliams:
@lhoward no, no, by all means, it may be OT but it's interesting.

lhoward:
Well, the SSP is a fork of mech_eap with a small number of changes to make GSS objects reference counted and add a small number of extra fields (e.g. for the NT token). Those changes should probably be upstreamed. The SSP itself is effectively a not-perfectly-abstracted shim between SSPI and GSS.

nicowilliams:
Reference-counting sounds useful, and getting a reference (incref) should be distinct from duplicating the object (as in `gss_duplicate_{name,cred}()`) in that an object to which you have a reference might be mutated by another thread (e.g., call `gss_add_cred()` to add cred elements) while a duplicate should not be.  And also, we'd not want to have `gss_duplicate_context()` really duplicate, but rather, get a reference (incref).

nicowilliams:
Also, I like the idea of having an SSP->GSS bridge shim in-tree.  I'd also like us to have a clean version of Martin's `gsskrb5` GSS->SSP shim.

lhoward:
Only the context and credential objects are reference counted and it's more for managing concurrency and shared credentials than anything else (it isn't exposed to the application directly). The code also predates `gss_duplicate_cred()`.

I wasn't planning to open-source the shim, although it could be done (I think it would need Painless' approval though, I would have to check). I'd also need to consider the business risk of doing so (PADL is the code owner).

nicowilliams:
Never mind -- no pressure!

lib/gssapi/mech/gss_mech_switch.c
```diff
@@ -162,11 +162,13 @@ _gss_string_to_oid(const char* s, gss_OID *oidp)
 }
 #endif
 
+typedef void (*f_ptr)();
```
jaltman: I feel that the introduction and use of f_ptr should be a separate change with a commit message explaining what warning messages it is suppressing.

lib/gssapi/krb5/gsskrb5_locl.h
```diff
@@ -83,13 +83,15 @@ typedef struct gsskrb5_ctx {
   struct krb5_ticket *ticket;
   time_t endtime;
   HEIMDAL_MUTEX ctx_id_mutex;
+  HEIMDAL_MUTEX *ctx_id_mutexp;
```
jaltman: Does this change belong in this commit?

lib/gssapi/mech/gss_mech_switch.c
nicowilliams: Hmm, maybe I can rework this one and avoid the typedef.

lib/gssapi/krb5/gsskrb5_locl.h
nicowilliams: Oh, I took constification too far.  Security contexts should never be `const`.  I'll fix it.  Thanks for catching this.

nicowilliams:
I braino'ed when I saw `gss_const_ctx_id_t`.  It should not exist, but I ran with it.  Oops.

nicowilliams:
@jaltman shall I push fixup commits or squashed and rebased commits?

nicowilliams:
All four commits can now be squashed down to one.

nicowilliams:
I still have to undo the mutex pointer thing.

nicowilliams:
OK, I've undone the mutex thing, and now the diffs are more manageable:  +456 −458.

PR #551, opened 2019-05-04
support for GSS_C_ANON_FLAG in krb5 mech

lhoward:
Creating pull request to track progress on support for anonymous PKINIT in GSS-API


kdc/krb5tgs.c
```diff
@@ -873,10 +899,15 @@ tgs_make_reply(krb5_context context,
...
     et.flags.ok_as_delegate = server->entry.flags.ok_as_delegate;
 
-    if(rspac->length) {
+    /*
+     * For anonymous tickets, we should filter out positive authorization data
+     * that could reveal the client's identity, and return a policy error for
+     * restrictive authorization data. Policy for unknown authorization types
+     * is implementation dependent.
+     */
+    if(rspac->length && !et.flags.anonymous) {
```
jaltman: space between if and ( please

lhoward:
Depends on #552

jaltman:
Please rebase on top of current master

lhoward:
> Please rebase now that the fix commits have been merged independent of this PR.

Rebased and pushed. It's a while ago since I looked at this, so I can't remember what you want to do about the RFC8062 divergences (noted in the code). What do we need to do to get this merged?

Also, since I wrote that code, @vdukhovni fixed `krb5_principal_is_anonymous()` to deal with the fact that Heimdal 7.5 and below KDCs leave the anonymous principal name type as `KRB5_NT_PRINCIPAL`. Should we support these with krb5 GSS anonymous support? If so, we need to carefully audit `krb5_principal_is_anonymous()` calls from the mechanism to see which ones need to also pass `KRB5_ANON_IGNORE_NAME_TYPE`.

cc @jaltman

lhoward:
OK, I added a separate commit with `KRB5_ANON_IGNORE_NAME_TYPE` but it needs careful review.

PR #552, opened 2019-05-08
krb5 anonymous fixes (ex-GSS)

lhoward:
Anonymous fixes up to but not including GSS_C_ANON_FLAG support

### DIFF
--- a/kdc/krb5tgs.c
+++ b/kdc/krb5tgs.c
@@ -373,7 +373,10 @@ check_PAC(krb5_context context,
 static krb5_error_code
 check_tgs_flags(krb5_context context,
 		krb5_kdc_configuration *config,
-		KDC_REQ_BODY *b, const EncTicketPart *tgt, EncTicketPart *et)
+		KDC_REQ_BODY *b,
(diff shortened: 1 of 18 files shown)


kdc/krb5tgs.c
```diff
@@ -814,12 +835,17 @@ tgs_make_reply(krb5_context context,
...
 
-    ret = copy_PrincipalName(&tgt_name->name, &rep.cname);
+    /*
+     * RFC 8062 states "if the ticket in the TGS request is an anonymous
+     * one, the client and client realm are copied from that ticket". So
+     * whilst the TGT flag check below is superfluous, it is included in
+     * order to follow the specification to its letter.
+     */
+    if (et.flags.anonymous && !tgt->flags.anonymous)
+	_kdc_make_anonymous_principalname(&rep.cname);
```
vdukhovni: The comment should instead explain the replacement of the client name with the anonymous name when an authenticated client wants an anonymous ticket.  The normal case barely needs commenting. (Not a release blocker).

PR #553, opened 2019-05-10
Sendto kdc

elric1:
Comment in the commit. It is a little naughty of me to re-use the fd to keep state in this way but it does avoid another variable in wait_ctx. I'm no wedded to it, though, we can decide to be cleaner.

### DIFF
--- a/lib/krb5/send_to_kdc.c
+++ b/lib/krb5/send_to_kdc.c
@@ -1188,7 +1188,7 @@ krb5_sendto_context(krb5_context context,
 
 	    action = KRB5_SENDTO_CONTINUE;
 	    if (ret == 0) {
-		_krb5_debug(context, 5, "submissing new requests to new host");
+		_krb5_debug(context, 5, "submitting new requests to new host");
 		if (submit_request(context, ctx, hi) != 0)
 		    action = KRB5_SENDTO_TIMEOUT;
 	    } else {
--- a/lib/krb5/send_to_kdc.c
+++ b/lib/krb5/send_to_kdc.c
@@ -329,6 +329,7 @@ static void
 debug_host(krb5_context context, int level, struct host *host, const char *fmt, ...)
 {
     const char *proto = "unknown";
+    const char *state;
     char name[NI_MAXHOST], port[NI_MAXSERV];
     char *text = NULL;
     va_list ap;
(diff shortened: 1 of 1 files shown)


jaltman:
I'm not sure this approach is correct.  The goal is to avoid delays and where possible additional round trips.  In most cases the Kerberos library cannot maintain state and cache which endpoints for a given server are reachable and which path is fastest.   Therefore, the only approach that can avoid unnecessary delays is to send to all endpoints for a given KDC and protocol simultaneously and use the first positive response that is returned.   The KDCs implement replay detection so if a request is received from more than one endpoint it will either (a) ignore the duplicate request, (b) return the prior response, or (c) return a replay error. 

Even one second is a very long delay that I believe should be avoided.

[RFC6555 Happy Eyeballs Dual Stack](https://tools.ietf.org/html/rfc6555) for Connected protocols and [RFC6157 IPv6 Transition in SIP](https://tools.ietf.org/html/rfc6157) as an example for unconnected protocols are relevant.

elric1:
> I'm not sure this approach is correct. The goal is to avoid delays and where possible additional round trips. In most cases the Kerberos library cannot maintain state and cache which endpoints for a given server are reachable and which path is fastest. Therefore, the only approach that can avoid unnecessary delays is to send to all endpoints for a given KDC and protocol simultaneously and use the first positive response that is returned. The KDCs implement replay detection so if a request is received from more than one endpoint it will either (a) ignore the duplicate request, (b) return the prior response, or (c) return a replay error.
> 
> Even one second is a very long delay that I believe should be avoided.
> 
> [RFC6555 Happy Eyeballs Dual Stack](https://tools.ietf.org/html/rfc6555) for Connected protocols and [RFC6157 IPv6 Transition in SIP](https://tools.ietf.org/html/rfc6157) as an example for unconnected protocols are relevant.

Well, this patch is closer to what you want than what we have now. My goal was to ensure that there is never a time where we are waiting whilst not actually waiting. And so, I believe that my patch is fixing the current algorithm. What you are proposing is changing it.

I'm not against reducing the delay from 1s which is quite long, but I'm not convinced that replay detection allows you to issue the requests in parallel. It's not necessarily true that all of the IP addresses returned for a multi-homed host are actually going to go to the same KDC and be rejected as replays. In fact, I don't think that we even share the replay cache between different KDC processes on a single KDC with multiple cores.

(That said, your goals can be attained by changing the 3 to a 0 on one of the lines of the patch.)

PR #558, opened 2019-05-15
make krb5_principal_is_anonymous public

jaltman:
_krb5_principal_is_anonymous() is used outside lib/krb5 and
therefore it needs to be properly exported and its flag macros
need to be in a public header: krb5.h not krb5_locl.h.

Including krb5_locl.h from within kuser_locl.h for instance
results in build failures on Solaris.

This change renames the function and makes it part of the public
api.

Change-Id: I130d1698b10bdbd150b95e8c7d32dfc362889ce6

### DIFF
--- a/kdc/kerberos5.c
+++ b/kdc/kerberos5.c
@@ -1613,7 +1613,7 @@ generate_pac(kdc_request_t r, Key *skey)
 krb5_boolean
 _kdc_is_anonymous(krb5_context context, krb5_const_principal principal)
 {
-    return _krb5_principal_is_anonymous(context, principal, KRB5_ANON_MATCH_ANY);
+    return krb5_principal_is_anonymous(context, principal, KRB5_ANON_MATCH_ANY);
 }
 
 static int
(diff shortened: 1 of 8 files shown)


lhoward:
I would prefer to keep it private API if possible just because I didn’t give much thought to its design, perhaps we need a way to expose semi-private API outside krb5_locl.h?

jaltman:
If the function is to be used outside of the kerberos library even by an in-tree process it must be properly exported by which I mean correct use of KRB5_LIB_FUNCTION and KRB5_LIB_CALL.    

The prototypes for all functions are automatically generated into krb5-protos.h which is included by krb5.h. 

krb5_locl.h cannot be included by source files outside of lib/krb5.  

We can leave the leading underscore as part of the name but we don't have an alternative to including the macros in krb5.h.

Another possibility is to publicly provide three functions

  krb5_principal_is_anonymous()
  krb5_principal_is_anonymous_authenticated()
  krb5_principal_is_anonymous_unauthenticated()

which removes the need for the macros entirely.

lhoward:
OK, LGTM, let's merge.

PR #566, opened 2019-05-16
Fix sendto_kdc on Windows

jaltman:
The combination of 8740528b2477f872147998ca19a66d2fae12631b
("Windows-compatible sentinel socket type and value") and
d497d7e4a796c6bda153e7f10fe9478724b855ab ("krb5_sendto_kdc:
failover for multiple AAAA/A RRs on one domain") broke
all the send to kdc loop on Windows.  rk_socket_t is
a HANDLE and rk_INVALID_SOCKET is the max value.  Therefore,
no valid socket will be larger and all communications
will fail.

Change-Id: I3464f78d67b19f14050ad7a01738fb32bac99385


jaltman:
I've verified this fixes the bug.

PR #567, opened 2019-05-16
hcrypto: avoid hcrypto_validate race

jaltman:
Do not increment 'validated' until after all of the validation
steps have been performed.  While its ok the validate more than
once we need to ensure that validation occurs at least once
before the cipher is used.

Change-Id: I6896c132ec2a7423c5166c1c074a2dedac54e00d


jaltman:
this results in infinite recursion.

PR #570, opened 2019-05-18
hcrypto-pkcs11: check mechanism flags

lhoward:
Before committing to a PKCS#11 mechanism, check that it can provide the required encryption or digest services by validating the flags returned by C_GetMechanismInfo().

### DIFF
--- a/lib/hcrypto/evp-pkcs11.c
+++ b/lib/hcrypto/evp-pkcs11.c
@@ -158,7 +158,9 @@ p11_module_init(void)
 }
 
 static CK_RV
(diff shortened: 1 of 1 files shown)


quanah:
Testing with this patch now.

quanah:
With the hack for the validated function in place, plus this patch, the entire test suite passes.  I'm now rebuilding without the hack to the validated function to see if it resolves that issue as well.

quanah:
This patch fixes all issues I encountered with Solaris 11. :)

PR #572, opened 2019-05-19
krb5: check KDC supports anonymous if requested

jaltman:
Verify the KDC recognized the request-anonymous flag by validating the returned
client principal name.

(cherry picked from commit 014e318d6bdefd8ecfcb99ca9928921f6a49d721)


lhoward:
Needs patch for krb5_principal_is_anonymous() without leading underscore

PR #577, opened 2019-05-21
kuser: allow kinit to renew anonymous PKINIT tickets

lhoward:
Anonymous PKINIT tickets discard the realm information used to locate the issuing AS. Store the issuing realm in the credentials cache in order to locate a KDC which can renew them.

### DIFF
--- a/kuser/kinit.c
+++ b/kuser/kinit.c
@@ -206,6 +206,9 @@ static struct getargs args[] = {
     { "help",		0,   arg_flag, &help_flag, NULL, NULL }
 };
 
+static char *
+get_default_realm(krb5_context context);
+
 static void
 usage(int ret)
 {
@@ -236,7 +239,7 @@ copy_configs(krb5_context context,
 	     krb5_principal start_ticket_server)
 {
     krb5_error_code ret;
-    const char *cfg_names[] = {"realm-config", "FriendlyName", NULL};
+    const char *cfg_names[] = {"realm-config", "FriendlyName", "anon-pkinit-realm", NULL};
     const char *cfg_names_w_pname[] = {"fast_avail", NULL};
     krb5_data cfg_data;
     size_t i;
@@ -277,6 +280,30 @@ copy_configs(krb5_context context,
 }
 
 static krb5_error_code
+get_anon_pkinit_tgs_name(krb5_context context,
+			 krb5_ccache ccache,
+			 krb5_principal *tgs_name)
+{
+    krb5_error_code ret;
+    krb5_data data;
+    char *realm;
+
+    ret = krb5_cc_get_config(context, ccache, NULL, "anon-pkinit-realm", &data);
+    if (ret == 0) {
+	realm = malloc(data.length + 1);
+	memcpy(realm, data.data, data.length);
+	realm[data.length] = '\0';
+    } else
+	realm = get_default_realm(context);
+
+    ret = krb5_make_principal(context, tgs_name, realm,
+			      KRB5_TGS_NAME, realm, NULL);
+    free(realm);
+
+    return ret;
+}
+
+static krb5_error_code
 renew_validate(krb5_context context,
 	       int renew,
 	       int validate,
@@ -296,7 +323,13 @@ renew_validate(krb5_context context,
 	krb5_warn(context, ret, "krb5_cc_get_principal");
 	return ret;
     }
-    ret = get_server(context, in.client, server, &in.server);
+
+    if (server == NULL &&
+	krb5_principal_is_anonymous(context, in.client,
+				    KRB5_ANON_MATCH_UNAUTHENTICATED))
+	ret = get_anon_pkinit_tgs_name(context, cache, &in.server);
+    else
+	ret = get_server(context, in.client, server, &in.server);
     if (ret) {
 	krb5_warn(context, ret, "get_server");
 	goto out;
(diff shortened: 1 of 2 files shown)


lhoward:
cc'ing @greghudson as this introduces a new ccache config key we may wish to synchronise across implementations

lhoward:
Added test.

greghudson:
Some notes:

* Nico has advocated for a "start realm" config entry which can be used not just for anonymous PKINIT, but for a kind of delegation where the client principal realm doesn't match the TGT realm.  See https://k5wiki.kerberos.org/wiki/Projects/StartRealmCCconfig for a short writeup.  I thought this had been implemented in Heimdal (it hasn't for MIT krb5 yet) but maybe not.  The start realm seems like it would apply to this purpose, but the name anon-pkinit-realm is scoped too closely to PKINIT.

* The start realm (or this equivalent) is also needed when doing cross-realm with anonymous tickets.  See http://krbdev.mit.edu/rt/Ticket/Display.html?id=8332

* We may be developing an underscores-vs-dashes inconsistency.  All of the MIT config names use underscores (fast_avail, proxy_impersonator, etc.) but it looks like Heimdal may already have an established name using dashes (realm-config) as well as one using camelcase (FriendlyName).

nicowilliams-work:
@greghudson `start_realm` does seem relevant here, indeed.  I don't care about ccconfig naming conventions -- if we can agree on one, we can be consistent about it, and if not, not.  Consistency is far far behind interoperability as a concern :)

quanah:
I care about consistency of naming conventions, myself.  There was already a bug where an MIT style krb5.conf file would utterly break Heimdal, and was shipped as a default config with RHEL7 and caused a bit of havoc.

lhoward:
Happy to change to start_realm

jaltman:
I'm fine with start_realm as well.

lhoward:
In that case we should probably honor it unconditionally (unless explicit server principal specified) rather than just for anon PKINIT principals.

Will fix today.

jaltman:
I will submit a PR for the change

jaltman:
> In that case we should probably honor it unconditionally (unless explicit server principal specified) rather than just for anon PKINIT principals.

Agreed.

lhoward:
Just tracking down a test failure in the gss-anon branch, but will address this very shortly.

lhoward:
See PR#579

jaltman:
@greghudson, @lhoward @nicowilliams and I discussed the use cases for `start_realm` which is primarily focused on which realm to issue referrals queries to.   This is distinct from the use case that `anon-pkinit-realm` is intended to solve which is preserving the realm from which the initial ticket was acquired so that the same realm can be contacted for ticket renewals.

We are going to continue to use `anon-pkinit-realm` for the new use case.   Both because it is simpler to analyze the distinct use cases but also because `start_realm` would point to the wrong realm if the unauthenticated initial ticket was for a cross-realm TGT.   In that case, `start_realm` needs to refer to the second component of the service principal name and not the service principal realm.

lhoward:
We will change the config key from `anon-pkinit-realm` to `anon_pkinit_realm` for MIT style compatibility.

jaltman:
We will change from `anon-pkinit-realm` to `anon_pkinit_realm` to address the concern regarding hyphens vs underscores.

greghudson:
> discussed the use cases for start_realm which is primarily focused on which realm to issue referrals queries to

I don't think it's just about referrals.  If I am contacting a service principal with a known realm, say `host/hostname@SREALM`, but my TGT is for a separate realm CREALM, I need a hint to start with a krbtgt/CREALM ticket to get a krbtgt/SREALM ticket, if my client principal isn't in CREALM (because it's anonymous or because of the unusual delegation scenario).

> but also because start_realm would point to the wrong realm if the unauthenticated initial ticket was for a cross-realm TGT. In that case, start_realm needs to refer to the second component of the service principal name and not the service principal realm.

If I understand this scenario right, a client makes an anonymous PKINIT AS request to REALM1, asking for krbtgt/REALM2, and gets a ticket for WELLKNOWN/ANONYMOUS@WELLKNOWN:ANONYMOUS -> krbtgt/REALM2@REALM1.

I assume your reasoning is that to renew this ticket, REALM1 must be contacted.  But to obtain any service tickets (either via referrals or explicit cross-realm), REALM2 must be contacted.  Therefore, the start_realm (REALM2) must be different from the renewal realm (REALM1).

I will note that looking up krbtgt/start_realm (krbtgt/REALM2 in this example) with TC_MATCH_SRV_NAMEONLY would obtain the ticket that needs to be renewed, and the realm of that ticket would indicate which realm to contact.  If you merely know REALM1, you still need to be told what server principal to renew, or you might figure that out by also looking at start_realm.

I guess for the current API (krb5_get_renewed_creds()), because the caller supplies a string-form service principal without realm, REALM1 is the additional piece of information you need.  That might be a flaw in the API; in principle one might want to be able to renew a cross-realm service ticket, and the  API has no way to express that.

I'm developing a sense that what we really want is a way to know the starting, non-ephemeral service principal of a credential cache, whether or not it's an initial ticket and whether it's a local TGT, cross TGT, or service ticket.  If we knew that, we'd know exactly what ticket to renew (no reason for the caller to even say), and that would also be the starting TGT for referrals or explicit cross-realm.

greghudson:
Very-late-breaking realization: isn't anon-pkinit-realm just the realm of the ticket server principal?  (So creds->server.realm within krb5_creds.)

Recall that fully anonymous PKINIT discards the realm of the ticket _client_, but not the ticket server.

PR #579, opened 2019-05-22
kuser: store anon PKINIT realm in "start_realm"

lhoward:
Store the anonymous PKINIT starting realm in the "start_realm" credentials cache configuration variable.

Always honor this when no server principal is passed to kinit, irrespective of whether the client is the unauthenticated PKINIT one. This allows for future uses.

### DIFF
--- a/kuser/kinit.c
+++ b/kuser/kinit.c
@@ -218,18 +218,60 @@ usage(int ret)
 }
 
 static krb5_error_code
+get_starting_tgs_name(krb5_context context,
+		      krb5_ccache ccache,
+		      krb5_principal *tgs_name)
+{
+    krb5_error_code ret;
+    krb5_data data;
+    char *realm;
+
+    ret = krb5_cc_get_config(context, ccache, NULL, "start_realm", &data);
+    if (ret)
+	return ret;
+
+    realm = malloc(data.length + 1);
+    memcpy(realm, data.data, data.length);
+    realm[data.length] = '\0';
+
+    ret = krb5_make_principal(context, tgs_name, realm,
+			      KRB5_TGS_NAME, realm, NULL);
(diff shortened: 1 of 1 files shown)


kuser/kinit.c
```diff
@@ -217,19 +217,61 @@ usage(int ret)
...
+{
+    krb5_error_code ret;
+    krb5_data data;
+    char *realm;
+
+    ret = krb5_cc_get_config(context, ccache, NULL, "start_realm", &data);
+    if (ret)
+	return ret;
+
+    realm = malloc(data.length + 1);
```
jaltman: `  if (realm == NULL) {
      ret = krb5_enomem(context); 
      goto out;
  }`

kuser/kinit.c
```diff
@@ -217,19 +217,61 @@ usage(int ret)
...
+    memcpy(realm, data.data, data.length);
+    realm[data.length] = '\0';
+
+    ret = krb5_make_principal(context, tgs_name, realm,
+			      KRB5_TGS_NAME, realm, NULL);
```
jaltman: `out:`

kuser/kinit.c
```diff
@@ -217,19 +217,61 @@ usage(int ret)
...
+    free(realm);
+
```
jaltman: Don't we need to free the krb5_data allocated by krb5_cc_get_config() -> krb5_copy_data()?

`free(data);`

jaltman:
Perhaps the commit message should reference https://k5wiki.kerberos.org/wiki/Projects/StartRealmCCconfig

jaltman:
See #577 for discussion of why we couldn't overload the meaning of `start_realm`

PR #583, opened 2019-05-22
hx509: replace time_t with int64_t for cert timestamps

jaltman:
On platforms with 32-bit time_t (e.g. Linux i386), certificates
with timestamps later than 03:14:07 UTC on 19 January 2038 fail
to be processed correctly.

Recent changes to include certificates in the test suite with
500 year lifetimes cause the test suite to fail on these platforms.

This change replaces all use of time_t with int64_t to permit
uniform processing of certificate timestamps on all platforms.

Change-Id: I8ada6392478f39862c62d5b6490682b026e49261

### DIFF
--- a/lib/hx509/print.c
+++ b/lib/hx509/print.c
@@ -64,12 +64,12 @@ struct cert_status {
 static int
 Time2string(const Time *T, char **str)
 {
-    time_t t;
+    int64_t t;
     char *s;
     struct tm *tm;
 
     *str = NULL;
-    t = _hx509_Time2time_t(T);
+    t = _hx509_Time2int64_t(T);
     tm = gmtime (&t);
     s = malloc(30);
     if (s == NULL)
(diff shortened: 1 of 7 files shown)


lib/hx509/print.c
```diff
@@ -64,12 +64,12 @@ struct cert_status {
...
 {
-    time_t t;
+    int64_t t;
     char *s;
     struct tm *tm;
 
     *str = NULL;
-    t = _hx509_Time2time_t(T);
+    t = _hx509_Time2int64_t(T);
     tm = gmtime (&t);
```
jaltman: This won't work because we can't pass a `int64_t *` to a function that accepts a 32-bit `time_t *`.

jaltman:
The reason the checks pass is that all of the Travis and Appveyor builders are platforms with 64-bit time_t.

jaltman:
What we need is https://sourceware.org/glibc/wiki/Y2038ProofnessDesign?rev=115

Issue #533, opened 2019-03-07
hx509/test_chain fails, something might have expired


SethEllsworth:
This was just noticed today. A build of our product that includes heimdal started failing make check.

At test_chain test in heimdal/lib/hx509.

Saw that test_chain was a script, ran the failing command, and got the error:
$ ./hxtool --statistic-file=../../lib/hx509/statfile verify cert:FILE:./data/test.crt anchor:FILE:./data/ca.crt crl:FILE:./data/crl1.der
verify_path: No revoke status found for certificates: 569956
no certs verify at all
$

Nothing code wise had changed recently, so after some testing, figured out it was time. 

$ sudo date -s "-2days"
Mon Mar  4 16:55:07 MST 2019
$ ./hxtool --statistic-file=../../lib/hx509/statfile verify cert:FILE:./data/test.crt anchor:FILE:./data/ca.crt crl:FILE:./data/crl1.der
path ok
$

Since we have some custom changes to the code base, I have validated with a fresh checkout of origin/heimdal-7-1-branch, and master.

Looks like the file was last updated roughly 10 years ago.

quanah:
In the OpenLDAP project, I made 500 year certs, may be worthwhile to do something similar here.

Looks like the majority of certs expire next month:
            Not After : Apr 24 20:29:40 2019 GMT

SethEllsworth:
FYI this still fails, test_ca+, on x86. Re-running the regen script with a 10 year limit works, so guessing a 32-bit 2038 issue.

jaltman:
What is the size of `time_t` on the 32-bit platform the test is failing on?   If its 32-bit, then that is the source of the problem.

jaltman:
@quanah Perhaps generate a cert with an expiration date of `Sun, 07 Feb 2106 06:28:14 GMT`

SethEllsworth:
It is 32 bits in our cross compile setup:
$ cat l.c
#include <time.h>
#include <stdio.h>

int main( void ) {
    fprintf( stdout, "time_t size: <%d>\n", sizeof(time_t) );
}

$ gcc -m32 -o l l.c
$ ./l
time_t size: <4>
$

kirotawa:
hi there!

any reasons for this issue be still open?

quanah:
The new certs break 32-bit OSes, so either the code needs to be updated so 32-bit builds can handle the newer certs or we need to commit certs that expire in 2038 or so.  Thus the issue remains open at this time.

[...]

PR #584, opened 2019-05-23
For issue #533

quanah:
Regenerate certs so that they expire before the 2038 armageddon so the test suite will pass on 32-bit operating systems until the underlying issues can be resolved.


jaltman:
@brianmay This commit includes the pre-Y2038 certificates for the 7.7 release.

PR #587, opened 2019-05-24
.travis.yml: add gcc 32-bit build on linux

jaltman:
Change-Id: I6dbb72d9ed5e45ae862a2e6742a2e6a5cee4a6de


jaltman:
This is going to be tricky since we have to figure out how to install 32-bit dev packages and pass the 32-bit flags to the libedit configure script.

PR #594, opened 2019-06-01
Do not set anonymous flag in S4U2Proxy request

iboukris:
CC @lhoward 

Constrained delegation fails on master, after commit: 014e318d
I get BADOPTION at: https://github.com/heimdal/heimdal/blob/014e318d6bdefd8ecfcb99ca9928921f6a49d721/lib/krb5/ticket.c#L727
Any idea why is S4U2Proxy automatically set to anonymous? I didn't see a mention of it elsewhere.
If I remove the anonymous flag as in this patch then it works okay.

To reproduce:
$ echo pwd | kinit -f impersonator@REALM
$ kgetcred --out-cache=evidence --impersonate=user@REALM impersonator@REALM
$ kgetcred --debug --delegation-credential-cache=evidence --out-cache=out -H HTTP/apache.realm@REALM
$ KRB5CCNAME=out curl --negotiate -u: http://apache.realm/whoami.aspx

### DIFF
--- a/lib/krb5/get_cred.c
+++ b/lib/krb5/get_cred.c
@@ -1644,10 +1644,8 @@ next_rule:
 	flags.b.forwardable = 1;
     if (options & KRB5_GC_NO_TRANSIT_CHECK)
 	flags.b.disable_transited_check = 1;
-    if (options & KRB5_GC_CONSTRAINED_DELEGATION) {
-	flags.b.request_anonymous = 1; /* XXX ARGH confusion */
+    if (options & KRB5_GC_CONSTRAINED_DELEGATION)
 	flags.b.constrained_delegation = 1;
-    }
     if (options & KRB5_GC_ANONYMOUS)
 	flags.b.request_anonymous = 1;
 


jaltman:
`request_anonymous = 1` was added by d5bb7a7c566841d52662b230248f06522bfa64ad.

It was removed by Apple in their Heimdal 398.1.2 code drop.

jaltman:
@iboukris please update the commit message and I will get this merged and cherry-picked for the 7.7 release.

iboukris:
@jaltman - done.

jaltman:
Since Heimdal clients have been requesting anonymous along with constrained delegation for 13 years, I wonder if we need to modify the KDC to ignore the anonymous flag if constrained delegation is requested.  Is there any use case for that combination?

iboukris:
I now see this too:
https://github.com/heimdal/heimdal/blob/a7bb4504f2f3dfb276f8aa154858f46ed1063011/kdc/kerberos5.c#L2437

iboukris:
I'm not clear about the meaning of the anonymous flag in constrained delegation TGS request, I wonder if we can make a better guess when we process a ticket with both S4U2Proxy and anonymous flags, whether or not the client really wanted anonymity or it was an old client sending both.

iboukris:
I read some of 4.2. section in rfc 6112, and it made me think that a S4U2Proxy in the current model cannot hide the impersonator principal as it is the same as 2nd ticket server. It might make sense to use an anonymous ticket (received via regular anonymous tgs request, or some kind of S4U2self) as a second ticket but that's pure imagination,  and I didn't find any hints on how the two work, or do not work together.

iboukris:
> After all, the reason this has suddenly become an issue is because the KDC finally supports authenticated anonymous requests.

Actually, I'm testing against Windows KDC (updated 2019) which seem to ignore the anonymous flag (at least for this request) as it replies with the cname of the impersonated client and not anonymous client, and it is the client that was unhappy with this returned ticket (since it requested anon).

jaltman:
@iboukris Thanks for the research and the additional information regarding your testing strategy.  Please take a look at https://github.com/heimdal/heimdal/pull/595

iboukris:
Note: the reason we had `request_anonymous = 1` was when Heimdal had inversed flags definition in asn1, and request_anonymous was actually constrained_delegation, so it had to be set in order for the client doing constrained-delegation to work against Windows KDC.
The wrong asn1 definition of the flags was fixed in: 86554f5a7f81da1efa2849fa6961ca71ad3b8e90
So now this line can be safely removed (it should have been removed with asn1 definition commit).

PR #595, opened 2019-06-01
kdc: history of request_anonymous vs cname-in-addl-tkt confusion

jaltman:
Drafts 0 through 10 of the Kerberos anonymity draft specified the
TicketFlags.anonymous flag as bit 14.  This was changed to bit 16
after it was discovered that Microsoft used bit 14 for S4U2Proxy.

d5bb7a7c566841d52662b230248f06522bfa64ad ("(krb5_get_creds): if
KRB5_GC_CONSTRAINED_DELEGATION is set, set both") set both the
anonymous and constrained_delegation TicketFlags when issuing a
S4U2Proxy request.  The setting of the anonymous TicketFlag was
removed by ea7615ade3af28843f358e715703226b760db73b("Do not set
anonymous flag in S4U2Proxy request").

014e318d6bdefd8ecfcb99ca9928921f6a49d721 ("krb5: check KDC
supports anonymous if requested") introduced a client side check
to ensure that an anonymous request is responded to with an
anonymized ticket.  The combination of setting the anonymous
TicketFlag and the anonymized ticket validation broke S4U2Proxy
requests to Windows KDCs because they ignore the anonymous TicketFlag
when constrained_delegation is requested.

The Heimdal KDC includes fallback logic to handle Heimdal clients
that set the anonymous TicketFlag as bit 14 in _kdc_is_anon_request().
However, it failed to adjust the kdc_options flags when it
determined that the request came from an old Heimdal client.

This change clears the constrained_delegation flag and sets the
request_anonymous flag when an old Heimdal client is detected.
It also clears the request_anonymous flag if both bit 14 and 16
are set.

Change-Id: If57b6f9fe95fdba0109c4450dba5548b4ae6eba9

### DIFF
--- a/kdc/kerberos5.c
+++ b/kdc/kerberos5.c
@@ -123,11 +123,11 @@ is_anon_as_request_p(kdc_request_t r)
     KDC_REQ_BODY *b = &r->req.req_body;
 
     /*
-     * Some versions of heimdal use bit 14 instead of 16 for
-     * request_anonymous, as indicated in the anonymous draft prior to
-     * version 11. Bit 14 is assigned to S4U2Proxy, but S4U2Proxy requests
-     * are only sent to the TGS and, in any case, would have an additional
-     * ticket present.
+     * Versions of Heimdal from 0.9rc1 through 1.50 use bit 14 instead
+     * of 16 for request_anonymous, as indicated in the anonymous draft
+     * prior to version 11. Bit 14 is assigned to S4U2Proxy, but S4U2Proxy
+     * requests are only sent to the TGS and, in any case, would have an
+     * additional ticket present.
      */
     return b->kdc_options.request_anonymous ||
 	   (b->kdc_options.cname_in_addl_tkt && !b->additional_tickets);
--- a/kdc/krb5tgs.c
+++ b/kdc/krb5tgs.c
@@ -373,7 +373,7 @@ is_anon_tgs_request_p(const KDC_REQ_BODY *b,
     KDCOptions f = b->kdc_options;
 
     /*
-     * Earlier (pre-7.6) versions of Heimdal would send both the
+     * Versions of Heimdal from 1.0 to 7.6, inclusive, send both the
      * request-anonymous and cname-in-addl-tkt flags for constrained
      * delegation requests. A true anonymous TGS request will only
      * have the request-anonymous flag set. (A corollary of this is


jaltman:
its interesting that nothing in the kdc code looks at `kdc_options.constrained_delegation`

iboukris:
> its interesting that nothing in the kdc code looks at `kdc_options.constrained_delegation`

That's really strange, MS-SFU says the client must set it, but I couldn't any mention for the kdc to check it.  I think we should add a check, MIT does check, per code:
https://github.com/krb5/krb5/blob/08e948cce2c79a3604066fcf7a64fc527456f83d/src/kdc/do_tgs_req.c#L302

I'll run some tests against Windows and Samba without the flag set, to see if it is checked somewhere.

jaltman:
@lhoward  are you referring to the current pull request or the addition of a check of the `kdc_options.constrained_delegation` flag?   

Regarding this commit, there shouldn't be a significant merge conflict.  #551 moves and renames the _kdc_is_anon_request() function to a static is_anon_request() function and then adds additional functionality.   We could pull out the renaming and changing to a static into its own commit if you wish.

jaltman:
I would like to add a change to rename kdc_options.constrained_delegation to kdc_options.cname-in-addl-tkt so the name matches the MS-SFU documentation.

lhoward:
Sounds good to me.

kdc/kerberos5.c
```diff
@@ -2432,12 +2432,29 @@ _kdc_tkt_add_if_relevant_ad(krb5_context context,
...
+	    b->kdc_options.request_anonymous = 1;
+	}
+	else if (b->kdc_options.request_anonymous) {
+	    /*
+	     * Heimdal versions between 1.0 and 7.6 set the anonymous_request
+	     * bit (16) in addition to constrained_delegation (14) when
+	     * requesting S4U2Proxy.  The Windows KDC ignores the
+	     * anonymous_request bit when satisfying a S4U2Proxy request.
+	     */
+	    b->kdc_options.request_anonymous = 0;
```
iboukris: So here we think it is a simple S4U2Proxy request wrongly flagged anonymous, are we sure _kdc_is_anon_request() is called from TGS? some of the calls seem to be from AS request (also, does it make sense to check if the tgt is anon?).
I think I would prefer to see this fixup in more specific location in the flow.

jaltman: This change only impacts AS requests and not TGS request.   

kdc/krb5tgs.c check_tgs_flags() might be a good location to perform these checks.  Its called at the start of tgs_make_reply().

jaltman: I will look at this again on Monday

lhoward: Is there an approach that doesn't involve rewriting the body flags? This feels wrong to me.

lhoward: BTW I committed the cname_in_addl_tkt change to master.

lhoward:
cdd0b70d should fix this. If you also want to support clients that use bit 14 for anonymous TGS requests (no such clients exist to my knowledge) then an additional check could be made in `is_anon_tgs_request_p()`.

jaltman:
@lhoward's changes pushed to master look good to me although I believe the in-tree comments could be improved and I think it is worthwhile having the history of these issues documented somewhere in the repository.   A commit message is good enough.  This pull request modifies some of the in-tree comments and provides the following commit message text

kdc: history of request_anonymous vs cname-in-addl-tkt confusion

Drafts 0 through 10 of the Kerberos anonymity internet draft,
https://tools.ietf.org/html/draft-ietf-krb-wg-anon, specified the
TicketFlags.anonymous flag as bit 14 and the KDCOptions.anonymous
flag as bit 14.  These were changed to bit 16 by MIT after it was
discovered that Microsoft used KDCOptions bit 14 for S4U2Proxy
cname-in-addl-tkt.

(Feb 2007) Heimdal added constrained delegation support prior to
1.0 but named the KDCOptions flag constrained_delegation instead of
cname-in-addl-tkt as per MS-SFU.  It also assigned bit 16 instead
of bit 14.  Perhaps this was done in the hope that the conflict
with Microsoft would be resolved in favor of the IETF internet
draft instead of the proprietary protocol extension.

adf912182266321b754ed1cb5a705ba2103e139a ("Add PA-ClientCanonicalized
and friends.") introduced the KDCOptions.constrained_delegation flag
as bit 16.

(June 2007) In order to make Heimdal's constrained delegation work
with Microsoft's implementation Heimdal began to set both KDCOptions
bits 14 and 16 when requesting constrained delegation.

d5bb7a7c566841d52662b230248f06522bfa64ad ("(krb5_get_creds): if
KRB5_GC_CONSTRAINED_DELEGATION is set, set both") set both the
anonymous and constrained_delegation TicketFlags when issuing a
S4U2Proxy request.

(June 2010) MIT reassigned the KDCOption.anonymous and
TicketFlags.anonymous flags to bit 16.  draft-ietf-krb-anon-11
was published with this change.

(July 2014) After the release of Heimdal 1.5.0 and prior to 1.5.1
it was noticed that Heimdal's anonymous TGT support did not
interoperate with MIT.

86554f5a7f81da1efa2849fa6961ca71ad3b8e90 ("Use correct value for
anonymous flags") swapped the bit assignments for request_anonymous
and constrained_delegation but failed to remove the setting of
KDCOptions bit 16 ("anonymous") when requesting constrained
delegation.

(May 2019) Prior to the 7.6 release many corrections to Heimdal's
anonymity support were introduced to bring it into compliance
with RFC8062.  This included support for requesting anonymous
tickets via the TGS service.  Because not all KDC can satisfy
anonymous requests the client must verify if the response was
anonymized.  This check wasn't added until after 7.6 was
released.

014e318d6bdefd8ecfcb99ca9928921f6a49d721 ("krb5: check KDC
supports anonymous if requested").

The combination of setting KDCOption.anonymous when requesting
constrained delegation and the anonymized ticket validation
broke S4U2Proxy requests to Windows KDCs.  Windows KDCs ignore
the KDCOption.anonymous flag when processing a TGS request
with KDCOption.cname-in-addl-tkt set.

ea7615ade3af28843f358e715703226b760db73b ("Do not set
anonymous flag in S4U2Proxy request") removed the behavior
of setting the KDCOption.anonymous flag that should have
been removed in July 2014.

(June 2019) The Heimdal KDC includes fallback logic to handle
Heimdal clients from 1.0 to 1.5.0, inclusive, that set the
KDCOptions.anonymous flag as bit 14.  Prior to the 7.7 release
this logic only handled AS request but failed to handle the
constrained delegation request case where both bits 14 and 16
were set in the TGS request.

cdd0b70d37d87026e8618ff44b8d636c0bf9cb6c ("kdc: don't misidentify
constrained delegation requests as anonymous") added the TGS
request validation to distinguish anonymous requests from
constrained delegation requests.

This change documents the history in the commit message and
updates some in-tree comments.

Change-Id: I625cd012e2e6c263c71948c6021cc2fad4d2e53a

kdc/kerberos5.c
jaltman: Thanks.

iboukris:
Looks good to me, and surely is useful.

lhoward:
LGTM. Should we ditch the additional tickets check in `is_anon_as_request_p()`?

PR #598, opened 2019-06-09
Validate KDC support of S4U extensions

iboukris:
Per MS-SFU (see 1.5,  3.1.5.1.2 and 3.1.5.2.4, for both S4U2Self and S4U2proxy):
Services can detect whether the KDC supports S4U by checking the cname
of the returned ticket. KDCs that do not support S4U ignore the
S4U2self and S4U2proxy data and return a service ticket with the cname
containing the name of the service that made the request ([RFC4120]
section 3.3.3). In service tickets from KDCs that support S4U, the
cname contains the name of the user.


iboukris:
I tested it manually against Windows and it looks okay, I'll try to add automated tests.

```
echo PASSWD | kinit -f HTTP/srv-kerberos-kcd.alpha.local@ALPHA.LOCAL
/usr/heimdal/bin/kgetcred --out-cache=evidence --impersonate=HTTP/srv-kerberos-kcd.alpha.local@ALPHA.LOCAL -H HTTP/srv-kerberos-kcd.alpha.local@ALPHA.LOCAL
- kdc returns a ticket then the client fails with KRB5KDC_ERR_PADATA_TYPE_NOSUPP

echo PASSWD | kinit -c evidence -f -S HTTP/srv-kerberos-kcd.alpha.local@ALPHA.LOCAL HTTP/srv-kerberos-kcd.alpha.local@ALPHA.LOCAL
/usr/heimdal/bin/kgetcred --delegation-credential-cache=evidence --out-cache=out -H HTTP/srv-kerberos-kcd.alpha.local@ALPHA.LOCAL
- kdc returns a ticket then the client fails with KRB5KDC_ERR_PADATA_TYPE_NOSUPP
```

jaltman:
@iboukris My preference is merge the refactoring commit as-is so its easier to see what the actual change in behavior is meant to be.   Please update the commit message for the first change to describe what the change does and why it was introduced.   

The second change looks good to me.  Automated tests would be much appreciated.

iboukris:
I managed to add a test for the protocol-transition case, but I'm having troubles with the constrained-delegation test. Since now we can't get an evidence ticket with self as client name via impersonation, I tried to get it via TGS or AS request, but this fails the constrained-delegation request regardless, as the KDC is not happy with an evidence ticket not acquired via impersonation.

I've commented it out while trying to understand the below error (the des error is probably unrelated):
```
2019-06-09T18:11:22 error message: No authorization data: 2
2019-06-09T18:11:22 TGS-REQ proxy-service@TEST.H5L.SE from IPv4:127.0.0.1 for host/datan.test.h5l.se@TEST.H5L.SE [canonicalize, cname-in-addl-tkt, forwardable]
2019-06-09T18:11:22 error message: encryption type des-cbc-crc is disabled: -1765328234
2019-06-09T18:11:22 error message: encryption type des-cbc-crc is disabled: -1765328234
2019-06-09T18:11:22 error message: Decrypt integrity check failed for checksum type hmac-sha1-96-aes256, key type aes256-cts-hmac-sha1-96: -1765328353
2019-06-09T18:11:22 KRB5SignedPath not signed correctly, not marking as signed
2019-06-09T18:11:22 Ticket not signed with PAC nor SignedPath service host/datan.test.h5l.se@TEST.H5L.SE failed for delegation to bar@TEST.H5L.SE for client proxy-service@TEST.H5L.SE (proxy-service@TEST.H5L.SE)from IPv4:127.0.0.1
2019-06-09T18:11:22 Failed building TGS-REP to IPv4:127.0.0.1
2019-06-09T18:11:22 tgs-req: sending error: -1765328371 to client
```

iboukris:
I think I figure out the TGS failure, I suspect it is a bug that we sign protocol-transition tickets with the impersonator principal inside KRB5SignedPath, I think it should be the impersonated one as in MIT. Still thinking though, and the case where the evidence is acquired via AS request still fails.

iboukris:
I think I figured the AS failure too, now both tests pass for me but it needs more thinking.

kdc/kerberos5.c
```diff
@@ -1705,6 +1705,31 @@ add_enc_pa_rep(kdc_request_t r)
 			   KRB5_PADATA_FX_FAST, NULL, 0);
 }
 
+static krb5_error_code
+get_local_tgt(krb5_context context,
```
lhoward: Bike-shedding, but prefer get_local_krbtgt() or get_local_tgs() as you're not getting a ticket.

kdc/kerberos5.c
```diff
@@ -1859,6 +1886,19 @@ _kdc_as_rep(kdc_request_t r,
 	goto out;
     }
 
+    /* If the server is not krbtgt, fetch local tgt key to sign authdata */
+    if (!is_tgs) {
+	Key *ad_skey;
```
lhoward: extra newline after variable definition

kdc/kerberos5.c
```diff
@@ -1859,6 +1886,19 @@ _kdc_as_rep(kdc_request_t r,
...
+	ret = get_local_tgt(context, config, r->client_princ->realm, &local_tgt);
+	if (ret)
+	    goto out;
```
lhoward: I would separate statements with an extra newline

kdc/kerberos5.c
```diff
@@ -2404,6 +2444,8 @@ _kdc_as_rep(kdc_request_t r,
 	krb5_crypto_destroy(r->context, r->armor_crypto);
 	r->armor_crypto = NULL;
     }
+    if(local_tgt)
```
lhoward: ```if (local_tgt)```

lib/krb5/ticket.c
```diff
@@ -714,61 +714,73 @@ _krb5_extract_ticket(krb5_context context,
...
 				     request == NULL); /* is TGS */
 	if (ret) {
-	    krb5_free_principal(context, tmp_principal);
 	    goto out;
 	}
     }
 
-    /* check client referral and save principal */
+    /* check KDC supported S4U extensions (MS-SFU) */
+    if((flags & EXTRACT_TICKET_MATCH_S4U_CLIENT) &&
```
lhoward: ```if ((flags...```

iboukris:
Hi Luke, I addressed your comments, but I moved the KRB5SignedPath changes to a separate PR as I need to think these changes some more and this PR got too overloaded anyway.

I suspect we'd need the get_local_tgs() logic (kinit -S) for PAC signing as well to sign the krbtgt checksum, but that's a little tricky to test right now (btw, I think MIT have the same bug).

At any rate, the changes in this PR look ready to me.

iboukris:
@jaltman @lhoward, I'm having second thoughts on this change - sorry.
CC: @greghudson 

I think the logic in MS-SFU in the first comment above makes a wrong assumption about kerberos implementations of canonicalize flag. It assumes that names are not canonicalized in response to TGS request (either per RFC4120 or RFC6806) and therefore if the KDC has changed the client name we can safely conclude that it supports S4U. However in MIT implementation for instance, we allow names to be canonicalized in response to TGS request (if canon requested), based on initial drafts of RFC6806. So I think we can't really conclude KDC S4U support just because the client name changed, it may have been just the impersonator name being canonicalized. See related discussion:
https://github.com/krb5/krb5/pull/912#issuecomment-507352647

In MIT we also do the negative check in S4U2Self, but there we also check that the returned client matches the expected impersonated client or at least we have verified KDC signature of PA-S4U-X509-USER, so the KDC supports S4U2Self for sure.

To sum it up, I think the correct way is to check that the returned client name is exactly the impersonated client we asked for, in both S4U protocols.

greghudson:
> However in MIT implementation for instance, we may allow names to be canonicalized in response to TGS request (if canon requested), based on initial drafts of RFC6806.

Are you saying the MIT KDC might change the names in TGS responses, or the MIT TGS client might allow changes to names?

iboukris:
> Are you saying the MIT KDC might change the names in TGS responses, or the MIT TGS client might allow changes to names?

I was referring to MIT TGS client we've discussed, but I now realize I'm confusing client and server canonicalization again, and MIT TGS client never allows client name to be canonicalized (in krb5int_process_tgs_reply).
For MIT KDC it would depend on the KDB module iirc.
I'll look for heimdal KDC and client behavior.

iboukris:
@greghudson I did some tests using heimdal test suit and the kdc won't canonicalize the client name. The client code would only allow client name to be canonicalized if EXTRACT_TICKET_ALLOW_CNAME_MISMATCH flag is set, and is only set in TGS for S4U. However, in samba code it looks like it is set unconditionally.

I think I still prefer matching against expected impersonated client, but it's debatable, and it does complicate it for S4U2Proxy when the code has no access to the expected client inside the evidence ticket.

lhoward:
I'm happy to merge this but, were you having second thoughts?

iboukris:
This one is tricky, I suspect it's going to get me troubles with samba test-suite, I need to think it more, so I'll change the status of this PR and PR #599 to draft - will revisit after PAC ticket-signature (sorry).

PR #599, opened 2019-06-10
draft: Fix KRB5SignedPath when evidence ticket is not acquired via protocol-transition

iboukris:
Follow up to #598.
- Fix check of KRB5SignedPath in constrained-delegation to match on impersonated client.
- Fix signing of KRB5SignedPath in protocol-transition to sign with impersonated client.
- Fix PAC and KRB5SignedPath KDC signatures in case of non krbtgt server requested via AS-RRQ to use local TGS key instead of the server key.

### DIFF
--- a/kdc/kerberos5.c
+++ b/kdc/kerberos5.c
@@ -1605,7 +1605,7 @@ send_pac_p(krb5_context context, KDC_REQ *req)
  */
 
 static krb5_error_code
-generate_pac(kdc_request_t r, Key *skey)
+generate_pac(kdc_request_t r, Key *skey, Key *tkey)
 {
     krb5_error_code ret;
     krb5_pac p = NULL;
@@ -1623,7 +1623,7 @@ generate_pac(kdc_request_t r, Key *skey)
     ret = _krb5_pac_sign(r->context, p, r->et.authtime,
 			 r->client->entry.principal,
 			 &skey->key, /* Server key */
-			 &skey->key, /* FIXME: should be krbtgt key */
+			 &tkey->key, /* krbtgt key */
 			 &data);
     krb5_pac_free(r->context, p);
     if (ret) {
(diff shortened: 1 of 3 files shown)


iboukris:
I revised the first commit (evidence acquired via TGS) and updated its commit message, hope it make sense.

iboukris:
I was able to confirm that we need the get_local_tgs() logic (kinit -S) for PAC signing as well to sign the kdc checksum, as Samba fails S4U2Proxy requests when the evidence ticket was acquired via AS due to that. I think the updated commit I just pushed should fix that, but I have yet to port it to samba code and test it.

lhoward:
Good stuff. I guess a temp variable could be cleaner than checking is_tgs but it doesn’t really matter to me.

kdc/kerberos5.c
```diff
@@ -1859,6 +1887,20 @@ _kdc_as_rep(kdc_request_t r,
 	goto out;
     }
 
+    /* If the server is not krbtgt, fetch local krbtgt key to sign authdata */
+    if (!is_tgs) {
+
```
lhoward: this blank line can go :)

iboukris: I'll make it go.
I also wonder if we should be more graceful here and not error out but just dismiss authdata, though I'm not sure that's really necessary.

iboukris:
I changed the second commit to avoid failing the request if we can't find a local tgs, instead we just sign with the server key in that case, as before.

I've also tested the logic related to PAC signing with a (somewhat) equivalent patch to samba code, and it fixes the issue, that is, kinit -S for evidence ticket works, see: https://github.com/iboukris/samba/commit/6318c32020930e9b3287346b51752d0fd6a33cdb

lhoward:
I reckon it’s ok to fail, you should never not have the local TGS key.

iboukris:
Thanks, switched back.

jaltman:
In a cross-realm situation we might not be able to build a `krbtgt/CLIENT-REALM@LOCAL-REALM` ticket.   Is the use of the client principal realm correct?   

Should it be the realm from the last entry in the transited path (if present)?

iboukris:
The changes are only for AS-REQ (the unusual case where non krbtgt server is requested) so I guessed the client must be local.

iboukris:
It looks like MIT constructs the local tgt using server's realm instead of client's, I think I should do the same: https://github.com/krb5/krb5/blob/01319a310cf06e4139b65afb12f998dbea636103/src/kdc/do_as_req.c#L668

jaltman:
The history of KRB5SignedPath is a bit convoluted because it started as a Heimdal feature which was then emulated by MIT.

lhoward:
Sounds good regarding server realm, although how common is it to serve multiple realms from the same KDC?

iboukris:
In AD a KDC serves only one realm. MS-PAC 2.8.2 simply states "The KDC signature is generated by the issuing KDC" but in general it's the KDC of the server realm that signs the service ticket. I changed it to server realm as suggested.

lhoward:
Right but KRB5SignedPath is a non-AD extension.

iboukris:
I realise that, my assumption is that it was modelled on MS-PAC. And regarding evidence ticket, it's the kdc of the server that would verify it.

lhoward:
True, yes makes sense :)

iboukris:
With the rest fixed by e7863e2af922809dad25a2e948e98c408944d551, closing this PR :)

PR #608, opened 2019-07-06
Add kx509 client and revamp kx509 service

nicowilliams:
This commit adds support for kx509 in libkrb5, and revamps the KDC's kx509 service (fixing bugs, adding features).

Of note is that kx509 is attempted optimistically by the client, with the certificate and private key stored in the ccache, and optionally in an external PEM or DER file.  This allows users to have PKIX credentials wherever they have Kerberos credentials.  If the service is configured to not require INITIAL tickets, then this functionality amounts to a PKIX credential delegation feature of sorts: wherever a user delegates Kerberos credentials, they can also automatically have PKIX credentials.

A number of related commits are in this PR.  Perhaps those should be separated and pushed first.

There's a lot of work to be done yet, but most if not all of it can be done in later commits.

The full list of commits is:

 - 0f97fec3a hx509: export hx509_parse_private_key (Windows)
 - 4233ef45a hx509: support reading private keys from PEM files
 - adaf680ef hxtool request-create: fix --key argument handling
 - 2fa8e21dd Add more hxtool EKU options, and KeyUsage too
 - a316c624d Add note about racy tests
 - 4d388ed77 Fix hx509_ca_tbs_free() segfault
 - 1091ceaff hxtool: add generate-key command
 - 90dca96c2 hx509: do not crash on missing FILE: name
 - 03a0dbe90 hx509: Add PKCS#8 private key format option
 - 8ba399c0a hx509: create files with safe mode_t
 - 88a01e0e8 hx509: add hx509_certs_destroy()
 - 412283a01 hx509: hx509_context_free(&null_var) should work
 - 64310ba7a hx509: make hx509_get_error_string(NULL, ...) work
 - aaf01a01f hxtool print: fix output device bugs
 - cb10e2e1c Fix missing error checking in lib/krb5/pkinit.c
 - c7b6a7aac kdc: fix invalid free at exit time
 - 2a3937aa9 kdc: fix leaks
 - a232ab702 pkinit: fix memory leak in libkrb5
 - d3f5f0f71 pkinit: fix leak in client
 - 22edd6bb0 klist: fix verbose printing of config entries
 - fac55f869 klist: partial fix of -v --json
 - b8ec34dca Add krb5_cc_configured_default_name()
 - ec1135100 Add kx509 client and revamp kx509 service

(ec1135100 being the HEAD commit).

### DIFF
--- a/doc/standardisation/rfc6717.txt
+++ b/doc/standardisation/rfc6717.txt
@@ -1,0 +1,731 @@
+
+
+
+
+
+
+Independent Submission                                           H. Hotz
+Request for Comments: 6717                   Jet Propulsion Lab, Caltech
+Category: Informational                                       R. Allbery
+ISSN: 2070-1721                                      Stanford University
+                                                             August 2012
+
+
+     kx509 Kerberized Certificate Issuance Protocol in Use in 2012
+
+Abstract
+
+   This document describes a protocol, called kx509, for using Kerberos
+   tickets to acquire X.509 certificates.  These certificates may be
+   used for many of the same purposes as X.509 certificates acquired by
+   other means, but if a Kerberos infrastructure already exists, then
+   the overhead of using kx509 may be much less.
+
+   While not standardized, this protocol is already in use at several
+   large organizations, and certificates issued with this protocol are
+   recognized by the International Grid Trust Federation.
+
+Status of This Memo
+
+   This document is not an Internet Standards Track specification; it is
+   published for informational purposes.
+
+   This is a contribution to the RFC Series, independently of any other
+   RFC stream.  The RFC Editor has chosen to publish this document at
+   its discretion and makes no statement about its value for
+   implementation or deployment.  Documents approved for publication by
+   the RFC Editor are not a candidate for any level of Internet
+   Standard; see Section 2 of RFC 5741.
+
+   Information about the current status of this document, any errata,
+   and how to provide feedback on it may be obtained at
+   http://www.rfc-editor.org/info/rfc6717.
+
+Copyright Notice
+
+   Copyright (c) 2012 IETF Trust and the persons identified as the
+   document authors.  All rights reserved.
+
+   This document is subject to BCP 78 and the IETF Trust's Legal
+   Provisions Relating to IETF Documents
+   (http://trustee.ietf.org/license-info) in effect on the date of
+   publication of this document.  Please review these documents
+   carefully, as they describe your rights and restrictions with respect
+   to this document.
+
+
+
+Hotz & Allbery                Informational                     [Page 1]
+
+RFC 6717                          kx509                      August 2012
+
+
+Table of Contents
+
+   1. Introduction ....................................................2
+      1.1. Requirements Language ......................................3
+   2. Protocol Data ...................................................3
+     2.1.  Request Packet .............................................3
+     2.2.  Reply Packet ...............................................4
+   3. Protocol Operation ..............................................7
+   4. Acknowledgements ................................................8
+   5. IANA Considerations .............................................8
+   6. Security Considerations .........................................9
+   7. References .....................................................10
+      7.1. Normative References ......................................10
+      7.2. Informative References ....................................10
+   Appendix A.  Certificate Caching and Deployment Considerations ....12
+   Appendix B.  Historic Extensions ..................................12
+   Appendix C.  Example Exchange .....................................12
+
+1.  Introduction
+
+   The two primary ways of providing cryptographically secure
+   identification on the Internet are Kerberos tickets [RFC4120] and
+   X.509 [RFC5280] [X.509] certificates.
+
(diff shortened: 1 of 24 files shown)


nicowilliams:
Missing functionality that needs to be there before pushing:

 - when extracting the private key and certificate from a ccache, update the ccache to list the hx509 store into which they were extracted, and

 - make kdestroy also destroy the hx509 store into which the PKIX creds were stored.

If the PKIX credentials are stored into multiple hx509 stores, we probably want to record each of their names.

nicowilliams:
Windows build fixed.

nicowilliams:
Short-term TBD (after pushing this):

 - add support for using a self-signed CSR or a self-signed Certificate so we can have:
    - proof-of-possession of SPKI
    - algorithm agility
    - a way to express requests for KU/EKU/SANs to be added / left out

Medium-term TBD:

 - finish ECC support in `lib/hx509` and switch to ECDSA as the default

Longer-term TBD:

 - add ed25519 and ed448 support, and switch to that as the default

 - port parts of the UMich CITI `kpkcs11` code to libhx509, with @jaltman's changes for supporting the Windows PKIX API

 - add a PKCS#11 IPC daemon so that we can keep apps from having direct access to private keys

 - similarly, port the Red Hat GSS-API proxy to Heimdal or write similar functionality so we can keep apps from having direct access to credentials

lhoward:
@nicowilliams One could probably shoehorn the PKCS#11 IPC daemon into kcm. Or indeed can you shovel them into the credentials cache so that you can leverage kcm/keyring as-is?

nicowilliams:
@lhoward I'm already setting the cert and priv key in the ccache as cc config entries!  So this should already work with kcm and the linux keyring ccache.

But kcm != isolation.  Clients get to see the session keys, and the tickets (cc config payloads are stored as tickets).

lhoward:
Is the Red Hat GSS proxy a separate thing that could be made to work with Heimdal? I suppose it's not BSD licensed though...

nicowilliams:
@lhoward yeah, it's not GPL, but BSD or MIT, I forget which.  I wrote the XDR for it, and as a result of that cooperation it got licensed with not-GPL.

It's got two components: a daemon, which only uses GSS and so should be trivial to port to Heimdal (or leave it out of tree even), and... invasive libgss mechglue changes that will basically need to be written from scratch for Heimdal.

nicowilliams:
Also, to make this truly useful we'll need to send PRs to OpenSSL and others, or to apps like curl/libcurl, to have some sort of configuration listing `{pkix_cred_store, hostname_pattern_or_domain}` instructing those TLS implementations or TLS apps when to use client certs (for what origins) and which certs.

kdc/kx509.c
```diff
@@ -415,114 +786,147 @@ _kdc_do_kx509(krb5_context context,
...
-
-    ret = kdc_kx509_verify_service_principal(context, cname, sprincipal);
-    if (ret)
+    /* Check that the service name is a valid kx509 service name */
+    if (ret == 0)
+        ret = krb5_ticket_get_server(context, ticket, &sprincipal);
+    if (ret == 0)
+        ret = krb5_unparse_name(context, sprincipal, &sname);
+    if (ret == 0)
+    if (ret == 0)
```
lhoward: superfluous if (ret == 0) here?

kuser/klist.c
```diff
@@ -161,7 +165,28 @@ print_cred_verbose(krb5_context context, krb5_creds *cred, int do_json)
...
+            s = krb5_principal_get_comp_string(context, cred->server, 1);
+            if (strcmp(s, "start_realm") == 0 ||
+                strcmp(s, "anon_pkinit_realm") == 0 ||
+                strcmp(s, "default-ntlm-domain") == 0 ||
+                strcmp(s, "FriendlyName") == 0 ||
+                strcmp(s, "fast_avail") == 0 ||
+                strcmp(s, "kx509store") == 0 ||
+                strcmp(s, "kx509_service_realm") ||
+                strcmp(s, "kx509_service_status") == 0)
+                printf(N_("Configuration item payload: %.*s\n", ""),
```
lhoward: maybe these could go in a table or a helper function?

doc/standardisation/rfc6717.txt
```diff
@@ -0,0 +1,339 @@
+
+
+
+
+
+
+Internet Engineering Task Force (IETF)                       K. Zeilenga
+Request for Comments: 6171                                 Isode Limited
```
lhoward: This is RFC6171 not 6717

nicowilliams: Hah, whoops, thanks!

kuser/klist.c
nicowilliams: In klist or in the library?

nicowilliams:
This was reviewed off-github by @vdukhovni.

PR #609, opened 2019-07-15
Optional backwards-compatible anon-pkinit behaviour

vdukhovni:
* Anonymous pkinit responses from the KDC where the name
  type is not well-known (as issued by 7.5 KDCs and earlier)
  are accepted by the client.  There is no need for the client
  to strictly enforce the name type.

* With historical_anon_pkinit = true, the kinit(1) client's
  "--anonymous" option only performs anon pkinit, and does
  not require an '@' prefix for the realm argument.

* With historical_anon_realm = true, the KDC issues anon
  pkinit tickets with the legacy pre-7.0 "real" realm.

### DIFF
--- a/kdc/default_config.c
+++ b/kdc/default_config.c
@@ -59,6 +59,7 @@ krb5_kdc_get_config(krb5_context context, krb5_kdc_configuration **config)
     c->check_ticket_addresses = TRUE;
     c->allow_null_ticket_addresses = TRUE;
     c->allow_anonymous = FALSE;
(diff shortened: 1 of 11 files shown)


nicowilliams:
Since all review is happening on #610, can we close this one?

vdukhovni:
I always open a separate PR per branch, the second review can wait for the first, but this branch will still need to be updated to match, and should be briefly reviewed.  We can put this on hold for now...

PR #610, opened 2019-07-15
Optional backwards-compatible anon-pkinit behaviour

vdukhovni:
* Anonymous pkinit responses from the KDC where the name
  type is not well-known (as issued by 7.5 KDCs and earlier)
  are accepted by the client.  There is no need for the client
  to strictly enforce the name type.

* With historical_anon_pkinit = true, the kinit(1) client's
  "--anonymous" option only performs anon pkinit, and does
  not require an '@' prefix for the realm argument.

* With historical_anon_realm = true, the KDC issues anon
  pkinit tickets with the legacy pre-7.0 "real" realm.

### DIFF
--- a/kdc/default_config.c
+++ b/kdc/default_config.c
@@ -59,6 +59,7 @@ krb5_kdc_get_config(krb5_context context, krb5_kdc_configuration **config)
     c->check_ticket_addresses = TRUE;
     c->allow_null_ticket_addresses = TRUE;
     c->allow_anonymous = FALSE;
+    c->historical_anon_realm = FALSE;
     c->strict_nametypes = FALSE;
     c->trpolicy = TRPOLICY_ALWAYS_CHECK;
     c->enable_pkinit = FALSE;
@@ -163,6 +164,12 @@ krb5_kdc_get_config(krb5_context context, krb5_kdc_configuration **config)
 				     c->allow_anonymous,
 				     "kdc",
 				     "allow-anonymous", NULL);
+
+    c->historical_anon_realm =
+	krb5_config_get_bool_default(context, NULL,
+				     c->historical_anon_realm,
+				     "kdc",
+				     "historical_anon_realm", NULL);
 
     c->strict_nametypes =
 	krb5_config_get_bool_default(context, NULL,
--- a/kdc/kdc.8
+++ b/kdc/kdc.8
@@ -155,6 +155,19 @@ Permit tickets with no addresses.
 This option is only relevant when check-ticket-addresses is TRUE.
 .It Li allow-anonymous = Va boolean
 Permit anonymous tickets with no addresses.
+.It Li historical_anon_realm = Va boolean
+Enables pre-7.0 non-RFC-comformant KDC behavior.
+With this option set to
+.Li true
+the client realm in anonymous pkinit AS replies will be the requested realm,
+rather than the RFC-conformant
+.Li WELLKNOWN:ANONYMOUS
+realm.
+This can have a security impact on servers that expect to grant access to
+anonymous-but-authenticated to the KDC users of the realm in question:
+they would also grant access to unauthenticated anonymous users.
+As such, it is not recommend to set this option to
+.Li true.
 .It Li max-kdc-datagram-reply-length = Va number
 Maximum packet size the UDP rely that the KDC will transmit, instead
 the KDC sends back a reply telling the client to use TCP instead.
--- a/kuser/kinit.c
+++ b/kuser/kinit.c
@@ -668,7 +668,7 @@ get_new_tickets(krb5_context context,
 	}
     } else if (pk_user_id || ent_user_id ||
 	       krb5_principal_is_anonymous(context, principal, KRB5_ANON_MATCH_ANY)) {
-
+        /* nop */;
     } else if (!interactive && passwd[0] == '\0') {
 	static int already_warned = 0;
 
@@ -1271,6 +1271,7 @@ main(int argc, char **argv)
     struct sigaction sa;
 #endif
     krb5_boolean unique_ccache = FALSE;
+    krb5_boolean historical_anon_pkinit = FALSE;
     int anonymous_pkinit = FALSE;
 
     setprogname(argv[0]);
@@ -1298,6 +1299,9 @@ main(int argc, char **argv)
 
     argc -= optidx;
     argv += optidx;
+
+    krb5_appdefault_boolean(context, "kinit", NULL, "historical_anon_pkinit",
+                            FALSE, &historical_anon_pkinit);
 
     /*
      * Open the keytab now, we use the keytab to determine the principal's
@@ -1327,6 +1331,16 @@ main(int argc, char **argv)
 	ret = krb5_make_principal(context, &principal, &argv[0][1],
 				  KRB5_WELLKNOWN_NAME, KRB5_ANON_NAME,
 				  NULL);
+	if (ret)
+	    krb5_err(context, 1, ret, "krb5_make_principal");
+	krb5_principal_set_type(context, principal, KRB5_NT_WELLKNOWN);
+	anonymous_pkinit = TRUE;
+    } else if (anonymous_flag && historical_anon_pkinit) {
+        char *realm = argc == 0 ? get_default_realm(context) :
+                      argv[0][0] == '@' ? &argv[0][1] : argv[0];
+
+	ret = krb5_make_principal(context, &principal, realm,
+				  KRB5_WELLKNOWN_NAME, KRB5_ANON_NAME, NULL);
 	if (ret)
 	    krb5_err(context, 1, ret, "krb5_make_principal");
 	krb5_principal_set_type(context, principal, KRB5_NT_WELLKNOWN);
--- a/lib/krb5/krb5.h
+++ b/lib/krb5/krb5.h
@@ -955,8 +955,11 @@ typedef struct krb5_name_canon_iterator_data *krb5_name_canon_iterator;
  */
 #define KRB5_ANON_MATCH_AUTHENTICATED	1 /* authenticated with anon flag */
 #define KRB5_ANON_MATCH_UNAUTHENTICATED	2 /* anonymous PKINIT */
-#define KRB5_ANON_MATCH_ANY		( KRB5_ANON_MATCH_AUTHENTICATED | KRB5_ANON_MATCH_UNAUTHENTICATED )
-
+#define KRB5_ANON_IGNORE_NAME_TYPE	4 /* don't check the name type */
+#define KRB5_ANON_MATCH_ANY	        ( KRB5_ANON_MATCH_AUTHENTICATED | \
+                                          KRB5_ANON_MATCH_UNAUTHENTICATED )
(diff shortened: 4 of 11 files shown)


kdc/kdc.8
```diff
@@ -155,6 +155,20 @@ Permit tickets with no addresses.
...
+rather than the RFC-conformant
+.Li WELLKNOWN:ANONYMOUS
+realm.
+This can have a security impact on servers that expect to grant access to
+anonymous, but authenticated to the KDC users of the realm in question.
+They would also grant access to unauthenticated anonymous users.
+The
+.Li true
+settings is not recommended.
 .It Li max-kdc-datagram-reply-length = Va number
```
lhoward: Perhaps reword something like the following?

With this option set to true, the client realm in anonymous PKINIT AS replies will be the requested realm, rather than the well known anonymous name. Unauthenticated anonymous principals will appear to services as authenticated anonymous. As such, it is not recommend to set this option to true.

kuser/kinit.c
```diff
@@ -1331,6 +1337,16 @@ main(int argc, char **argv)
 	    krb5_err(context, 1, ret, "krb5_make_principal");
 	krb5_principal_set_type(context, principal, KRB5_NT_WELLKNOWN);
 	anonymous_pkinit = TRUE;
+    } else if (anonymous_flag && historical_anon_pkinit) {
+        char *realm = argc == 0 ? get_default_realm(context) :
```
lhoward: Potential memory leak, `get_default_realmI()` returns allocated memory.

lib/krb5/principal.c
```diff
@@ -1260,17 +1260,43 @@ krb5_principal_is_anonymous(krb5_context context,
...
-         p->name.name_type != KRB5_NT_UNKNOWN) ||
-        p->name.name_string.len != 2 ||
+    /*
+     * Heimdal versions 7.5 and below left the name-type at KRB5_NT_PRINCIPAL
+     * even with anonymous pkinit responses.  To retain interoperability with
+     * legacy KDCs, the name-type is not checked by the client after requesting
+     * a fully anonymous ticket.
+     */
+    if ((flags & KRB5_ANON_MATCH_NAME_TYPE) &&
+        p->name.name_type != KRB5_NT_WELLKNOWN &&
```
lhoward: If `krb5_principal_is_anonymous()` has shipped (I think it has?) then it would be better not to change the behaviour. Rather, I would suggest inverting the logic so the additional flag is `KRB5_ANON_MATCH_IGNORE_NT` for example.

lib/krb5/krb5.h
```diff
@@ -955,8 +955,9 @@ typedef struct krb5_name_canon_iterator_data *krb5_name_canon_iterator;
  */
 #define KRB5_ANON_MATCH_AUTHENTICATED	1 /* authenticated with anon flag */
 #define KRB5_ANON_MATCH_UNAUTHENTICATED	2 /* anonymous PKINIT */
+#define KRB5_ANON_MATCH_NAME_TYPE	4 /* match well-known or unknown */
 #define KRB5_ANON_MATCH_ANY		( KRB5_ANON_MATCH_AUTHENTICATED | KRB5_ANON_MATCH_UNAUTHENTICATED )
-
+#define KRB5_ANON_MATCH_ANY_NT		( KRB5_ANON_MATCH_ANY | KRB5_ANON_MATCH_NAME_TYPE )
```
lhoward: Would prefer to be consistent, either use NT or NAME_TYPE but don't mix.

lib/krb5/principal.c
```diff
@@ -1260,17 +1260,43 @@ krb5_principal_is_anonymous(krb5_context context,
...
-    anon_realm = strcmp(p->realm, KRB5_ANON_REALM) == 0;
+    /*
+     * While unauthenticated clients SHOULD get "WELLKNOWN:ANONYMOUS" as their
+     * realm, Heimdal KDCs prior to 7.0 returned the requested realm.  While
+     * such tickets might lead *servers* to unwittingly grant access to fully
+     * anonymous clients, trusting that the client was authenticated to the
+     * realm in question, doing it right is the KDCs job, the client should not
+     * refuse such a ticket.
+     *
+     * If we ever do decide to enforice WELLKNOWN:ANONYMOUS for unathenticted
```
lhoward: s/unathenticated/unauthenticated/
s/enforice/enforce/

lib/krb5/principal.c
```diff
@@ -1260,17 +1260,43 @@ krb5_principal_is_anonymous(krb5_context context,
...
         strcmp(p->name.name_string.val[1], KRB5_ANON_NAME) != 0)
         return FALSE;
 
-    anon_realm = strcmp(p->realm, KRB5_ANON_REALM) == 0;
+    /*
+     * While unauthenticated clients SHOULD get "WELLKNOWN:ANONYMOUS" as their
+     * realm, Heimdal KDCs prior to 7.0 returned the requested realm.  While
+     * such tickets might lead *servers* to unwittingly grant access to fully
+     * anonymous clients, trusting that the client was authenticated to the
+     * realm in question, doing it right is the KDCs job, the client should not
```
lhoward: s/KDCs/KDC's/

lib/krb5/principal.c
```diff
@@ -1260,17 +1260,43 @@ krb5_principal_is_anonymous(krb5_context context,
...
+     * refuse such a ticket.
+     *
+     * If we ever do decide to enforice WELLKNOWN:ANONYMOUS for unathenticted
+     * clients, it is essential that calls that pass KRB5_ANON_MATCH_ANY still
+     * ignore the realm, as in that case either case matches one of the two
+     * possible conditions.
+     */
+    if (flags & KRB5_ANON_MATCH_UNAUTHENTICATED)
+        return TRUE;
```
lhoward: Hmm, I don't know about this. Other (not merged) code such as the `GSS_C_ANON_FLAG` support relies on this function. I would prefer to put the backwards compatibility logic into the caller and have `krb5_principal_is_anonymous()` be RFC compliant.

vdukhovni: Most of the other comments seem reasonable, and easy to fix.  Here I disagree.  If this function is going to enforce the name type, then what do you suggest a caller that wants to interoperate with &le; 7.5 Heimdal KDCs should do?  The function becomes unusable and we need a new one? Or what?

lhoward: I will need to think it through more carefully. My gut feel is that –

* the function should by default implement the RFC compliant behavior – the UNAUTHENTICATED test matches unauthenticated only, and the AUTHENTICATED test matches authenticated only
* if compatibility behavior is desired, it should preferably be set by the caller with additional flags, or failing that, krb5.conf should be checked within the function

FWIW, the function was originally supposed to be an internal function but it was exposed because it was difficult to expose the flags without adding them to krb5.h (krb5_locl.h can't be easily imported outside lib/krb5). And if the flags are to be public, then we thought so should the function.

kdc/default_config.c
```diff
@@ -59,6 +59,7 @@ krb5_kdc_get_config(krb5_context context, krb5_kdc_configuration **config)
     c->check_ticket_addresses = TRUE;
     c->allow_null_ticket_addresses = TRUE;
     c->allow_anonymous = FALSE;
+    c->historical_anon_realm = FALSE;
```
elric1: I think that we should consider making the default true in the 7.x releases. This is a breaking change for Heimdal shops and such changes should be restricted to major releases.

lhoward: I defer to @jaltman on this one. Arguably the previous behavior is a security issue, as it reveals the realm for anonymous PKINIT.

jaltman: I agree with @lhoward.  The combination of anonymous PKINIT plus cross-realm results in a loss of privacy which is a security issue.   We should belatedly create a CVE describing the prior behavior.

elric1: Right, but this potentially actually breaks existing installations which may have come to depend on the prior behaviour. This kind of change should generally be reserved for a major release or a default off option.

I don't think that the severity of the issue is severe enough to break things. It's a small reduction in anonymity only. I haven't properly checked, but can't you just back track the transited path and see where they came from in any case? In a cursory glance at the code, it appears that your client's realm will be added to the path because crealm != WELLKONWN:ANONYMOUS.

vdukhovni: I don't even see it as a reduction in anonymity, since the real realm will be in the transit path.

lhoward: Hmm, well yes potentially adding the anonymous client's realm is a bug in the KDC.

kdc/kdc.8
vdukhovni: Done in the next update.

kuser/kinit.c
vdukhovni: This code runs exactly once in `main()`. I am not inclined to bother...

lib/krb5/principal.c
vdukhovni: OK, done in next update.

lib/krb5/principal.c
vdukhovni: Next update.

lib/krb5/principal.c
vdukhovni: Next update.

lib/krb5/principal.c
vdukhovni: I still think the client is not the right place to enforce lax policy by the KDC.  I'm leaving this as-is for now.

kdc/default_config.c
vdukhovni: I've not yet restored the historical default, will chat with Roland and Nico today or tomorrow.

nicowilliams: IMO the loss of privacy is not CVE-worthy.  Does anyone use Kerberos, with Heimdal 7.x, in such a way that they have cross-realm trusts and apps where there's a privacy concern?  I don't believe anyone is.

kdc/kdc.8
```diff
@@ -155,6 +155,25 @@ Permit tickets with no addresses.
 This option is only relevant when check-ticket-addresses is TRUE.
 .It Li allow-anonymous = Va boolean
 Permit anonymous tickets with no addresses.
+.It Li historical_anon_realm = Va boolean
+Enables pre-7.0 non-RFC-comformant KDC behavior.
+With this option set to
+.Li true
+the client realm in anonymous pkinit as-replies will be the requested realm,
```
nicowilliams: s/as-replies/AS replies/

kdc/kdc.8
```diff
@@ -155,6 +155,25 @@ Permit tickets with no addresses.
...
+rather than the RFC-conformant
+.Li WELLKNOWN:ANONYMOUS
+realm.
+This can have a security impact on servers that expect to grant access to
+anonymous, but authenticated to the KDC users of the realm in question.
```
nicowilliams: s/anonymous, but authenticated/anonymous-but-authenticated/

nicowilliams: Also, replace the sentence-ending period with a colon.

vdukhovni: Sure

PR #620, opened 2019-08-25
Issue 619:

SethEllsworth:
Heimdal's read_string() does not restart I/O when interrupted by a
    signal, returning with an error code instead. We don't want this to
    happen when the user resizes the window triggering a SIGWINCH, so we
    skip installing a handler for it. The default action for this signal is
    "ignore", so the I/O won't be interrupted (no need to set SA_RESTART).

### DIFF
--- a/lib/hcrypto/ui.c
+++ b/lib/hcrypto/ui.c
@@ -125,7 +125,7 @@ read_string(const char *preprompt, const char *prompt,
     sigemptyset(&sa.sa_mask);
     sa.sa_flags = 0;
     for(i = 1; i < sizeof(sigs) / sizeof(sigs[0]); i++)
-	if (i != SIGALRM)
+	if (i != SIGALRM && i != SIGWINCH)
 	    if (sigaction(i, &sa, &sigs[i]) == 0)
 		oksigs[i] = 1;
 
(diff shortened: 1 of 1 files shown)


lib/hcrypto/ui.c
```diff
@@ -125,7 +125,7 @@ read_string(const char *preprompt, const char *prompt,
     sigemptyset(&sa.sa_mask);
     sa.sa_flags = 0;
     for(i = 1; i < sizeof(sigs) / sizeof(sigs[0]); i++)
-	if (i != SIGALRM)
+	if (i != SIGALRM && i != SIGWINCH)
```
jaltman: SIGWINCH isn't defined on all platforms so its use must be conditionalized 

   #if defined(SIGWINCH)
       && i != SIGWINCH
   #endif

SethEllsworth: How does this look:
`    if (i != SIGALRM
#if defined(SIGWINCH)
            && i != SIGWINCH
#endif
       )
`

SethEllsworth: Added the whole line to the condition, and tried to match the indentation style.

nicowilliams: This function makes me sad.  Its purpose is to prompt, of course, possibly with echo off, and to terminate when the user... types `^C` I guess.  But messing with signal handling in a library function is a bad idea.  A better way to write this would be to not use stdio, block signals, put the terminal into raw mode, and look for `^C` (or whatever it's been mapped to with `stty`) instead of catching signals, then restore the tty mode to cooked (if it had been cooked) or whatever it was set to prior to returning.

Anyways, the change looks fine to me.  Can you squash commits and give the one commit a subject like "lib/hcrypto/ui: ignore SIGWINCH" or similar?  Or would you like us to do this?

PR #622, opened 2019-09-04
Make gss_store_cred*() work

nicowilliams:
krb5_cc_cache_match() searches all ccache collections for a ccache that
has credentials for a given principal name.  This includes MEMORY
ccaches, which means it can find the same ccache as is referenced by a
GSS cred handle given to gss_store_cred(), which means that
gss_store_cred() can fail.

For now we work around this by including a private variant of
krb5_cc_cache_match() that only searches the default ccache, not all
collections.  Eventually we should ensure that krb5_cc_default() also
searches all collection-type (other than MEMORY) ccaches for a default
credential, then we can go back to using krb5_cc_cache_match() (though
we'll need to make sure that MEMORY is searched last or not at all).

### DIFF
--- a/lib/gssapi/krb5/store_cred.c
+++ b/lib/gssapi/krb5/store_cred.c
@@ -33,6 +33,72 @@
 
 #include "gsskrb5_locl.h"
 
+static int
+same_princ(krb5_context context, krb5_ccache id1, krb5_ccache id2)
+{
+    krb5_error_code ret;
+    krb5_principal p1 = NULL;
+    krb5_principal p2 = NULL;
+    int same = 0;
+
+    ret = krb5_cc_get_principal(context, id1, &p1);
+    if (ret == 0)
+        ret = krb5_cc_get_principal(context, id2, &p2);
+    if (ret == 0)
+        same = krb5_principal_compare(context, p1, p2);
+    krb5_free_principal(context, p1);
+    krb5_free_principal(context, p2);
+    return same;
+}
+
+/*
+ * Like krb5_cc_cache_match(), but only looking in the default collection.
+ *
+ * We need this to avoid looking for MEMORY ccaches, which risks matching the
+ * same credential that we're storing.  We could make sure that MEMORY ccaches
+ * are searched for last in krb5_cc_cache_match(), then ignore any MEMORY
+ * ccaches we find there, but, if we might then store in a ccache that will not
+ * be found later as the default ccache, then it's not worth it.
+ *
+ * XXX In order to remove this, we'll first need to make sure that
+ *     krb5_cc_default() searches all collections when KRB5CCNAME is not set,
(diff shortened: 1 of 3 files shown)


nicowilliams:
I've spent too much time on this.  For a while I wanted to redesign how collections work, but in the end I've decided that the cccol APIs are OK as-is.

The thing that's strange about the cccol APIs is that they operate on the default collection of credentials caches, not on any specified one (there is no way to specify one).  The idea is that we should always have just a default ccache that is a collection and then store multiple principals' credentials there.  Fine.

There are usability issues with the current cccol APIs and the UI (`kswitch`) though.  Specifically, it's not really OK to have to globally `kswitch`.  In principle one can set `KRB5CCNAME` to point to a particular ccache in the collection should one know its name, but in practice that's a user-unfriendly interface.  To ameliorate this I will eventually (not in this PR) propose the following:

 - that we add a `KRB5CCPRINC=<principal>` env var for selecting a default ccache from the default collection that matches a particular principal
 - perhaps also that we name ccaches in `DIR` collections as the hash of the principal being stored

lib/gssapi/krb5/store_cred.c
```diff
@@ -46,125 +112,143 @@ _gsskrb5_store_cred_into(OM_uint32         *minor_status,
...
         /* Nothing to do */
+        HEIMDAL_MUTEX_unlock(&input_cred->cred_id_mutex);
         krb5_cc_close(context, id);
-        HEIMDAL_MUTEX_unlock(&cred->cred_id_mutex);
         *minor_status = 0;
         return GSS_S_DUPLICATE_ELEMENT;
     }
 
-    ret = krb5_cc_initialize(context, id, cred->principal);
+    ret = krb5_cc_initialize(context, id, input_cred->principal);
```
vdukhovni: Sadly, for the default fcache, this likely truncates the file, leaving concurrent users briefly out of luck. Fixing this, might also make it possible to fix "kinit -R", ...

nicowilliams: We'll fix this separately.

lib/gssapi/krb5/store_cred.c
```diff
@@ -46,125 +112,143 @@ _gsskrb5_store_cred_into(OM_uint32         *minor_status,
...
-
-    if (cred->principal == NULL) {
-	HEIMDAL_MUTEX_unlock(&cred->cred_id_mutex);
+    if (input_cred->principal == NULL) {
 	*minor_status = GSS_KRB5_S_KG_TGT_MISSING;
-	return GSS_S_FAILURE;
+	return GSS_S_NO_CRED;
     }
 
+    /* Extact the ccache name from the store if given */
```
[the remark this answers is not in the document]
nicowilliams: Fixed.

PR #631, opened 2019-09-26
Support builds as non-logged-in root

nicowilliams:
We add a new `%{loginname}` to mean "that which is returned by `getlogin_r()`", and make `%{username}` mean only whatever `$USER`, or `$LOGNAME`, or `getpwuid_r()` say.  And we make sure that the default principal name, and the principal name `kadmin` defaults to, are based on the former when running as root and the latter otherwise.


nicowilliams:
@vdukhovni reviewed.

PR #640, opened 2019-10-23
In krb5.conf make underscores and dashes equivalent

elric1:
The source of endless typos due to inconsistency. Why not accept both...


lib/krb5/config_file.c
```diff
@@ -83,6 +83,28 @@ static krb5_error_code parse_list(struct fileptr *f, unsigned *lineno,
 				  krb5_config_binding **parent,
 				  const char **err_message);
 
+static int
+todash(int c)
+{
+
```
nicowilliams: What's with this empty line?

lib/krb5/config_file.c
```diff
@@ -83,6 +83,28 @@ static krb5_error_code parse_list(struct fileptr *f, unsigned *lineno,
...
+    if (c == '_')
+	return '-';
+    return c;
+}
+
+static int
+strcmp_du(const char *s1, const char *s2)
+{
+
```
nicowilliams: Ditto.

lib/krb5/config_file.c
elric1: KNF, decls\ncode.

nicowilliams:
Can we make `lib/krb5/verify_krb5_conf.c` warn about dashes?

elric1:
I'm looking at it.

abartlet:
What does MIT krb5 do in this case?  Given the number of mixed installs
(Samba using Heimdal on a typical MIT-based Linux system for example),
I think these should be in as lock-step as possible.  The includedir
exension on the MIT side has caused a massive headache for example.

More broadly my concern from the long and painful history in Samba is
that synonyms can bite.  We (Samba) even allowed common misspelings,
which then have to be maintained forevermore!

The big issue is that you can't really know what tools outside Heimdal
(eg custom sed scripts) are parsing the krb5.conf and might therefore
fail until it is too late (because this is widely deployed).

The other issue, which may not be a realisic concern, is that while
easy in a custom file parser, each special case like this makes moving
to a standard 'ini' file parser much harder in the future. 

I do realise this would be more ergonomic, but I hope this provides
some useful counterpoints to help while you conssider the implications
here. 

Thanks!

Andrew Bartlett

elric1:
Yes, understood.
What we have at the moment is inconsistency. We're trying to fix it. It appears that MIT Kerberos uses underscores exclusively from their documentation, but we use both. I think that we should aim to move to underscores but we need to preserve backwards compatibility.

elric1:
We also have a lot of inconsistencies, i.e. we have "no-addresses" and MIT has "noaddresses".

frozencemetery:
@abartlet Can you elaborate on what you mean by the includedir extension?  Heimdal added that in #248 - are you just referring to the time where MIT had it and Heimdal didn't yet?

We (freeipa folks) have looked into using "standard" INI parsers, and the conclusion is that it will probably never happen - there are too many oddities.  Best we can do is export a good parser interface from the Kerberos implementation itself.

vdukhovni (approved):
LGTM.

jaltman:
@greghudson what is your opinion on this proposal?

If MIT agrees I would take the change a step further and on the left hand side convert "_", "-" and "." to "".    But I don't think Heimdal should take this change on its own.

greghudson:
I am concerned that I can't find prior art among standard config formats--YAML, TOML, and JSON (as well as JSON5 and HJSON) all have case-sensitive and punctuation-sensitive keys as far as I can tell.  Windows INI files have case-insensitive keys as you'd expect from Windows, but no punctuation-insensitivity that I'm aware of.

I'm not dead set against the idea, but I share some of the concerns voiced by @abartlet.  Any application of the Postel principle comes with a cost, although the benefits are greater in the config file space than the protocol space because humans are involved.

jaltman:
@greghudson I believe the precedent is gmail e-mail addresses which discard punctuation.

I believe the critical concern here is how to address the fact that in the past MIT and Heimdal have selected similar but not identical names for common settings.   The AuriStorFS command parser, which uses the krb5 profile format, addresses this by constructing manual alias tables which accept all of the acceptable left-hand-side value names for a given value.   

Case-sensitive left-hand-side values are slightly problematic for Heimdal on Windows because of its support for Windows Registry backed profiles which are specified by "REG:<key-path>/<value>".

Whether we go with this approach or not, I believe there would be value in constructing a public registry of known krb5.conf and kdc.conf profile sections and values along with their documented semantics.  If organizations that use krb5 profiles are encouraged to publish their usage, we can hopefully avoid confusion in the future.   

Heimdal and MIT can also use the registry to implement compatibility lookup code to improve interoperability when sharing a profile between both implementations.

elric1:
I've rebased to current, let's keep chatting about this and come to a conclusion.

lib/base/config_file.c
```diff
@@ -86,6 +86,26 @@ static heim_error_code parse_list(struct fileptr *f, unsigned *lineno,
                                   heim_config_binding **parent,
                                   const char **err_message);
 
+static int
+todash(int c)
```
nicowilliams, line 90: I'd have made this `tounderscore()` since that's what we want the canonical form to be, however since we only use this for comparisons, I don't care.

nicowilliams:
I'm inclined to merge this.  I'd consider also changing `lib/krb5/krb5.conf.5` so that we only show parameters with underscores.  And maybe warn about use of dashes in `lib/krb5/verify_krb5_conf.c`.

PR #643, opened 2019-10-30
Delegate destination TGT support

elric1:
This allows you to configure (in `/etc/krb5.conf`) that certain realms will "delegate-destination-tgt".  That is, rather than delegating a full TGT of the form krbtgt/START@START, you delegate a TGT of the form krbtgt/DEST@DEST. The interesting thing here is that the destination realm becomes a bit like the hotel california because you can't come back to the original realm as that will violate its transited policy.

This functionality works well for DMZ realms and the like where you want to allow access---and you want to be able to delegate creds which people can use to continue to SSH around in the DMZ, but you do not want them to be able to SSH back into the main environment.

### DIFF
--- a/lib/krb5/Makefile.am
+++ b/lib/krb5/Makefile.am
@@ -195,6 +195,7 @@ dist_libkrb5_la_SOURCES =			\
 	log.c					\
 	mcache.c				\
 	misc.c					\
+	mk_cred.c				\
 	mk_error.c				\
 	mk_priv.c				\
 	mk_rep.c				\
--- a/lib/krb5/mk_cred.c
+++ b/lib/krb5/mk_cred.c
@@ -1,0 +1,323 @@
+/*
+ * Copyright (c) 1997 - 2004 Kungliga Tekniska Högskolan
+ * (Royal Institute of Technology, Stockholm, Sweden).
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * 3. Neither the name of the Institute nor the names of its contributors
+ *    may be used to endorse or promote products derived from this software
+ *    without specific prior written permission.
+ *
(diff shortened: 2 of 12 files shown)


lib/krb5/Makefile.am
```diff
@@ -195,6 +195,7 @@ dist_libkrb5_la_SOURCES =			\
 	log.c					\
 	mcache.c				\
 	misc.c					\
+	mk_cred.c				\
```
nicowilliams: This change also needs to be made to `lib/krb5/NTMakefile`.

lib/krb5/mk_cred.c
```diff
@@ -0,0 +1,323 @@
...
+            }                                                   \
+        } while (0)
+
+#define CHECKED_COPY(cp_func, dst, src) do {                    \
+            if (cp_func(src, dst)) {                            \
+                ret = krb5_enomem(context);                     \
+                goto out;                                       \
+            }                                                   \
+        } while (0)
+#define CHECKED_COPY_PPC2KCI(cp_func, dst, src)                 \
```
nicowilliams: Not used.

lib/krb5/mk_cred.c
```diff
@@ -0,0 +1,323 @@
...
+        int32_t usec;
+
+        krb5_us_timeofday (context, &sec, &usec);
+
+        CHECKED_ALLOC_ASSIGN(enc_krb_cred_part.timestamp, sec);
+        CHECKED_ALLOC_ASSIGN(enc_krb_cred_part.usec, usec);
+    } else {
+        enc_krb_cred_part.timestamp = NULL;
+        enc_krb_cred_part.usec = NULL;
+        /* XXX Er, shouldn't we set the seq nums?? */
```
nicowilliams: For GSS this won't matter.  For other protocols, it might, but we can fix it if we ever need it (we won't).

lib/krb5/mk_cred.c
```diff
@@ -0,0 +1,323 @@
...
+    if (auth_context->local_address && auth_context->local_port) {
+        ret = krb5_make_addrport(context,
+                                 &enc_krb_cred_part.s_address,
+                                 auth_context->local_address,
+                                 auth_context->local_port);
+        if (ret)
+            goto out;
+    }
+
+    /* XXX: Is this needed? */
```
nicowilliams: Good question.  We should look at the rd side to find out.

lib/krb5/Makefile.am
nicowilliams: And the appveyor build works now.  (Still failing to build doc, but whatever.)

PR #647, opened 2019-11-07
Strftime expand tokens

elric1:
So, here we define %{strftime:<string>} in our path expansion routines and then subject logging file definitions to token expansion. Now, we can do something like this:

[kdc]
        kdc = FILE:/var/log/auth/%{strftime:%Y%m%d%H}

for nice automatically rolling over logs.

### DIFF
--- a/lib/krb5/expand_path.c
+++ b/lib/krb5/expand_path.c
@@ -52,7 +52,8 @@ typedef int PTYPE;
  * the returned path may or may not exist.
  */
 static krb5_error_code
-_expand_temp_folder(krb5_context context, PTYPE param, const char *postfix, char **ret)
+_expand_temp_folder(krb5_context context, PTYPE param, const char *postfix,
+		    const char *arg, char **ret)
 {
     TCHAR tpath[MAX_PATH];
     size_t len;
@@ -89,7 +90,8 @@ extern HINSTANCE _krb5_hInstance;
  * krb5.dll is located.
  */
 static krb5_error_code
-_expand_bin_dir(krb5_context context, PTYPE param, const char *postfix, char **ret)
+_expand_bin_dir(krb5_context context, PTYPE param, const char *postfix,
+		const char *arg, char **ret)
 {
     TCHAR path[MAX_PATH];
     TCHAR *lastSlash;
@@ -139,7 +141,8 @@ _expand_bin_dir(krb5_context context, PTYPE param, const char *postfix, char **r
  *
  */
 static krb5_error_code
-_expand_userid(krb5_context context, PTYPE param, const char *postfix, char **ret)
+_expand_userid(krb5_context context, PTYPE param, const char *postfix,
+	       const char *arg, char **ret)
 {
     int rv = EINVAL;
     HANDLE hThread = NULL;
@@ -237,7 +240,8 @@ _expand_userid(krb5_context context, PTYPE param, const char *postfix, char **re
  */
 
 static krb5_error_code
-_expand_csidl(krb5_context context, PTYPE folder, const char *postfix, char **ret)
+_expand_csidl(krb5_context context, PTYPE folder, const char *postfix,
+	      const char *arg, char **ret)
 {
     TCHAR path[MAX_PATH];
     size_t len;
@@ -266,7 +270,8 @@ _expand_csidl(krb5_context context, PTYPE folder, const char *postfix, char **re
 #else
 
 static krb5_error_code
-_expand_path(krb5_context context, PTYPE param, const char *postfix, char **ret)
+_expand_path(krb5_context context, PTYPE param, const char *postfix,
+	     const char *arg, char **ret)
 {
     *ret = strdup(postfix);
     if (*ret == NULL)
@@ -275,7 +280,8 @@ _expand_path(krb5_context context, PTYPE param, const char *postfix, char **ret)
 }
 
 static krb5_error_code
-_expand_temp_folder(krb5_context context, PTYPE param, const char *postfix, char **ret)
+_expand_temp_folder(krb5_context context, PTYPE param, const char *postfix,
+		    const char *arg, char **ret)
 {
     const char *p = NULL;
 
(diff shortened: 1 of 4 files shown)


lib/krb5/expand_path.c
```diff
@@ -349,6 +359,19 @@ _expand_loginname(krb5_context context, PTYPE param, const char *postfix, char *
...
 
+static krb5_error_code
+_expand_strftime(krb5_context context, PTYPE param, const char *postfix,
+		 const char *arg, char **ret)
+{
+    time_t t;
+    char buf[1024];
+
+    t = time(NULL);
+    strftime(buf, sizeof(buf), arg, localtime(&t));
```
vdukhovni: The return value of strftime should probably be checked.

lib/krb5/expand_path.c
```diff
@@ -449,11 +476,20 @@ _expand_token(krb5_context context,
...
-    }
+    colon = strchr(token+2, ':');
+    if (!colon)
+	colon = token_end;
+
+    for (i = 0; i < sizeof(tokens)/sizeof(tokens[0]); i++)
+	if (!strncmp(token+2, tokens[i].tok, (colon - token) - 2)) {
+	    char *arg = NULL;
+	    if (*colon == ':')
+		asprintf(&arg, "%.*s", (int)(token_end - colon - 1), colon + 1);
```
vdukhovni: Should we check that `asprintf` succeeded?

lib/krb5/expand_path.c
```diff
@@ -449,11 +476,20 @@ _expand_token(krb5_context context,
 	    return _expand_extra_token(context, p[1], ret);
     }
 
-    for (i = 0; i < sizeof(tokens)/sizeof(tokens[0]); i++) {
-	if (!strncmp(token+2, tokens[i].tok, (token_end - token) - 2))
-	    return tokens[i].exp_func(context, tokens[i].param,
-				      tokens[i].postfix, ret);
-    }
+    colon = strchr(token+2, ':');
```
vdukhovni: This might run past `token_end`,  in which case we should again use `token_end`.

lib/krb5/expand_path.c
elric1: Well, that raises an interesting point. strftime(3) returns only a length, there are no errors. I was operating under the assumption that it was acceptable to accept a truncated result and therefore we do not need to check the error because, IIRC, strftime will nul-terminate the truncated string. We could, however, log an error and bail---but that raises the question: where would the log go if we can't open the log file? So, there's a few choices here.

lib/krb5/expand_path.c
elric1: We "do". That is, we pass NULL to the function which is what it would expect if no :<stuff> were specified. Like the above comment, there's a question about what the correct behaviour is on error. I was assuming that we can just proceed---but it's not optimal. It's not brilliant to just fail, either. So, we have another choice to make.

lib/krb5/expand_path.c
elric1: Yeah, this is actually worse because I really should not use strchr(3) as I have no guarantee that I'll find either a nul or a colon before I SEGV. Please find the updated open coded strchr(3)-like code.

lib/krb5/expand_path.c
vdukhovni: The FreeBSD manpage suggests that strftime returns zero on error, and then buffer is indeterminate. There's never truncation.  If we get 0, then we can't use the buffer...
```
     No more than maxsize characters will be placed into the array.  If the
     total number of resulting characters, including the terminating NUL
     character, is not more than maxsize, strftime() returns the number of
     characters in the array, not counting the terminating NUL.  Otherwise,
     zero is returned and the buffer contents are indeterminate.
```
I would naively also expect zero for invalid conversion specifiers.

lib/krb5/expand_path.c
vdukhovni: But a test with `date +"%f%J%i" ` shows that invalid conversions just emit the letter in question and drop the `%`, so I get `fJi` for the above...

lib/krb5/expand_path.c
vdukhovni: But a test with `date +"%f%J%i"` shows that invalid conversions just emit the letter in question and drop the `%`, so I get `fJi` for the above...

lib/krb5/expand_path.c
vdukhovni: For path expansions, I'd like to fail over using some path other than intended... Which may mean we can't log, but so be it?

elric1: Yes, any decision that we make here is suboptimal but I agree that it is better to fail to log than it is to log to the wrong file. I'll change this and the above to fail to log. In practice, this shouldn't really happen, but we should check.

Another question arises which is: should we try to ensure that the entire log path doesn't have a malloc(3) in it? This may be one of the cases where, perhaps, it would be better to use a fixed size buffer on the stack. After all, we do not want 10K single log messages in any case. Yes, that's the change that I'll make. Fixed size buffers.

elric1: Okay, the rest of the token expansion code doesn't use fixed size buffers, so I'm not going to in this one. I do punt on strftime(3) errors being defined as either 0 or max. That said, there is an overlap between strftime(3) errors and valid but degenerate cases, but perhaps we must accept this.

lib/krb5/log.c
```diff
@@ -197,24 +196,56 @@ open_syslog(krb5_context context,
...
+	if (f->disp == FILEDISP_IFEXISTS) {
+	    /* Cache failure for 1s */
+	    gettimeofday(&tv, NULL);
+	    if (tv.tv_sec == f->tv.tv_sec)
+		return;
+	} else {
+	    flags |= O_CREAT;
+	}
+
+	_krb5_expand_path_tokens(context, f->filename, 1, &filename);
```
vdukhovni: Unchecked return value?

elric1:
I've pushed a new one. That was an uninitialised variable.

lib/krb5/expand_path.c
nicowilliams: Oddly, I've seen advice that the return value *and* the result of `asprintf()` needs to be checked.  I forget what old C libraries are perverse enough to make that needed.

nicowilliams:
I was thinking we should have a `%iso8601date` and `%iso8601datetime`.

PR #648, opened 2019-11-13
WinCNG reusable hash support

lhoward:
Untested

### DIFF
--- a/lib/hcrypto/evp-wincng.c
+++ b/lib/hcrypto/evp-wincng.c
@@ -33,11 +33,16 @@
 #include <config.h>
 #include <roken.h>
 #include <assert.h>
+#include <versionsupport.h>
 
 #include <evp.h>
 #include <evp-wincng.h>
 
 #include <bcrypt.h>
+
+#ifndef BCRYPT_HASH_REUSABLE_FLAG
+#define BCRYPT_HASH_REUSABLE_FLAG 0x00000020
+#endif
 
 /*
  * CNG cipher provider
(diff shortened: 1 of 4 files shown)


lhoward:
Actually the latter statement isn't quite true – `wincng_md_hash_cleanup()` does need to be called from `EVP_DigestInit_ex()` too, in the case where the underlying digest changes. It turns out we don't need to change the provider interface at all, we just need to avoid calling `BCryptCreateHash()` if the hash object already exists.

(Of course, I realized this after cleaning up the provider interface... oh well. I'll omit those changes for now.)

lhoward:
Question: given the version information is retrieved directly from the PEB, is it pointless caching the result of `IsWindows8OrGreater()`?

nicowilliams:
What is PEB?

Anyways, patch LGTM.  Jeff might want to review.

jaltman:
@nicowilliams  PEB is the Windows Process Environment Block 
https://docs.microsoft.com/en-us/windows/win32/api/winternl/ns-winternl-peb

lhoward:
I’ll see if I can conduct some performance tests today.

jaltman:
Given the number of statements that must be executed during the IsWindows8OrGreater() implementation I believe the caching is worthwhile.   Perhaps the caching should be within each of the IsWindowsXXX() functions so it benefits all callers.

jaltman:
I've successfully tested this change on  Windows 10 [Version 10.0.19033.1].   
I still need to test on  Windows 7 [Version 6.1.7601].

jaltman:
This change has been successfully tested on Windows 7 [Version 6.1.7601].

PR #651, opened 2019-11-19
Declare variables at the beginning of a block.

elric1:
Looks like this partially fixes the Appveyor build.

We now need to fix the building of the documentation. It's missing a Perl module at the moment, which is likely just a tweak to Appveyor @jaltman? Or alternately, we could change appveyor to pass `--disable-heimdal-documentation` to `./configure`. This should fix the build as the documentation is the last thing to build---and we are getting there.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -383,6 +383,7 @@ tags
 /lib/roken/strpftime-test
 /lib/roken/test-auxval
 /lib/roken/test-detach
+/lib/roken/test-getuserinfo
 /lib/roken/test-readenv
 /lib/roken/tsearch-test
 /lib/roken/vis.h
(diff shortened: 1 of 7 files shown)


elric1:
btw: I think that we should definitely do this because I did fix some issues with my last PR because they broke Windows. They would have broken UNIX as well---it just wasn't noticed. If we can keep the Appveyor build clean and error free, without taking unnecessary shortcuts, I think that this would help to improve our code quality and test coverage.

elric1:
Please note, I am _not_ suggesting that we disable all of these tests on Windows, rather I am blasting through characterising the work that we need to accomplish to get this build and test cycle to work from end to end. Each commit that disables a particular commit should be changed into a commit that fixes the test before we merge this PR.

elric1:
@viktor: I'll open issues for test-detach and test_cc on Windows. Then let's pull this for now to fix the build.

PR #656, opened 2019-11-21
GSS_KRB5_CRED_NO_TRANSIT_CHECK_X/GSS_KRB5_CRED_ITERATE_ACCEPTOR_KEYTAB_X

metze-samba:
Members of Active Directory domains should just rely on there
KDCs (domain controllers) to do SID-Filtering (and name checking)
on trust boundaries, I have verified this with a modified Samba KDC
and a Windows 2012R2 DC. The Windows DC rejects invalid cross-realm tickets
with KRB5KDC_ERR_POLICY, before generating a new (service or referral)
ticket. So any service ticket is already policy checked by the KDC
even if this does not result in setting the transited_policy_checked in the ticket.
    
This means an accepting service can tell gss_accept_sec_context()
to skip any transited checking, as the trust topoligy is only
fully known to the KDC anyway.
    
The detailed background for this can be found in the bug report
and the mailing list:
https://lists.samba.org/archive/samba-technical/2019-September/thread.html#134285
https://lists.samba.org/archive/samba-technical/2019-November/thread.html#134553
http://mailman.mit.edu/pipermail/krbdev/ should also have references.
    
BUG: https://bugzilla.samba.org/show_bug.cgi?id=12907


metze-samba:
What's the best way to bisect the test failures? Is there a way to a travis ci build without a pull request?

metze-samba:
It's failing with this:

AIL: check-context

===================

Creating database

Doing database check

Starting kdc

./check-context: line 176: 10513 Segmentation fault: 11  (core dumped) KRB5_CONFIG=${KDC_KRB5_CONFIG} ${kdc} --detach --testing

kdc failed to start

FAIL check-context (exit status: 1)

While is passes a local make check, there if fails check-iprop, even with plain master.

Any hints how to debug this?

iboukris:
@metze-samba, with this fixup check-context pass locally for me:

```diff
$ git diff
diff --git a/tests/gss/Makefile.am b/tests/gss/Makefile.am
index 213283270..86dc24132 100644
--- a/tests/gss/Makefile.am
+++ b/tests/gss/Makefile.am
@@ -85,4 +85,5 @@ EXTRA_DIST = \
        check-ntlm.in \
        check-context.in \
        ntlm-user-file.txt \
-       krb5.conf.in
+       krb5.conf.in \
+       krb5.conf.kdc.in
```

metze-samba:
Ok, it no longer segfaults, but we get this:

FAIL: check-context

===================

Creating database

Doing database check

Starting kdc

Test gss_acquire_cred_with_password

test_context: gss_acquire_cred_with_password: gss-code: 851968  Miscellaneous failure (see text) -- mech-code: 2529638918 Client (user1@TEST.H5L.SE) unknown

test failed

signal killing kdc

FAIL check-context (exit status: 1)

iboukris:
Ouch, that's the error I had before the fixup .. weird.

iboukris:
I was able to reproduce by doing the same, "mkdir ci-build; cd ci-build; ../configure ...", see demo fixup at: https://github.com/iboukris/heimdal/commit/3d04b05df25e443c09a5128eb4494eea328431b0

abartlet:
Beyond the WIP tag, what else is needed for this to be considered for merging?

This is a set of patches we might otherwise need to carry in Samba's lorikeet-heimdal and I'm just trying to get a sense as to their prospects as we try and wrangle our upgrade.

Thanks!

lib/krb5/rd_req.c
```diff
@@ -220,7 +221,20 @@ krb5_decrypt_ticket(krb5_context context,
...
+	    skip_transit_check = TRUE;
+	} else if(flags & KRB5_VERIFY_AP_REQ_SKIP_TRANSITED_CHECK) {
+	    skip_transit_check = TRUE;
+	} else {
+	    skip_transit_check = krb5_config_get_bool_default(context,
+							      NULL,
+							      FALSE,
+							      "libdefaults",
+							      "acceptor_skip_transit_check",
+							       NULL);
```
lhoward, line 234: extra space here

lhoward: acceptor is a GSS-ism so, probably this can be renamed to something that uses Kerberos terminology given it lives in libkrb5.

lib/krb5/rd_req.c
```diff
@@ -554,6 +570,36 @@ krb5_rd_req_in_set_keytab(krb5_context context,
     return 0;
 }
 
+KRB5_LIB_FUNCTION krb5_error_code KRB5_LIB_CALL
+krb5_rd_req_in_set_verify_ap_req_flags(krb5_context context,
+			     krb5_rd_req_in_ctx in,
```
lhoward, line 575: indentation (unless it's GitHub doing it)

lhoward:
This looks good to me, just fix a few indentation things and think about whether the libdefaults key should not use the word acceptor.

lib/krb5/rd_req.c
metze-samba: @lhoward first thanks very much for the review!

As I'm aiming to have things upstreamed in Heimdal and MIT,
lets discuss the option name in https://github.com/krb5/krb5/pull/1005
there I'm waiting for comments from @greghudson

jaltman:
I am opposed to the creating of a krb5.conf setting to disable transit checks in cases where the KDC has refused to set the "transited_policy_checked" flag in the service ticket.

metze-samba:
Heimdal has the check_pac option on by default, we could just a verified PAC as indication to skip the transited check,
that would not need any additional options. I'll just add a GSS_KRB5_CRED_REQUIRE_PAC_X option
to force krb5_rd_req_in_set_pac_check(TRUE), even if the krb5.conf sets check_pac = false.

I think that is simpler and avoids admin interaction completely.

What do you think?

nicowilliams:
> The Windows DC rejects invalid cross-realm tickets
with KRB5KDC_ERR_POLICY, before generating a new (service or referral)
ticket. So any service ticket is already policy checked by the KDC
even if this does not result in setting the transited_policy_checked in the ticket.

RFC4120 provides a ticket flag for this: `TRANSITED-POLICY-CHECKED`.  Do AD KDCs not set this ticket flag??

It is always possible to configure every possible path as trusted, therefore an option to disable transit path checking is clearly fine.  My preference would be that we have a GSS name attribute for conveying transit path, but otherwise an API option is OK.

Keep in mind that Heimdal does support automatic hierarchical trust path including short-circuit paths.

nicowilliams:
Do AD KDCs not check the client name and realm?

Consider

- a service in realm A that doesn't use PACs, or which uses both, the SIDs in the PAC and _also_ the `cname@CREALM`, and
- suppose A trusts some realm B, and
- also C and other realms but not transiting through B, and
- B is compromised and issues a ticket for any user `@B`, `@C`, etc., maybe even `@A` (though this last probably would trip loop detection, right?!)...

Will... the KDCs for A issue tickets to a possessor of a `krbtgt/A@B` with the claimed `cname@C` (etc) but with correct SIDs, or will it refuse to issue the ticket?

lhoward:
[MS-KILE] 3.1.5.4 says:

> The TRANSITED-POLICY-CHECKED flag ([RFC4120] section 2.7): KILE does not check for transited domains on servers or a KDC. Application servers MUST ignore the TRANSITED-POLICY-CHECKED flag. For details on decoding a cross-realm TGT and crealm filtering see [MS-PAC] section 4.1.2.3.

nicowilliams:
So, we have a lot of questions.  Yes, we're looking at docs like https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-adts/e9a2d23c-c31e-4a6f-88a0-6646fdb51a3c and still we have questions.

Some of mine are above.

lhoward:
Pasting useful info from [MS-PAC].

>4.1.2.3 crealm Filtering
When decoding a cross-realm TGT, the crealm fields inside the TGT are compared to the expected name of the realm for the interrealm trust. If the names do not match the TGT, they are rejected, subject to other mitigating constraints.<32>
These constraints can include allowing fully trusted domains to supply any crealm name on the basis that it would have validated it prior to passing it along, or any other settings that are established out of band. The full set of constraints is implementation-specific.

>The TGT's crealm field is compared against the realm names listed on the TDO, as specified in [MS-ADTS], corresponding to the cross-realm trust. If there is a mismatch, the TGT is rejected. TDOs marked as within the forest pass all crealm names through. TDOs marked as forest transitive indicate that the server will only accept crealm names if it is a name claimed by the forest on the TDO. If the TDO used for the cross-realm TGT has neither indicator set, the server checks if the fully qualified domain name (FQDN) (2) matches the FQDN (2) of any domain in the server's forest; if so, the TGT is accepted. Finally, if the crealm field matches the FQDN (2) of the TDO, then it is accepted.

PR #658, opened 2019-11-26
Improve encoding of krb5_{,un}parse_principal().

elric1:
By default, we encode all unprintable characaters as well as the
set specified in RFC1964's section on export names.  We use unvis(3)
when parsing but open-code the unparser.  unvis(3) needed to be
slightly corrected with a macro and svis(3) would have required a
much less pleasant macro.  The out text format is almost compatible
with unvis(3) but spaces are quoted as "\ " rather than "\s".  This
is specified in RFC 1964.

We define a new constant KRB5_PRINCIPAL_UNPARSE_EXPORT_NAME which
if passed in as a flag will preserve the existing behaviour.

### DIFF
--- a/lib/krb5/principal.c
+++ b/lib/krb5/principal.c
@@ -55,6 +55,7 @@ host/admin@H5L.ORG
 #include <arpa/nameser.h>
 #endif
 #include <fnmatch.h>
+#include <vis.h>
 #include "resolve.h"
 
 #define princ_num_comp(P) ((P)->name.name_string.len)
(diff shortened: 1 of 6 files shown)


vdukhovni:
Is this PR still something we need to do?
cc: @nicowilliams

elric1:
I think that we should consider it. We currently don't quote many characters which may spam logs with unprintables, IIRC, which is what we're trying to fix here.

lib/krb5/principal.c
```diff
@@ -205,200 +206,225 @@ krb5_principal_get_num_comp(krb5_context context,
...
-		krb5_set_error_message(context, ret,
-				       N_("part after realm in principal name", ""));
-		goto exit;
-	    } else {
-		comp[n] = malloc(q - start + 1);
-		if (comp[n] == NULL) {
-		    ret = krb5_enomem(context);
-		    goto exit;
+
+    for (pos=in, i=0, j=0; in[i]; i++) {
```
lhoward, line 345: whitespace around =

lib/krb5/principal.c
```diff
@@ -205,200 +206,225 @@ krb5_principal_get_num_comp(krb5_context context,
...
-    free(s);
-    return 0;
-exit:
-    while (n>0) {
-	free(comp[--n]);
+	set_default_princ_type(*out, KRB5_NT_PRINCIPAL);
+
+out:
+    if (ret) {
+	for (i=0; i < princ_num_comp(*out); i++)
```
lhoward, line 417: ditto

lib/krb5/principal.c
```diff
@@ -205,200 +206,225 @@ krb5_principal_get_num_comp(krb5_context context,
...
+	    free((*out)->name.name_string.val[i]);
+	free((*out)->name.name_string.val);
+	free((*out)->realm);
+	free(*out);
```
lhoward, line 421: is there a reason we can't use `krb5_free_principal()` here?

lhoward (changes requested):

nicowilliams:
What actually is changing besides using vis/unvis?  Are there bugs in Heimdal's RFC 1964 principal name parsing/unparsing?

elric1:
I may have gone a little overboard on this, it was a long time ago that I wrote it. But Heimdal's 1964 {,un}parse are broken and not reversible. When I went to fix it, I was interested in making the logging not nec. 1964 compliant but safe and so I wanted to use vis/unvis to escape more control characters. I think that I used vis/unvis for compat with external libraries and because people are used to it and can read it.

nicowilliams:
I'd rather not have a vis/unvis depedency here, especially since it's a libroken special and if we might diverge from other implementations.  vis/unvis is fine for logging, but not for RFC 1964 principal parsing/unparsing -- but I didn't quite undertand if you're only using it for logging.

Do you have an example of a principal name that Heimdal's RFC 1964 codec is broken for?

PR #665, opened 2019-12-10
Update bx509d to log the new way

nicowilliams:


kdc/bx509d.c
```diff
@@ -423,18 +416,20 @@ bad_reqv(struct MHD_Connection *connection,
...
-
+    _kdc_audit_addkv((kdc_request_t)r, 0, "http-status-code", "%d",
+                     http_status_code);
+    (void) gettimeofday(&r->tv_end, NULL);
     if (code == ENOMEM) {
-        if (context)
-            kdc_log(context, kdc_config, 4, "Out of memory");
-        return resp(connection, http_status_code, MHD_RESPMEM_PERSISTENT,
+        if (r->context)
+            kdc_log(r->context, kdc_config, 3, "Out of memory");
```
vdukhovni: If this is not debug logging, "our fault" errors get logged at level 1, garbage from clients at level 2, routine audit trail at level 3. So the log level should probably be 1.

kdc/bx509d.c
```diff
@@ -447,105 +442,119 @@ bad_reqv(struct MHD_Connection *connection,
...
     }
+    _kdc_audit_addkv((kdc_request_t)r, KDC_AUDIT_VIS, "e-text", "%s", formatted);
+    _kdc_audit_trail((kdc_request_t)r, code);
     krb5_free_error_message(context, k5msg);
 
     if (ret == -1 || msg == NULL) {
         if (context)
-            kdc_log(context, kdc_config, 4, "Out of memory");
-        return resp(connection, MHD_HTTP_SERVICE_UNAVAILABLE,
+            kdc_log(r->context, kdc_config, 3, "Out of memory");
```
vdukhovni: Ditto.

kdc/bx509d.c
```diff
@@ -698,98 +707,169 @@ do_CA(krb5_context context,
...
+            (token_end - token) > INT_MAX || (token_end - token) < 2)
+            _kdc_audit_addkv((kdc_request_t)r, 0, "auth", "<unknown>");
+        else
+            _kdc_audit_addkv((kdc_request_t)r, 0, "auth", "%.*s",
+                             (int)(token_end - token), token);
+
+    }
+
+    if (ret == 0 && r->kv == NULL) {
+        kdc_log(r->context, kdc_config, 3, "Out of memory");
```
vdukhovni: Ditto.

kdc/bx509d.c
```diff
@@ -698,98 +707,169 @@ do_CA(krb5_context context,
...
-    kdc_log(context, kdc_config, 4, "Issued certificate to %s",
-            cprinc_from_token);
-    ret = good_bx509(connection, pkix_store);
-
-    if (pkix_store)
-        (void) unlink(strchr(pkix_store, ':') + 1);
-    free(cprinc_from_token);
-    free(pkix_store);
-    return ret == -1 ? MHD_NO : MHD_YES;
+    kdc_log(r->context, kdc_config, 3, "Issued certificate to %s", r->cname);
```
vdukhovni: This should be level 4.

kdc/process.c
```diff
@@ -202,9 +203,11 @@ _kdc_audit_trail(kdc_request_t r, krb5_error_code ret)
     }
     kvbuf[j] = '\0';
 
-    kdc_log(r->context, r->config, 3, "%s %s %s %s %s%s",
-	    r->reqtype, retval, r->from, r->cname,
-	    r->sname, kvbuf);
+    kdc_log(r->context, r->config, 3, "%s %s %d %s %s %s%s",
+	    r->reqtype, retval, ret, r->from,
```
vdukhovni: I thought you would only add the integer ret when we don't have a known string...

kdc/bx509d.c
nicowilliams: Got it.

kdc/process.c
nicowilliams: Will do.

PR #668, opened 2019-12-26
NegoEx integration branch

lhoward:
Mechanism glue and SPNEGO fixes; NegoEx implementation.

NegoEx is a pseudo-mechanism that is negotiated by SPNEGO, which has simpler and well defined negotiation protection semantics, has considerably improved interoperability with Windows SSPs, and most importantly permits the exchange of arbitrary metadata before committing to an authentication mechanism. This is very useful, for example, for discovering trust anchors. The NegoEx implementation was initially written by me for MIT Kerberos back into 2011, however I only re-licensed it under a BSD/MIT license a year ago, and it has only recently been merged. @greghudson simplified and rewrote much of it, replacing casting-on-the-wire with marshalling code and catching several edge cases. The implementation here is the MIT one, ported to Heimdal.
 
Sundry fixes:

- Some NTLM fixes to pass tests and support resetting crypto state in SPNEGO
- Make mechanism entry points that are optional truly optional (do not require a symbol)
- Use TAILQ instead of SLIST in mechglue so that mechanisms are ordered correctly

The following Apple patches are ingested:

- SPNEGO implementation
- enhanced mechglue logging
- channel binding helpers
- gss_cred_hold()
- gss_iter_cred()
- gss_cred_label()
- buffer, name and credential helper APIs
- mechglue plugins for credential replacement

Not all of the Apple changes are necessary but if we are ever to synchronise with their krb5 and ntlm mechanisms, they will be useful. On the other hand, as many of the helpers (e.g. the channel binding APIs) have no existing consumer outside their mechanism implementations, we might avoid them unless we plan to eventually merge Apple's krb5 mechanism.

The commits are ordered such that it should be possible to drop those corresponding to undesired functionality. Note however that new mechanisms may take advantage of some of these helpers, e.g. I have used some in SAnon.


lhoward:
Reviewed by @nicowilliams

PR #669, opened 2019-12-28
TAILQ in mechglue

lhoward:
Use TAILQ instead of SLIST in mechanism glue to preserve ordering of mechanism configuration file and desired mechanism OID sets. Otherwise the order was being inverted by virtue of entries being added to the head of the list.


lhoward:
Merged in NegoEx patch.

PR #670, opened 2019-12-30
Implementation of draft-howard-gss-sanon

lhoward:
Add support for SAnon, a simple key agreement protocol that provides no authentication of initiator or acceptor using x25519 ECDH key exchange. More information can be found in the accompanying [Internet Draft](https://datatracker.ietf.org/doc/html/draft-howard-gss-sanon).


nicowilliams:
Can we move the ed25519 implementation to `lib/hcrypto`?  I need it for the JWT code...  I can always move it myself later, so feel free not to.

lib/gssapi/mech/gss_mech_switch.c
```diff
@@ -462,6 +460,9 @@ _gss_load_mech(void)
 	}
 	fclose(fp);
 #endif
+
+out:
+	add_builtin(__gss_sanon_initialize());
```
nicowilliams, line 465: Nice way to make it last.  I'm thinking though that eventually it'd be nice to be able to have the gss config specify order of all mechanisms, even builtin ones, and then we'll need to do something about adding this builtin earlier, maybe.

lib/gssapi/mech/gss_oid.c
```diff
@@ -151,6 +151,9 @@ gss_OID_desc GSSAPI_LIB_VARIABLE __gss_c_ntlm_reset_crypto_oid_desc = { 11, rk_U
 /* GSS_NEGOEX_MECHANISM - 1.3.6.1.4.1.311.2.2.30 */
 gss_OID_desc GSSAPI_LIB_VARIABLE __gss_negoex_mechanism_oid_desc = { 10, rk_UNCONST("\x2b\x06\x01\x04\x01\x82\x37\x02\x02\x1e") };
 
+/* GSS_SANON_X25519_MECHANISM - 1.3.6.1.4.1.5322.26.1.110 */
+gss_OID_desc GSSAPI_LIB_VARIABLE __gss_sanon_x25519_mechanism_oid_desc = { 10, rk_UNCONST("\x2b\x06\x01\x04\x01\xa9\x4a\x1a\x01\x6e") };
```
nicowilliams, line 155: This reminds me: we need some tool to test that all these hard-coded DER OID byte strings are correct.  I'm not sure how to build such a tool.  Perhaps we could have a text file with `varname OID` with the OID in dotted form, and then "compile" it to a `.c` file, then we'd not need any test of correctness here.

nicowilliams: Oh, what are the OID arc names?

nicowilliams: Ahh: `iso(1) org(3) dod(6) internet(1) private(4) enterprise(1) padl(5322)`.

nicowilliams: Oh, never mind.  This _is_ generated, but also committed.

lib/gssapi/sanon/accept_sec_context.c
```diff
@@ -0,0 +1,149 @@
...
+    major = GSS_S_COMPLETE;
+
+    *context_handle = (gss_ctx_id_t)sc;
+
+    if (src_name)
+	*src_name = _gss_sanon_wellknown_name;
+    if (ret_flags)
+	*ret_flags = flags;
+    if (time_rec)
+	_gss_sanon_lifetime_left(&tmp, 0, time_rec);
```
nicowilliams, line 133: Just... set it to `GSS_C_INDEFINITE`.

lib/gssapi/sanon/acquire_cred.c
```diff
@@ -0,0 +1,62 @@
...
+{
+    if (!_gss_sanon_is_valid_name_p(desired_name)) {
+	*minor = EINVAL;
+	return GSS_S_BAD_NAME;
+    }
+
+    *output_cred_handle = _gss_sanon_default_cred;
+
+    if (time_rec)
+	_gss_sanon_lifetime_left(minor, 0, time_rec);
```
nicowilliams, line 57: Ditto.

lib/gssapi/sanon/add_cred.c
```diff
@@ -0,0 +1,70 @@
...
+    if (input_cred_handle != GSS_C_NO_CREDENTIAL &&
+	input_cred_handle != _gss_sanon_default_cred) {
+	*minor = EINVAL;
+	return GSS_S_NO_CRED;
+    }
+
+    *output_cred_handle = _gss_sanon_default_cred;
+
+    if (initiator_time_rec)
+	_gss_sanon_lifetime_left(minor, 0, initiator_time_rec);
```
nicowilliams, line 63: Ditto.

lib/gssapi/sanon/context_time.c
```diff
@@ -0,0 +1,96 @@
...
+    krb5_context context;
+
+    GSSAPI_KRB5_INIT (&context);
+
+    if (endtime == 0) {
+	*lifetime_rec = GSS_C_INDEFINITE;
+	return GSS_S_COMPLETE;
+    }
+
+    kret = krb5_timeofday(context, &now);
```
nicowilliams, line 52: There's no point to using `krb5_timeofday()`, I don't think.  After all, we don't have a ccache where a KDC offset might be found, we're not opening the ccache, we're not talking to a KDC.  So just use `time()` or `gettimeofday()`.

nicowilliams: Then you won't need to `GSSAPI_KRB5_INIT (&context)` here either.

lib/gssapi/sanon/context_time.c
```diff
@@ -0,0 +1,96 @@
...
+    gsskrb5_ctx rfc3961 = (gsskrb5_ctx)sc->rfc3961;
+
+    if (rfc3961 == NULL)
+	return GSS_S_NO_CONTEXT;
+
+    HEIMDAL_MUTEX_lock(&rfc3961->ctx_id_mutex);
+    endtime = rfc3961->endtime;
+    HEIMDAL_MUTEX_unlock(&rfc3961->ctx_id_mutex);
+
+    major_status = _gss_sanon_lifetime_left(minor, endtime, time_rec);
```
nicowilliams, line 85: When would the context time not be `GSS_C_INDEFINITE` anyways?  When the credential was acquire with some specific `time_req`?  But you can get more time...  I'd say simplify and always indicate `GSS_C_INDEFINITE`.

lib/gssapi/sanon/accept_sec_context.c
```diff
@@ -0,0 +1,149 @@
...
+{
+    static gss_buffer_desc empty = GSS_C_EMPTY_BUFFER;
+    OM_uint32 major, tmp;
+    sanon_ctx sc = (sanon_ctx)*context_handle;
+    gss_buffer_desc mech_input_token = GSS_C_EMPTY_BUFFER;
+    gss_buffer_desc hok_mic = GSS_C_EMPTY_BUFFER;
+    OM_uint32 flags;
+
+    /* if verifier_cred_handle is not one of below, we have a mechglue problem */
+    heim_assert(verifier_cred_handle == GSS_C_NO_CREDENTIAL ||
```
nicowilliams, line 55: You don't always assert in other cases.

lib/gssapi/sanon/sanon_locl.h
```diff
@@ -0,0 +1,70 @@
...
+
+#define SANON_FLAG_LOCAL 1
+#define SANON_FLAG_COMPLETE 2
+
+typedef struct sanon_ctx_desc {
+    uint8_t sk[crypto_scalarmult_curve25519_BYTES];
+    uint8_t pk[crypto_scalarmult_curve25519_BYTES];
+    uint32_t flags;
+    krb5_keyblock session_key;
+    krb5_auth_context_data auth_context; /* need to make krb5 mech happy */
```
nicowilliams, line 55: You might need this to be a `krb5_auth_context` rather than a `krb5_auth_context_data` so you can `krb5_auth_con_free()` it in case the krb5 mech and library allocate more stuff into it.  Have you confirmed there's no leaks?

nicowilliams: BTW, we now have `gss-token` in-tree.  It might be useful for testing.

lhoward:
> Can we move the ed25519 implementation to `lib/hcrypto`? I need it for the JWT code... I can always move it myself later, so feel free not to.

Can do a simple move, but then also thought – are we then obligated to go down the rabbit warren of matching the OpenSSL APIs for this?

lib/gssapi/mech/gss_mech_switch.c
lhoward: This is also arguably less necessary now we have mechanism order sorting.

lib/gssapi/sanon/sanon_locl.h
lhoward: Only by inspection. SAnon only uses the functionality in `krb5/cfx.c`, which calls:

- `krb5_auth_con_getlocalseqnumber()`
- `krb5_auth_con_setlocalseqnumber()`
- `krb5_auth_con_getlocalsubkey()`
- `krb5_auth_con_getremotesubkey()`
- `krb5_auth_con_getkey()`

none of which allocate anything in `auth_context`.

nicowilliams:
> >  Can we move the ed25519 implementation to lib/hcrypto? I need it for the JWT code... I can always move it myself later, so feel free not to.

> Can do a simple move, but then also thought – are we then obligated to go down the rabbit warren of matching the OpenSSL APIs for this?

Nothing would so obligate us.  It'd be nice to have OpenSSL-equivalent PK APIs, but that's a larger project.

lib/gssapi/sanon/display_status.c
```diff
@@ -0,0 +1,85 @@
...
+			  int status_type,
+			  const gss_OID mech_type,
+			  OM_uint32 *message_context,
+			  gss_buffer_t status_string)
+{
+    krb5_context context;
+    char *buf = NULL;
+    int e = 0;
+
+    GSSAPI_KRB5_INIT (&context);
```
nicowilliams, line 48: FYI, I checked all the `*minor_status` values you're setting, and basically none will be anything other than `EINVAL` or `ENOMEM`, even the ones from various `krb5` functions because those either can't fail as called or can only fail due to `ENOMEM`.

lib/gssapi/sanon/sanon_locl.h
nicowilliams: OK, thanks.

nicowilliams:
Maybe we could tweak the krb5 exported interprocess context token format somewhat to make it easy to create one here and then import it with `gss_import_sec_context()`, then use a proper `gss_ctx_id_t` instead of an `auth_context`.  That seems like a cleaner interface, and then we'll have no need whatsoever for a `krb5_context` in SAnon.

lib/gssapi/sanon/canonicalize_name.c
```diff
@@ -0,0 +1,52 @@
...
+			     const gss_OID mech_type,
+			     gss_name_t *dest_name)
+{
+    *minor = 0;
+
+    if (src_name == GSS_C_NO_NAME)
+	return GSS_S_BAD_NAME;
+
+    if (_gss_sanon_is_anonymous_name_p(src_name))
+	*dest_name = _gss_sanon_wellknown_anon_name();
```
nicowilliams, line 46: Is there any reason not to just always do this?

An SAnon mech name (MN) should probably always have the same anon form.  That is, a `gss_name_t` obtained from `gss_accept_sec_context()`, `gss_inquire_cred*()`, `gss_inquire_context()`, or `gss_canonicalize_name()` should probably all have exactly the same display form and name-type.

lhoward: I might be missing something but, if you call `gss_canonicalize_name()` on a non-anonymous name, and then pass that to `gss_init_sec_context()`, you may get anonymous authentication when you didn't want it. See the [draft](https://datatracker.ietf.org/doc/html/draft-howard-gss-sanon#section-4.1).

lib/gssapi/sanon/display_status.c
lhoward: I always get confused by `gss_display_status()`. What does the above mean, that we can eliminate it, or that we don't need `krb5_get_error_message()` and can just call `strerror()` instead?

lhoward:
> Maybe we could tweak the krb5 exported interprocess context token format somewhat to make it easy to create one here and then import it with `gss_import_sec_context()`, then use a proper `gss_ctx_id_t` instead of an `auth_context`. That seems like a cleaner interface, and then we'll have no need whatsoever for a `krb5_context` in SAnon.

I was going to suggest a `gss_krb5_lucid_context_version_t` as used by NFSv4 but that is export-only. I agree it's a cleaner interface. Let me look at the implementation cost.

nicowilliams:
I thought about the lucid thing, but you'll note we don't have the import side for that...

lib/gssapi/sanon/canonicalize_name.c
nicowilliams: You also explained on IM.  I'll resolve this now then.

lhoward:
> I thought about the lucid thing, but you'll note we don't have the import side for that...

Implemented!

lhoward:
I also added a bit of a hack for Windows interoperability. I didn't want to change the protocol so, in order to convey Windows-specific feature requests (e.g. `GSS_C_DCE_STYLE`), we stuff those flags in the NegoEx metadata. Here's the new draft text:

```
   The initiator metadata, if present, contains a set of GSS-API flags
   encoded as a 4 byte little endian integer.  This is used to convey to
   the acceptor any Windows-specific GSS-API flags (see [RFC4757]
   Section 7.1).  Other GSS-API flags MUST NOT be present in the
   metadata.
```

lhoward:
So, `gss_set_sec_context_option(GSS_KRB5_IMPORT_RFC4121_CONTEXT_X)` allows you to get back a `gss_ctx_id_t` that you can use for RFC4121 message protection/PRF services, given an arbitrary session key. It's useful for composing with mechanisms such as SAnon.

The input buffer is formatted as follows:

```
uint8_t is_initiator;
uint32_t gss_flags;
int32_t enctype;
uint8_t session_key[0];
```

The session key length is determined from the encryption type.

lhoward:
OK, I think the code is in pretty good shape, assuming no major changes after IETF review of the protocol spec. The things that are a bit controversial protocol wise are:

- allowing SAnon to be requested without GSS_C_ANON_FLAG by using an anonymous initiator credential or target name
- allowing well known anonymous name in GSS_C_NT_USER_NAME as equivalent to GSS_C_NT_ANONYMOUS
- RFC4757 extended GSS flags support for Windows interoperability (these are tunnelled in NegoEx metadata rather than being part of the protocol proper)

I believe there is merit in these.

nicowilliams:
> > I thought about the lucid thing, but you'll note we don't have the import side for that...

> Implemented!

Excellent, now we can test that too.

I had no idea that `gss_set_sec_context_option()` could allocate a context!  Nice!

lhoward:
> > > I thought about the lucid thing, but you'll note we don't have the import side for that...
> 
> > Implemented!
> 
> Excellent, now we can test that too.
> 
> I had no idea that `gss_set_sec_context_option()` could allocate a context! Nice!

You can thank draft-engert-ggf-gss-extensions-00.txt for that ;)

lhoward:
BTW, that's the one thing I'm a bit iffy about – `gss_set_sec_context_option()` could be another way to handle `GSS_C_DCE_STYLE` by having the acceptor create a context for use by DCE RPC – then, we wouldn't have to send the flag from the initiator. SSPI does support this. But, it would mean changing the existing DCE RPC GSS-API implementation.

nicowilliams:
Merged via 20f9b2be4..5057d04f6 (beda11a07, 4a7eb7437, 5057d04f6).  I split out the hcrypto X25519 bits into a separate commit, and fixed up some things while chatting with @lhoward (who reviewed the changes).

PR #671, opened 2020-01-10
Fix some issues found by coverity

dariaphoebe:
The AuriStor coverity instance hasn't been able to build for a while; These fixes are for some new issues which have since appeared.

### DIFF
--- a/kadmin/load.c
+++ b/kadmin/load.c
@@ -332,7 +332,7 @@ parse_extensions(char **strp, HDB_extensions **e)
 	void *d;
 
 	len = strlen(p);
-	d = emalloc(len);
+	d = emalloc(len+1);
 
 	len = hex_decode(p, d, len);
 	if (len < 0) {
--- a/kdc/kx509.c
+++ b/kdc/kx509.c
@@ -757,8 +757,6 @@ check_authz(krb5_context context,
             break;
 
         if ((ret = der_parse_heim_oid(s, ".", &oid))) {
-            free(cprinc);
-            free(s);
             goto out;
         }
         for (k = 0; k < sizeof(eku_whitelist)/sizeof(eku_whitelist[0]); k++) {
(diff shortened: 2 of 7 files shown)


dariaphoebe:
poke poke. merge this.

kadmin/load.c
```diff
@@ -332,7 +332,7 @@ parse_extensions(char **strp, HDB_extensions **e)
 	void *d;
 
 	len = strlen(p);
-	d = emalloc(len);
+	d = emalloc(len+1);
```
lhoward: Are you sure this is right?

dariaphoebe: Coverity's position is 
"
   10. strlen_assign: Setting variable len to the return value of strlen called with argument p.
334        len = strlen(p);
   CID 12491 (#1 of 1): Out-of-bounds access (OVERRUN)11. alloc_strlen: Allocating insufficient memory for the terminating null of the string.
335        d = emalloc(len);
336"

I don't see what it's complaining about, and seems like it's a false positive.

SethEllsworth: Need to add 1 for the null terminator. A char string of 'len' is 4 bytes, because it's 'l e n \0'

But strlen( 'len' ) just returns 3. So making a place to copy it to needs to be 1 larger for the \0. Otherwise it can lead to overread or out of bounds access.

Sorry for sounding like a 'C' class teacher. Got bit by this last month.

lhoward: Your observation would be correct for `hex_encode()`. 

However `hex_decode()` is decoding binary information from a `NUL` terminated hex string. It returns binary data and a length. It should not need to be terminated.

lhoward: `hex_decode()` should really be called with `len / 2`.

SethEllsworth: My apologies. Just like me, Coverity didn't look into the decode function to realize the resultant usage of the allocated memory didn't need to be null terminated.

dariaphoebe:
closing, revised PR in #679

PR #673, opened 2020-01-29
Windows: Fix type cast warning.

rodwiddowson:
Since at least SDK V6.1 HCRYPTPROV has been specified as ULONG_PTR
this means that comparing one with NULL causes a cast warning.

Fix this by introducing an explicit NULL_HCRYPTPROV.

### DIFF
--- a/lib/hcrypto/rand-w32.c
+++ b/lib/hcrypto/rand-w32.c
@@ -47,9 +47,9 @@ static HCRYPTPROV
 _hc_CryptProvider(void)
 {
     BOOL rv;
-    HCRYPTPROV cryptprovider = (HCRYPTPROV)NULL;
+    HCRYPTPROV cryptprovider = (HCRYPTPROV)0;
 
-    if (g_cryptprovider != (HCRYPTPROV)NULL)
+    if (g_cryptprovider)
 	goto out;
 
     rv = CryptAcquireContext(&cryptprovider, NULL,
--- a/lib/hcrypto/rand-w32.c
+++ b/lib/hcrypto/rand-w32.c
@@ -41,15 +41,15 @@
 
 #include "randi.h"
 
-volatile static HCRYPTPROV g_cryptprovider = NULL;
+volatile static HCRYPTPROV g_cryptprovider;
 
 static HCRYPTPROV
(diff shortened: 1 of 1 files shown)


lib/hcrypto/rand-w32.c
```diff
@@ -41,15 +41,16 @@
 
 #include "randi.h"
 
-volatile static HCRYPTPROV g_cryptprovider = NULL;
+#define NULL_HCRYPTPROV ((ULONG_PTR)0)
+volatile static HCRYPTPROV g_cryptprovider = NULL_HCRYPTPROV;
```
lhoward: I'm not too fussed but, I'd probably prefer not defining something that looks like it could come from the Windows headers. Just use ```(HCRYPTPROV)0``` explicitly.

jaltman: This is a global variable.   It doesn't require initialization.   It will automatically be initialized to a zero value.   Changing it to be

  volatile static HCRYPTPROV g_cryptprovider;

is sufficient.

lhoward: True. My comment applies to line 51.

jaltman: I agree.

rodwiddowson:
Just so's you know...

I also toyed with just saying 0 (ungarnished), but I like the casting since that makes is clear we *are* dealing with a type.  

Also since it was H (for handle) I liked NULL not 0 (and don't even get me started on 32 bit handles and 64 bit handles).

As always, stylistic comments on submissions welcome.  I borrow your product so the least I can do is to make your life easy.

lhoward:
Ah yes, all good points @rodwiddowson! I agree it's down to aesthetics, myself I prefer to use 0 for `ULONG_PTR` and `uintptr_t`, and NULL for `PVOID` and `void *`. To me a handle is logically closer to a file descriptor than a pointer, even if it may actually be a pointer.

lhoward:
I will merge tomorrow when I'm back at my development machine.

rodwiddowson:
@lhoward Noted.

lhoward:
Merged into master

PR #675, opened 2020-02-10
7-1-branch: tgs-req: strip forwardable and proxiable if the server is disallowed

iboukris:


iboukris:
The CI failure seem unrelated.

PR #676, opened 2020-02-17
Add in the ability to query the cert's RFC822 Name.

JStewart-QSFT:
Also add in the ability to find any heim_ia5_string fields in the certificate's subjectAlternativeName.

### DIFF
--- a/lib/hx509/libhx509-exports.def
+++ b/lib/hx509/libhx509-exports.def
@@ -94,6 +94,8 @@ EXPORTS
 	hx509_cert_check_eku
 	hx509_cert_cmp
 	hx509_cert_find_subjectAltName_otherName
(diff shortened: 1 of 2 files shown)


lib/hx509/libhx509-exports.def
```diff
@@ -94,6 +94,8 @@ EXPORTS
 	hx509_cert_check_eku
 	hx509_cert_cmp
 	hx509_cert_find_subjectAltName_otherName
+	hx509_cert_find_subjectAltName_heim_ia5_string
```
nicowilliams: You'd also have to add these to `version-script.map`...

nicowilliams:
Where will these be used?  There's already a SAN iterator for CSRs, `hx509_request_get_san()` -- maybe we need a certificate SAN iterator too.

PR #678, opened 2020-03-01
Move some lib/krb5/ infrastructure bits to lib/base/

nicowilliams:
This PR moves configuration file, path expansion, plugin loading and running, logging, and warning, from libkrb5 to libheimbase.

The motivation for this is three-fold:

 - to keep the GSS-API mechglue more generic,
 - to make moving `bx509d` out of `kdc/` possible along with moving PKIX infrastructure in `kdc/` common to `kx509` and `bx509d` out to `lib/hx509`,
 - and to enable future work in Heimdal that is not Kerberos-specific.

I.e., to let non-Kerberos parts of Heimdal have config files (including support for plists on OS X and the registry on Windows), plugins, and logging just like Kerberos in Heimdal.

The changes are very mechanical: rename files, `s/krb5_/heim_/g`, add `lib/base/context.c`, create `krb5` API wrappers for moved APIs, and build-system support.

The log comment for the main commit proposes that reviewers make use of `--word-diff=color` when reviewing changes to the renamed files (which were renamed in a separate, rename-only commit, except for `heim_err.et`, which only had a single-line change and git is able to follow its move).  Thus, this:

```
 52 files changed, 3938 insertions(+), 2011 deletions(-)
```

is not as big as it seems.  If you review the moved files with `--word-diff=color` and the rest as usual, then this is a much smaller review than the `git log -n1 --stat` output would imply.

(The commits from the `subsidiaries-rebased` branch, aka PR #677, are included, and have been reviewed separately by @gstrauss -- I expect to push those on Monday or Tuesday.)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -21,10 +21,11 @@ build_script:
   - set dbg__type=Debug
   - title Heimdal Build %CPU% %dbg__type%
   - echo PATH=%PATH%
+  - C:\msys64\usr\bin\bash -lc "cp /c/Windows/System32/msvcr100d.dll /c/projects/heimdal"
   # Newer texinfo has no .exe's, so we have to invoke it as
   # "perl ...\makeinfo ...".  See doc/NTMakefile.
   - nmake /f NTMakefile APPVEYOR=1 MAKEINFO=makeinfo NO_INSTALLERS=1
-  - 7z a heimdal-out.zip C:\projects\heimdal\out\
+  - 7z a heimdal.zip C:\projects\heimdal
 
 test_script:
   # Packages are not validated in the Appveyor build, FYI.
@@ -33,9 +34,11 @@ test_script:
 artifacts:
   - path: heimdal-out.zip
     name: heimdal-out
+  - path: heimdal.zip
+    name: heimdal
 
 on_failure:
-  - 7z a heimdal-out.zip C:\projects\heimdal\out\
+  - 7z a heimdal-out.zip C:\projects\heimdal
   - appveyor PushArtifact heimdal-out.zip
 
 # To get RDP access to an appveyor worker for debugging a build, just


nicowilliams:
The changes were so mechanical that the only bugs I ran into were build system bugs and one solitary code bug (where a `va_list` was being passed to a varargs printf-like function, and which somehow failed to elicit a warning from gcc/clang).

nicowilliams:
Also, `make dist` currently does not work on a pristine, `./configure`d tree, but does work after a full build, and the resulting tarball does build (on Linux anyways).

nicowilliams:
A few more things that could move into `lib/base/`:

 - more of the error message functions (load the error tables into the `heim_context` instead of `krb5_context`),
 - `lib/krb5/expand_hostname.c`,
 - `krb5_set_config_files()`, `krb5_prepend_config_files()`, `krb5_prepend_config_files_default()`, `krb5_get_default_config_files()` (the `heim` version of which could take an environment variable name as an argument),  could all be generalized, and should be because non-Kerberos Heimdal code that needs configuration is likely to need to provide the same experience as the Kerberos Heimdal code.

lib/base/config_reg.c
```diff
@@ -612,38 +609,51 @@ load_config_from_regpath(krb5_context context,
  * Load configuration from registry
  *
  * The registry keys 'HKCU\Software\Heimdal' and
- * 'HKLM\Software\Heimdal' are treated as krb5.conf files.  Each
+ * 'HKLM\Software\Heimdal' are treated as heim.conf files.  Each
```
lhoward: is `heim.conf` a thing or this is a search/replace error?

lhoward:
NB I did not review the ccache collection stuff

lib/base/config_reg.c
nicowilliams: Yup, it's one search/replace too many!  Good catch!

nicowilliams:
> NB I did not review the ccache collection stuff

Good!  @gstrauss is doing that.  (And it's a trivial rebase to not have them in this branch.)

nicowilliams:
Thanks for the review!

PR #681, opened 2020-03-11
Make the FILE ccache type a collection type!

nicowilliams:
This makes the `FILE` ccache type a collection type.

Minimally, the `FILE` type will consider all ccaches whose prefix is the default `FILE` ccache name, followed by `:name` as subsidiaries.

As well, one can now configure multiple default locations in `krb5.conf` using the new `default_file_cache_collection` parameter in `[libdefaults]`.  See the man page.

### DIFF
--- a/lib/gssapi/krb5/store_cred.c
+++ b/lib/gssapi/krb5/store_cred.c
@@ -143,7 +143,6 @@ _gsskrb5_store_cred_into2(OM_uint32         *minor_status,
     const char *cs_app_name = NULL;
     OM_uint32 major_status, junk;
     OM_uint32 overwrite_cred = store_cred_flags & GSS_C_STORE_CRED_OVERWRITE;
-    OM_uint32 default_cred = store_cred_flags & GSS_C_STORE_CRED_DEFAULT;
 
     *minor_status = 0;
 
@@ -212,11 +211,9 @@ _gsskrb5_store_cred_into2(OM_uint32         *minor_status,
      *  - the default ccache
      */
     if (cs_ccache_name) {
-        default_cred = 0;
         ret = krb5_cc_resolve(context, cs_ccache_name, &id);
     } else if (cs_unique_ccache) {
         overwrite_cred = 1;
-        default_cred = 0;
         ret = krb5_cc_new_unique(context, cs_unique_ccache, NULL, &id);
     } else if (principal_is_best_for_user(context, cs_app_name,
                                           input_cred->principal,
@@ -254,8 +251,6 @@ _gsskrb5_store_cred_into2(OM_uint32         *minor_status,
     if (ret == 0)
         ret = krb5_cc_copy_match_f(context, input_cred->ccache, id, NULL, NULL,
                                    NULL);
-    if (ret == 0 && default_cred)
-        krb5_cc_switch(context, id);
 
     if ((store_cred_flags & GSS_C_STORE_CRED_SET_PROCESS) && envp == NULL)
         envp = &env;
--- a/lib/gssapi/krb5/store_cred.c
+++ b/lib/gssapi/krb5/store_cred.c
@@ -160,7 +160,9 @@ _gsskrb5_store_cred_into2(OM_uint32         *minor_status,
 	return GSS_S_NO_CRED;
     }
 
-    if (cs_ccache_name) {
+    if (cs_ccache_name && strcmp(cs_ccache_name, "unique") == 0) {
+        ret = krb5_cc_new_unique(context, NULL, NULL, &id);
+    } else if (cs_ccache_name) {
         default_cred = 0;
         ret = krb5_cc_resolve(context, cs_ccache_name, &id);
     } else {
--- a/lib/gssapi/krb5/store_cred.c
+++ b/lib/gssapi/krb5/store_cred.c
@@ -89,6 +89,35 @@ set_proc(OM_uint32 *minor, gss_buffer_set_t env)
     return GSS_S_COMPLETE;
 }
 
+/*
+ * A principal is the best principal for a user IFF
+ *
+ *  - it has one component
+ *  - the one component is the same as the user's name
+ *  - the real is the user_realm from configuration
+ */
+static int
+principal_is_best_for_user(krb5_context context,
+                           const char *app,
+                           krb5_const_principal p,
+                           const char *user)
+{
+    char *user_realm;
+    int ret;
+
+    if (!user)
+        return 0;
+
+    krb5_appdefault_string(context, app, NULL, "user_realm", NULL,
+                           &user_realm);
+    ret = user_realm &&
+        krb5_principal_get_num_comp(context, p) == 0 &&
+        strcmp(user, krb5_principal_get_comp_string(context, p, 0)) == 0 &&
+        strcmp(user_realm, krb5_principal_get_realm(context, p)) == 0;
+    free(user_realm);
+    return ret;
+}
+
 OM_uint32 GSSAPI_CALLCONV
 _gsskrb5_store_cred_into2(OM_uint32         *minor_status,
 			  gss_const_cred_id_t input_cred_handle,
@@ -107,6 +136,7 @@ _gsskrb5_store_cred_into2(OM_uint32         *minor_status,
     time_t exp_current;
     time_t exp_new;
     gss_buffer_set_t env = GSS_C_NO_BUFFER_SET;
+    const char *cs_unique_ccache = NULL;
     const char *cs_ccache_name = NULL;
     const char *cs_user_name = NULL;
     const char *cs_app_name = NULL;
@@ -142,11 +172,12 @@ _gsskrb5_store_cred_into2(OM_uint32         *minor_status,
     /* Extract the ccache name from the store if given */
     if (cred_store != GSS_C_NO_CRED_STORE) {
 	major_status = __gsskrb5_cred_store_find(minor_status, cred_store,
+                                                 "unique_ccache_type",
+                                                 &cs_unique_ccache);
+	if (GSS_ERROR(major_status))
+	    return major_status;
+	major_status = __gsskrb5_cred_store_find(minor_status, cred_store,
 						 "ccache", &cs_ccache_name);
-	if (major_status == GSS_S_COMPLETE && cs_ccache_name == NULL) {
-	    *minor_status = GSS_KRB5_S_G_UNKNOWN_CRED_STORE_ELEMENT;
-	    major_status = GSS_S_NO_CRED;
-	}
 	if (GSS_ERROR(major_status))
 	    return major_status;
 	major_status = __gsskrb5_cred_store_find(minor_status, cred_store,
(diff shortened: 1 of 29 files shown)


jaltman:
Please make sure that any changes to the FILE ccache take into account that that on Windows file paths have a colon in them.   We cannot use colon has a separator for subsidiaries.

lib/krb5/fcache.c
```diff
@@ -1162,22 +1169,225 @@ fcc_get_version(krb5_context context,
...
 
+static const char *
+my_basename(const char *fn)
+{
+    const char *base;
+
+    if (strncmp(fn, "FILE:", sizeof("FILE:") - 1))
+        return "";
+    fn += sizeof("FILE:") - 1;
+    for (base = fn + strlen(fn); base >= fn; base--) {
```
vdukhovni: I would have scanned forward... But this looks correct...

lib/krb5/fcache.c
```diff
@@ -1162,22 +1169,225 @@ fcc_get_version(krb5_context context,
...
+{
+    size_t baselen;
+
+    if (strncmp(base, "FILE:", sizeof("FILE:") - 1) == 0)
+        base += sizeof("FILE:") - 1;
+
+    baselen = strlen(base);
+    return strncmp(fn, base, baselen) == 0 &&
+        ((!match_sub_only && fn[baselen] == '\0') ||
+         (fn[baselen] == ':' && fn[baselen + 1] != '\0'));
```
vdukhovni: Don't know what's going on here. Please add commentary what `match_sub_only` is supposed to mean.

lib/krb5/fcache.c
```diff
@@ -1162,22 +1169,225 @@ fcc_get_version(krb5_context context,
...
+    if (name == NULL) {
+        *res = 1;
+        return 0;
+    }
+    if (def_locs == NULL)
+        def_locs = def_loc;
+    for (i = 0; !(*res) && def_locs[i]; i++) {
+        char *e = NULL;
+
+        if ((ret = _krb5_expand_default_cc_name(context, def_locs[i], &e)))
```
vdukhovni: Tack on `!=0` at the end?

nicowilliams:
Windows paths can't have colons at all?  Or are you just concerned about letter drives?

lib/krb5/fcache.c
```diff
@@ -1162,22 +1169,225 @@ fcc_get_version(krb5_context context,
...
-    iter = calloc(1, sizeof(*iter));
-    if (iter == NULL) {
-	krb5_set_error_message(context, ENOMEM, N_("malloc: out of memory", ""));
-	return ENOMEM;
+    /*
+     * Note: do not allow krb5_cc_default_name() to recurse via
+     * krb5_cc_cache_match().
+     * Note that context->default_cc_name will be NULL even though
+     * KRB5CCNAME is set in the environment if
+     * krb5_cc_set_default_name() hasn't
```
vdukhovni: s/hasn't/???/

jaltman:
A Windows file path is <drive>:<path>[:<alt-data-stream>].    A colon is both a drive letter separator and an alternate data stream separator.

nicowilliams:
I just tried it too, and indeed, the Windows explorer does not allow me to have a `:` in any path component.

nicowilliams:
Thanks.  I'll fix it.

lib/krb5/fcache.c
```diff
@@ -1162,22 +1169,225 @@ fcc_get_version(krb5_context context,
...
+        return 0;
+    }
+    iter->first = 0;
+
+    if (iter->d)
+        closedir(iter->d);
+    iter->d = NULL;
+    iter->curr_location = NULL;
+    if (iter->locations &&
+        (iter->curr_location = iter->locations[++(iter->location)]))
```
vdukhovni: Tack on `!= NULL` at the end for clarity?

nicowilliams:
In the past I've written code to check for `:` as the second char of paths on `WIN32`, but now I see I only ever need to check for any `:` on `WIN32`.

lib/krb5/fcache.c
```diff
@@ -1162,22 +1192,225 @@ fcc_get_version(krb5_context context,
...
+    const char *base = my_basename(iter->curr_location);
+    char *s;
+
+    *fn = NULL;
+    if (iter->d == NULL)
+        return 0;
+    for (iter->dentry = readdir(iter->d);
+         iter->dentry;
+         iter->dentry = readdir(iter->d)) {
+        if (!matchfile(iter->dentry->d_name, base, 1))
```
vdukhovni: Replace `1` with a macro that indicates what option it is turning on.  Add a long comment explaining the purpose of this function.

lib/krb5/fcache.c
```diff
@@ -1186,38 +1419,44 @@ fcc_get_cache_next(krb5_context context, krb5_cc_cursor cursor, krb5_ccache *id)
...
+        }
     }
-
-    *id = cc;
-
-    return 0;
+    for (ret = next_dir_match(context, iter, &name);
+         ret == 0 && name != NULL;
+         ret = next_dir_match(context, iter, &name)) {
+        if ((ret = try1(context, name, id)) || *id) {
```
vdukhovni: Tabs here?

jaltman:
A colon shouldn't be used in a path component on any filesystem regardless of operating system.  Its not in the POSIX safeset.

The Explorer Shell is more restrictive then the Win32 filesystem rules.   However, all of the following characters are reserved and cannot be used on Windows.

The following reserved characters:

```
    < (less than)
    > (greater than)
    : (colon)
    " (double quote)
    / (forward slash)
    \ (backslash)
    | (vertical bar or pipe)
    ? (question mark)
    * (asterisk)
```

jaltman:
Just in case it matters, a reminder that that `/`. `///`, `////`, etc are equivalent but not `//` which is reserved by POSIX for filesystem specific interpretation.

lib/krb5/fcache.c
nicowilliams: No, that's 5-space indent for the `for` loop bits, and 4-space for the body.

lib/krb5/fcache.c
```diff
@@ -206,27 +210,21 @@ fcc_resolve(krb5_context context,
...
-    }
-
-    f = calloc(1, sizeof(*f));
-    if(f == NULL) {
-	krb5_set_error_message(context, KRB5_CC_NOMEM,
-			       N_("malloc: out of memory", ""));
-	return KRB5_CC_NOMEM;
+    if ((f = calloc(1, sizeof(*f))) == NULL ||
+        (f->res = strdup(res)) == NULL ||
+        (f->sub = sub ? strdup(res) : NULL) == (sub ? NULL : "") ||
```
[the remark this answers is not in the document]
nicowilliams: Turns out the value doesn't matter, as long as it's appropriately `NULL` or not.  I could turn that into a boolean, but having the `sub` stashed away may prove useful in debugging, and in the future, so I'll keep it as a string.

lib/krb5/fcache.c
```diff
@@ -198,39 +202,56 @@ fcc_lock(krb5_context context, krb5_ccache id,
     return ret;
 }
 
+static krb5_error_code KRB5_CALLCONV
+fcc_get_default_name(krb5_context, char **);
+
+#ifdef WIN32
+#define FILESUBSEP "+"
```
vdukhovni: Talk to Greg about whether it wouldn't be better to just use the same separator on Unix and Windows...

lib/krb5/fcache.c
nicowilliams: Looking for the last component separator is easier to do from the end.

nicowilliams: (slightly less efficient, yes)

lib/krb5/fcache.c
nicowilliams: Nah.

lib/krb5/fcache.c
nicowilliams: Fixed.

lib/krb5/fcache.c
nicowilliams: Nah.

lib/krb5/fcache.c
nicowilliams: I've removed the argument.  I used to have two uses of this function.  I also renamed it.

lib/krb5/fcache.c
nicowilliams: Commentary added.

lib/krb5/fcache.c
nicowilliams: Given Jeff's comment about `:` not being in the safe set in POSIX, I will do so.

nicowilliams:
(trivia) IIRC POSIX reserves `//` on account of Apollo Domain, whose usage predates Windows UNCs.

lib/gssapi/krb5/store_cred.c
```diff
@@ -160,7 +160,9 @@ _gsskrb5_store_cred_into2(OM_uint32         *minor_status,
 	return GSS_S_NO_CRED;
     }
 
-    if (cs_ccache_name) {
+    if (cs_ccache_name && strcmp(cs_ccache_name, "unique") == 0) {
```
nicowilliams: I wonder where we should document these things.

lib/krb5/dcache.c
```diff
@@ -779,9 +779,10 @@ dcc_get_default_name(krb5_context context, char **str)
...
                                        NULL);
 
-    /* What if def_cc_colname does not start with DIR:?  We tolerate it. */
-    return _krb5_expand_default_cc_name(context, def_cc_colname,
-					str);
+    /* [libdefaults] default_cc_collection is for testing */
+    if (strncmp(def_cc_colname, "DIR:", sizeof("DIR:") - 1))
+        def_cc_colname = KRB5_DEFAULT_CCNAME_DIR;
+    return _krb5_expand_default_cc_name(context, def_cc_colname, str);
 }
```
nicowilliams: As a general rule the `get_default_name()` methods of ccaches should always return just their hard-coded default.  The `DIR` ccache does this instead so that `tests/kdc/check-cc.in` can test it.  It'd be nice to generalize this somewhat.

lib/krb5/krb5.conf.5
```diff
@@ -209,6 +209,15 @@ If you want to change the type only use
 .Li default_cc_type .
 The string can contain variables that are expanded at runtime. See the TOKEN
 EXPANSION section.
+.It Li default_file_cache_collection = Va FILE:/path/with/tokens ...
+This multi-valued parameter allows more than one path to be
+configured for the FILE credentials cache type to look in.  The FILE
+credentials cache type will also consider file names whose prefixes
+match these and end in
+.Va :name
```
nicowilliams: Bug.  It's now `+name`.

lib/krb5/fcache.c
nicowilliams: I'll change it.

lib/krb5/fcache.c
```diff
@@ -1162,62 +1205,293 @@ fcc_get_version(krb5_context context,
...
+}
+
+/*
+ * This checks that a directory entry matches a required basename and has a
+ * non-empty subsidiary component.
+ */
+static int
+matchbase(const char *fn, const char *base)
+{
+    size_t baselen = strlen(base);
```
nicowilliams: This could be passed in...

lib/krb5/fcache.c
```diff
@@ -1162,62 +1205,315 @@ fcc_get_version(krb5_context context,
     return FCACHE(id)->version;
 }
 
+static const char *
+my_basename(const char *fn)
+{
+    const char *base, *p;
```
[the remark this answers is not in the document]
nicowilliams: No, to `fn1.

lib/krb5/fcache.c
```diff
@@ -1162,62 +1205,315 @@ fcc_get_version(krb5_context context,
...
+ */
 struct fcache_iter {
-    int first;
+    const char *curr_location;
+    char *def_ccname;   /* The default ccname */
+    char **locations;   /* All the other places we'll look for a ccache */
+    char *dname;        /* dirname() of curr_location */
+    DIR *d;
+    struct dirent *dentry;
+    int location;       /* Index of current locations */
```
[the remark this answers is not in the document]
nicowilliams: Well, it's an index to `locations`.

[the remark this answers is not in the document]
vdukhovni: The phase "Index of current ..." really wants singular.

nicowilliams:
Thanks @vdukhovni!

PR #683, opened 2020-04-02
Proposed fix for issue 634

jaltman:


jaltman:
The underlying problem here is that hdb_err.et is part of lib/hdb which has dependencies on successfully building lib/krb5 first.    

I argue that hdb_err.et should be built as part of lib/krb5 not lib/hdb because lib/krb5 must initialize the hdb error table as part of krb5_context initialization.

nicowilliams:
This looks right but fails to build.  You should move `hdb_err.et` to `lib/krb5`.

I'll probably be moving all error handling functions and error table initialization into `lib/base` anyways at some point, but you could move `hdb_err.et` now.

PR #684, opened 2020-04-06
gss: smarter mechanism selection in SPNEGO

lhoward:
Make SPNEGO (and NegoEx) smarter about mechanism selection. If the initiator has specified GSS_C_MUTUAL_FLAG and/or GSS_C_ANON_FLAG, only consider mechanisms that support GSS_C_MA_AUTH_TARG and/or GSS_C_MA_AUTH_INIT_ANON (respectively) for negotiation.

### DIFF
--- a/lib/gssapi/mech/gss_mo.c
+++ b/lib/gssapi/mech/gss_mo.c
@@ -451,8 +451,6 @@ gss_indicate_mechs_by_attrs(OM_uint32 * minor_status,
 			    gss_OID_set *mechs)
 {
     struct _gss_mech_switch *ms;
-    gss_OID_set mech_attrs = GSS_C_NO_OID_SET;
-    gss_OID_set known_mech_attrs = GSS_C_NO_OID_SET;
     OM_uint32 major;
 
     major = gss_create_empty_oid_set(minor_status, mechs);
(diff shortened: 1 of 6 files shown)


lhoward:
Actually, if a mech doesn't implement gss_indicate_mechs_by_attrs(), it should probably be treated as supporting all attributes (as we know not to the contrary). So this is a bug we should fix on our side.

nicowilliams:
> Make SPNEGO (and NegoEx) smarter about mechanism selection. If the initiator has specified GSS_C_MUTUAL_FLAG and/or GSS_C_ANON_FLAG, only consider mechanisms that support GSS_C_MA_AUTH_TARG and/or GSS_C_MA_AUTH_INIT_ANON (respectively) for negotiation.

Hmm, recall that these flags are not critical.  You can not get them.

What I think this means is that mechanisms that do offer such services should be preferred over ones that don't, not that ones that don't can't be negotiated.

nicowilliams:
```commit 5a240e89f84939e147d29ce14675dd2ed3b11242
Author: Luke Howard <lukeh@padl.com>
Date:   Mon Apr 6 15:42:48 2020 +1000

    gss: GSS_Indicate_mechs_by_attrs backwards compat

    If a loadable mechanism does not support GSS_Inquire_attrs_for_mech() and no
    attributes were marked as critical, include it unconditionally in the set of
    returned mechanisms.

    This is to avoid nasty surprises with mechanisms that do not support RFC 5587,
    as future versions of Heimdal may more aggressively prune mechanisms from
    negotiation based on attribute support. (The main deployed third-party
    mechanisms are Moonshot and Samba NTLM; both of these do support
    GSS_Inquire_attrs_for_mech(), so this is defensive programming only.)
```

Maybe `gss.conf` could have a way to list a mechanism's attributes for this case?  (And also, so that if we refactor or add new mech attributes we can override or augment the ones returned by an existing mechanism.)

lhoward:
> Maybe `gss.conf` could have a way to list a mechanism's attributes for this case?  (And also, so that if we refactor or add new mech attributes we can override or augment the ones returned by an existing mechanism.)

This idea is good but probably overkill, given the original approach is mostly defensive programming. The only third-party GSS-API mechanism I'm aware of that's used in the production with Heimdal is Moonshot's GSS-API (which supports RFC 5587). There's the Samba NTLM mechanism, which also supports it, but I believe is only used with MIT. True, there are all the custom SAP mechanisms, but is anyone using them with Heimdal? And if so, with SPNEGO?

lhoward:
> Hmm, recall that these flags are not critical. You can not get them.
> 
> What I think this means is that mechanisms that do offer such services should be preferred over ones that don't, not that ones that don't can't be negotiated.

Good point. So the initiator should keep its mechanism selection as-is and then sort by feature availability? How does one sort by multiple things? I suppose MUTUAL is more "desirable" than ANON, or is it? :)

lhoward:
Going to put this in the too hard basket for now and focus on finishing SAnon.

PR #685, opened 2020-04-07
gss: order SPNEGO proposed mechs by req_flags

lhoward:
Sort the list of mechanisms proposed by the initiator so that mechanisms are preferred by their advertised support for GSS flags. For example, if GSS_C_MUTUAL_FLAG is requested, a mechanism that offers GSS_C_MA_AUTH_TARG will be preferred over one that doesn't. The flag/mechanism attribute combinations are also assigned a weight (mutual trumps anonymous, for example).

### DIFF
--- a/cf/roken-frag.m4
+++ b/cf/roken-frag.m4
@@ -367,6 +367,8 @@ AC_BROKEN([					\
 	lstat					\
 	memmove					\
 	memset_s				\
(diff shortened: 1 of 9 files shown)


lhoward:
The fast-path check of  `mech_flag_mask()` may be a premature optimisation. It could be removed.

PR #686, opened 2020-04-07
pull up automake fixes from master so proper dependencies exist even with newer automake

dariaphoebe:
automake 1.16 and later don't work on the 7.1-branch without these fixes


jaltman:
The failure is due to the lack of texinfo in the AppVeyor builder.  Its unrelated to these changes.

PR #687, opened 2020-04-09
gss: remove dead GSS_C_PEER_HAS_UPDATED_SPNEGO

lhoward:
`GSS_C_PEER_HAS_UPDATED_SPNEGO` was previously used to indicate to the SPNEGO implementation that the peer is not running an older version of Windows that cannot correctly support the generation and verification of mechListMIC. (The last version of Windows to have this bug was Windows 2003, support for which ended in 2010.)

SPNEGO is now harmonized with RFC4178 and MIT, and omits mechListMIC when the most preferred mechanism was selected, unless the peer or mechanism explicitly requires it. Samba has tested this version of SPNEGO with no known issues.

Remove GSS_C_PEER_HAS_UPDATED_SPNEGO as it has no consumers. A publicly exported symbol is being removed by this patch, but it was a private API that should have had no consumers.


lhoward:
Going to try a different approach; please ignore sorry.

PR #688, opened 2020-04-09
gss: fix downlevel Windows interop regression

lhoward:
The recent changes to SPNEGO removed support for `GSS_C_PEER_HAS_UPDATED_SPNEGO`, through which the Kerberos mechanism could indicate to SPNEGO that the peer did not suffer from SPNEGO conformance bugs present in some versions of Windows:

><7> Section 3.1.5.1: Windows 2000, Windows Server 2003, and Windows XP do not process the mechListMIC field. No mechListMIC value is produced when either the client or server completes the exchange. If a mechListMIC value is supplied to either the client or server, it is ignored. If the initiator in the GSS exchange has the last GSS token, the server returns a NegTokenResp token that has the negState field set to accept_complete and no mechListMIC field.
>
>On all other product versions shown in the applicability list at the beginning of this section, the following processing is used for the mechListMIC field:
>
>-  If AES Kerberos ciphers are negotiated by Kerberos, the signature in the SPNEGO mechListMIC field has to be processed by the recipient.

This patch restores this workaround, documented in [MS-SPNG] Appendix A <7> Section 3.1.5.1. Whilst improving interoperability with these admittedly now unsupported versions of Windows, it does introduce a risk that Kerberos with pre-AES ciphers could be negotiated in lieu of a stronger and more preferred mechanism.

Note: this patch inverts the mechanism interface from `GSS_C_PEER_HAS_UPDATED_SPNEGO` to `GSS_C_PEER_HAS_BUGGY_SPNEGO`, so that new mechanisms (which did not ship with these older versions of Windows) are not required to implement it.


lhoward:
cc @macrotex

lhoward:
If we don't care about W2K3 then we should just remove this code entirely.

jaltman:
While we don't build on these old versions of Windows there are still many of them in use in the world.   I'm reminded of that every time I go to an ATM.

lhoward:
Of course, the risk of picking rc4-hmac to downgrade SPNEGO is less of an issue if those ciphers are turned off...

lib/gssapi/version-script.map
```diff
@@ -171,7 +171,6 @@ HEIMDAL_GSS_2.0 {
 		__gss_krb5_mechanism_oid_desc;
 		__gss_ntlm_mechanism_oid_desc;
 		__gss_spnego_mechanism_oid_desc;
-		__gss_c_peer_has_updated_spnego_oid_desc;
```
nicowilliams, line 174: Do you need to export `__gss_c_inq_peer_has_buggy_spnego_oid_desc`?

Issue #273, opened 2017-04-11
Heimdal shouldn't bundle out-of-date libtommath


brianmay:
Hello,

I believe Heimdal is using a version of libtommath from 2010. I see that Redhat took up this issue in 2014:

* https://bugzilla.redhat.com/show_bug.cgi?id=1118462
* https://pagure.io/packaging-committee/issue/387

Since then, the one remaining obstacle was removed (see https://github.com/libtom/libtommath/pull/69#issuecomment-293423126).

Are there any further issues preventing Heimdal from using the system installed version of libtommath?

The upstream libtommath maintainer thinks that they might have a better fix for the x32 issue, then
what was merged in 005c8886d7fab4eacc2a54f80b777874934c6cf4.

Thanks

sjaeckel:
> The upstream libtommath maintainer thinks that they might have a better fix for the x32 issue, then
what was merged in 005c888.

c.f. point 2 @ https://github.com/libtom/libtommath/pull/69#issuecomment-293209612

FYI the x32 build is also already building fine for master, but I think it'd be better to test on develop as there are already several bugs fixed. I'll do a new release then after your testing.

nicowilliams:
I thought @vdukhovni had updated libtommath in `master` recently.

vdukhovni:
I have figured out how to and tested an updated libtommath, but have not yet pushed that to master...

nicowilliams:
@vdukhovni Maybe we should treat `libtommath` like `libedit` and `SQLite3`, and autodetect its presence/absence.

sjaeckel:
any progress on that?

PR #689, opened 2020-04-12
Import libtommath v1.2.0

lhoward:


lhoward:
cf #273

lhoward:
NB: we could use system libtommath, but Heimdal has an extra function Love added so, we'd need to check for that too. Perhaps upgrading is simpler.

lhoward:
OK, upstream won't take the `mp_find_prime()` patch as it's too close to their `mp_prime_next_prime()` API. So I need someone to review the following carefully:

```diff
diff --git a/lib/hcrypto/rsa-ltm.c b/lib/hcrypto/rsa-ltm.c
index 4ac5617cb..803da1d23 100644
--- a/lib/hcrypto/rsa-ltm.c
+++ b/lib/hcrypto/rsa-ltm.c
@@ -536,7 +536,7 @@ ltm_rsa_generate_key(RSA *rsa, int bits, BIGNUM *e, BN_GENCB *cb)
     do {
        BN_GENCB_call(cb, 2, counter++);
        CHECK(random_num(&p, bitsp), 0);
-       CHECK(mp_find_prime(&p,128), MP_YES);
+       CHECK(mp_prime_next_prime(&p,128,0), MP_OKAY);
 
        mp_sub_d(&p, 1, &t1);
        mp_gcd(&t1, &el, &t2);
@@ -548,7 +548,7 @@ ltm_rsa_generate_key(RSA *rsa, int bits, BIGNUM *e, BN_GENCB *cb)
     do {
        BN_GENCB_call(cb, 2, counter++);
        CHECK(random_num(&q, bits - bitsp), 0);
-       CHECK(mp_find_prime(&q,128), MP_YES);
+       CHECK(mp_prime_next_prime(&q,128,0), MP_OKAY);
 
        if (mp_cmp(&p, &q) == 0) /* don't let p and q be the same */
            continue;
```

lhoward:
Also commit reducing number of trials in RSA key generation needs careful review from someone like @nicowilliams.

lib/hcrypto/rsa-ltm.c
```diff
@@ -517,6 +517,7 @@ ltm_rsa_generate_key(RSA *rsa, int bits, BIGNUM *e, BN_GENCB *cb)
 {
     mp_int el, p, q, n, d, dmp1, dmq1, iqmp, t1, t2, t3;
     int counter, ret, bitsp;
+    int trials = mp_prime_rabin_miller_trials(bits);
```
sjaeckel, line 520: This shouldn't be fixed to the total number of bits!

You're generating two primes with roughly `bits/2` and as the number of trials decreases with a growing number of bits this wouldn't be an appropriate number to check the two half-sized primes.

lib/hcrypto/rsa-ltm.c
```diff
@@ -536,7 +537,7 @@ ltm_rsa_generate_key(RSA *rsa, int bits, BIGNUM *e, BN_GENCB *cb)
     do {
 	BN_GENCB_call(cb, 2, counter++);
 	CHECK(random_num(&p, bitsp), 0);
-	CHECK(mp_find_prime(&p,128), MP_YES);
+	CHECK(mp_prime_next_prime(&p, trials, 0), MP_OKAY);
```
sjaeckel, line 540: This should be `CHECK(mp_prime_next_prime(&p, mp_prime_rabin_miller_trials(bitsp), 0), MP_OKAY);`

lib/hcrypto/rsa-ltm.c
```diff
@@ -548,7 +549,7 @@ ltm_rsa_generate_key(RSA *rsa, int bits, BIGNUM *e, BN_GENCB *cb)
     do {
 	BN_GENCB_call(cb, 2, counter++);
 	CHECK(random_num(&q, bits - bitsp), 0);
-	CHECK(mp_find_prime(&q,128), MP_YES);
+	CHECK(mp_prime_next_prime(&q, trials, 0), MP_OKAY);
```
sjaeckel, line 552: and this one accordingly `CHECK(mp_prime_next_prime(&p, mp_prime_rabin_miller_trials(bits - bitsp), 0), MP_OKAY);`

lib/hcrypto/rsa-ltm.c
lhoward: I'm out of my depth here so – I can either apply your patches (and get @vdukhovni to review) or revert to Love's original `mp_find_prime()` and bring it direct into libhcrypto.

lhoward: OK, got it, We need to increase the number of trials the shorter the number of bits. Patch incoming.

lhoward:
See a08ee262e5c5bf4f04cd7940c056f74655dd742d, I just broke it out into a separate variable because of line length.

lib/hcrypto/rsa-ltm.c
nicowilliams: @lhoward has `int trials = mp_prime_rabin_miller_trials(bitsp);` above, probably so as to avoid having to wrap this line.  It's just a style thing.

nicowilliams:
At some point I guess we'll have `./configure.ac` or `cf/something.m4` autoconf support for using a system-provided libtommath?

nicowilliams:
Anyways, LGTM.

lhoward:
Didn’t get around to system provided support but I can take a look separately.

nicowilliams:
> Didn’t get around to system provided support but I can take a look separately.

That's fine.  This gets us closer.

PR #691, opened 2020-04-17
gss: try all mechanisms in gss_accept_sec_context()

lhoward:
Rather than including specific cases in the mechanism glue for NTLMSSP and unframed Kerberos, try all mechanisms when accepting a first initiator token, moving onto the next one if `GSS_S_DEFECTIVE_TOKEN` or `GSS_S_BAD_MECH` is returned.


lhoward:
The only problem here is that there's no more logging on the first round, if you specified a credential for which we couldn't find a mechanism credential. Not sure if there's a way around this.

lhoward:
Fixing properly is more complicated. Closing for now.

nicowilliams:
Superseded by #692.  Ah, I should adopt your synopsis though.

PR #692, opened 2020-04-17
Better support for "non-standard" GSS mechs

nicowilliams:
If an initial security context token doesn't have a standard header per
RFC2743 then try all mechanisms until one succeeds or all fail.

We still try to guess NTLMSSP, raw Kerberos, and SPNEGO, from tasting
the initial security context token.

### DIFF
--- a/lib/gssapi/mech/gss_accept_sec_context.c
+++ b/lib/gssapi/mech/gss_accept_sec_context.c
@@ -93,36 +93,22 @@ parse_header(const gss_buffer_t input_token, gss_OID *mech_oid)
 	return GSS_S_COMPLETE;
 }
 
-static OM_uint32
-choose_mech(const gss_buffer_t input, gss_OID *mech_oid)
+static gss_OID
+choose_mech(const gss_buffer_t input)
 {
 	OM_uint32 status;
+        gss_OID mech_oid = GSS_C_NO_OID;
 
 	/*
 	 * First try to parse the gssapi token header and see if it's a
 	 * correct header, use that in the first hand.
 	 */
 
-	status = parse_header(input, mech_oid);
-	if (status == GSS_S_COMPLETE)
-	    return GSS_S_COMPLETE;
+	status = parse_header(input, &mech_oid);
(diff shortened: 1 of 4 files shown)


nicowilliams:
I've still got to add your code, @lhoward, for the mechs to return `GSS_S_DEFECTIVE_TOKEN`...

nicowilliams:
I've to squash the fixup commits, so don't merge till I've done that.

lhoward:
Hmm, LGTM generally but not sure why tests are failing...

lib/gssapi/mech/gss_accept_sec_context.c
```diff
@@ -182,34 +172,123 @@ gss_accept_sec_context(OM_uint32 *minor_status,
...
+                                    defective_token_error.length == 0) {
+                                    defective_token_error = *output_token;
+                                    output_token->length = 0;
+                                    output_token->value = NULL;
+                                }
+                                gss_release_buffer(&junk, output_token);
+                                continue;
+                        }
+                        gss_release_buffer(&junk, &defective_token_error);
+                        goto got_one;
```
nicowilliams, line 242: Gotta set `ctx->gc_mech`!

lib/gssapi/mech/gss_accept_sec_context.c
```diff
@@ -182,34 +172,123 @@ gss_accept_sec_context(OM_uint32 *minor_status,
...
```
nicowilliams, line 279: Ditto!

nicowilliams:
> Hmm, LGTM generally but not sure why tests are failing...

See the fixup commits I just pushed.  One was a missing assignment of `ctx->gc_mech` in the token tasting case in the mech glue, and the other was a `GSS_S_FAILURE` that needed to be a `GSS_S_DEFECTIVE_TOKEN` in SPNEGO.

And with that, tests pass!

nicowilliams:
ahh, you pushed 4199118c7, which I now don't need to fix in this PR.

nicowilliams:
Thanks for the review.

PR #694, opened 2020-04-19
Fix gss_set_neg_mechs() (status: BROKEN)

nicowilliams:
This PR does the following:

 - adds a `gss_duplicate_oid_set()` function
 - stores the neg mechs in the mechglue union cred
 - makes SPNEGO a `GM_USE_MG_CRED` mechanism: it will use mechglue union creds as its creds
    - this means SPNEGO no longer has credential acquisition/inquiry/release methods
    - this requires more `GM_USE_MG_CRED` special handling in the mechglue
 - SPNEGO no longer has methods for setting or getting the negotiable mechanisms list

This is currently BROKEN.  The issue is we need more `GM_USE_MG_CRED` special handling.  Specifically, the `gmc->cred` of a union cred element for SPNEGO needs to be a _different_ mechglue union cred and it needs to be acquired somewhat specially:

 - when a cred element is acquired, either with `gss_acquire_cred*()` or `gss_add_cred*()`, we need to acquire a cred handle for SPNEGO with desired_mechs() having all the mechs (except SPNEGO or any `GM_USE_MG_CRED` mech)

We can't delay this to `gss_init_sec_context()` time: we'd have to be willing to hold on to a `cred_store` (including passwords, perhaps) until then.

A few notes:

 - `gss_get_neg_mechs()` called on a credential that has not had `gss_set_neg_mechs()` called will return `GSS_S_FAILURE` because our SPNEGO will need to know whether `gss_set_neg_mechs()` was called, and it's this or know the details of union creds.  RFC2744 doesn't say what to do in this case, but it does allow returning an error.

 - calling `gss_acquire_cred*()` with a subset of available concrete mechs + SPNEGO will internally also acquire for all the other concrete mechs as well, and this can't really be helped (especially when that subset is the empty set!)

 - ditto calling `gss_add_cred*()` to add a SPNEGO cred element to a credential that has elements for concrete mechanisms

 - maybe we should support calling `gss_init_sec_context()` with a non-default credential that lacks a SPNEGO element, and `desired_mech` set to SPNEGO -- maybe this is how it should have been specified all along


nicowilliams:
Discussing with Luke.  He proposes we never have a cred element for SPNEGO, just skip it.  If you acquire a cred only for SPNEGO, you get a handle that represents the default credential.

nicowilliams:
This is really a clone of #695.  I couldn't push to #695's branch after doing a bit of git rebase fu to put all changes to `gss_duplicate_oid_set()` in one commit.

PR #695, opened 2020-04-19
Fix gss_set_neg_mechs() (2)

lhoward:
PR #694 with review comments addressed (couldn't write to Nico's branch).

This PR does the following:

- adds a `gss_duplicate_oid_set()` function
- stores the neg mechs in the mechglue union cred
- makes SPNEGO a `GM_USE_MG_CRED` mechanism: it will use mechglue union creds as its creds
- this means SPNEGO no longer has credential acquisition/inquiry/release methods
- this requires more `GM_USE_MG_CRED` special handling in the mechglue
- SPNEGO no longer has methods for setting or getting the negotiable mechanisms list


lib/gssapi/mech/gss_acquire_cred_from.c
```diff
@@ -166,17 +166,29 @@ gss_acquire_cred_from(OM_uint32 *minor_status,
...
+		/*
+		 * Now SPNEGO supports GM_USE_MG_CRED it's no longer necessary
+		 * to specifically acquire SPNEGO credentials. However a legacy
+		 * application will do this with the intention of negotiating
+		 * all available mechanisms (subject to refinement by calling
+		 * GSS_Set_neg_mechs()). Interpret GM_USE_MG_CRED to mean all
+		 * mechanisms.
+		 */
+		if (m->gm_flags & GM_USE_MG_CRED)
+		    mechs = _gss_mech_oids;
```
nicowilliams, line 180: I think we should do this IFF there's only `GM_USE_MG_CRED` mechs in `desired_mechs`.  That's why my version of this has an `only_neg_mechs` variable.

lib/gssapi/mech/gss_add_cred_from.c
```diff
@@ -88,14 +88,75 @@ _gss_mg_add_mech_cred(OM_uint32 *minor_status,
     } else
 	major_status = GSS_S_UNAVAILABLE;
 
-    if (major_status == GSS_S_COMPLETE && out)
```
nicowilliams, line 91: But we don't want to return an error, nor call the mechanism's acquisition methods, if the `desired_mech` is a `GM_USE_MG_CRED` mechanism.

lib/gssapi/mech/gss_inquire_cred.c
```diff
@@ -99,7 +99,8 @@ gss_inquire_cred(OM_uint32 *minor_status,
 			gss_name_t mc_name = GSS_C_NO_NAME;
 			OM_uint32 mc_lifetime = GSS_C_INDEFINITE;
 
-			if (mc->gmc_mech->gm_inquire_cred == NULL)
+			if (mc->gmc_mech->gm_inquire_cred == NULL ||
+			    (mc->gmc_mech->gm_flags & GM_USE_MG_CRED))
```
nicowilliams, line 103: But now we'll never have a mech cred element for a `GM_USE_MG_CRED` mech, no?

lib/gssapi/mech/gss_inquire_cred.c
```diff
@@ -174,6 +176,17 @@ gss_inquire_cred(OM_uint32 *minor_status,
 		}
 	}
 
+	if (found && mechanisms) {
```
nicowilliams, line 179: Ah, so the answer is "no" to the above question.  So the above code can be removed.

lib/gssapi/mech/gss_acquire_cred_from.c
lhoward: Will fix.

lib/gssapi/mech/gss_add_cred_from.c
lhoward: That can't happen. `add_mech_cred_internal()`, which calls the function you commented above, is always called with a concrete mechanism.

lhoward: There's even an assertion to validate this :)

lhoward: However: if `desired_mech` is a `GM_USE_MG_CRED` mechanism, are you saying we should make it a NOOP rather than adding creds for all mechanisms?

lib/gssapi/mech/gss_inquire_cred.c
lhoward: Well, never say never. Perhaps it wants to customize the output but still use mechglue credentials. Or, if you think it's a hard guarantee we should change this to an assert.

lib/gssapi/mech/gss_inquire_cred.c
lhoward: I got a bit confused as to which no/above you mean...

lib/gssapi/mech/gss_acquire_cred_from.c
lhoward: Fixed in current version.

lib/gssapi/mech/gss_inquire_cred.c
nicowilliams: Hmm, ok.

lib/gssapi/mech/gss_inquire_cred.c
nicowilliams: > But now we'll never have a mech cred element for a GM_USE_MG_CRED mech, no?

lib/gssapi/mech/gss_add_cred_from.c
nicowilliams: This we resolved on IM last night with proposed new code from @lhoward.

lib/gssapi/mech/gss_acquire_cred_from.c
nicowilliams: This we resolved on IM last night with proposed new code from @lhoward.

lib/gssapi/mech/gss_inquire_cred.c
lhoward: Ah, I see what you mean.

lhoward: Sorry – I got the comments about the two bits of code confused. Yes, we will never have a mech cred element for a `GM_USE_MG_CRED` mech so this should become an assert instead. (Hmm, unless we need to import exported creds from a previous release? What are the lifecycle guarantees of `gss_export_cred()`?)

The `found` check is in case the mech's `gss_inquire_cred()` failed. But we can unconditionally add `GM_USE_MG_CRED` mechs to the list if you prefer.

lhoward: Re: `gss_export_cred()` we should workaround that on `gss_import_cred()` if we need to.

lhoward: We probably don't need to do anything about `gss_import_cred()` as SPNEGO isn't going to implement `gm_import_cred()`. But... given I just argued the opposite above, I'll add a check for `GM_USE_MG_CRED`.

lhoward: Ah, OK, so `gss_import_cred()` should skip `GM_USE_MG_CRED` mechs rather than returning `GSS_S_BAD_MECH`. But – should we be storing `neg_mechs` in the exported cred so it round-trips?

lhoward: I mean, if we wanted to be clever, we *could* store `neg_mechs` in the exported cred as a SPNEGO credential with just a list of mechs. That would let us import old exported creds (but it wouldn't work the other way). Whether we need to address this at all depends on whether the exported credential format is stable between releases?

lhoward: If we need (or want) to support import/cred creds in both directions with old versions then, I guess we need to recursively embed the credential handle (filtered by `neg_mechs`).

nicowilliams: I guess export cred should save the neg mechs.

lib/gssapi/mech/gss_add_cred_from.c
```diff
@@ -162,45 +219,51 @@ gss_add_cred_from(OM_uint32 *minor_status,
...
-					     cred_store, NULL,
-					     initiator_time_rec, acceptor_time_rec);
-        if (major_status != GSS_S_COMPLETE)
-            _gss_mg_error(m, *minor_status);
-        goto done;
-    }
+	major_status = GSS_S_UNAVAILABLE; /* in case of no mechs */
+
+	/*
+	 * Treat adding SPNEGO credentials (GM_USE_MG_CRED) as adding
```
nicowilliams, line 230: This is correct if `input_cred_handle == GSS_C_NO_CREDENTIAL`, otherwise, `if  input_cred_handle != GSS_C_NO_CREDENTIAL && (m->gm_flags & GM_USE_MG_CRED))` then this call is a no-op and we can just return `GSS_S_COMPLETE`.

lhoward: Thanks, sorry I missed this in your original patch. Will fix now.

lib/gssapi/mech/gss_inquire_cred.c
nicowilliams: > But – should we be storing neg_mechs in the exported cred so it round-trips?

Yes, that'd be nice.

nicowilliams: Though it's hardly critical.  Who uses export/import creds?

nicowilliams: And no, I don't think we want to be backwards-compatible for IPC tokens with multiple versions of Heimdal installed on the same host.

lhoward: OK, we will now encode neg mechs as the SPNEGO mech with a list of OIDs. This format is incompatible with previous versions of Heimdal.

lib/gssapi/mech/gss_cred.c
```diff
@@ -34,12 +34,70 @@
...
+
+    data.length = mech->length;
+    data.data = mech->elements;
+
+    ret = krb5_store_data(sp, data);
+    if (ret)
+	goto out;
+
+    for (i = 0, len = 0; i < oids->count; i++)
+	len += 4 + oids->elements[i].length;
```
nicowilliams, line 64: Why not just store the count of OIDs?  Why do we need to add all these other things?

lib/gssapi/mech/gss_cred.c
```diff
@@ -99,6 +157,23 @@ gss_export_cred(OM_uint32 * minor_status,
 	_gss_secure_release_buffer(minor_status, &buffer);
     }
 
+    if (cred->gc_neg_mechs != GSS_C_NO_OID_SET) {
+	major = export_oid_set(minor_status, GSS_SPNEGO_MECHANISM,
+			       cred->gc_neg_mechs, &buffer);
+	if (major == GSS_S_COMPLETE) {
+	    bytes = krb5_storage_write(sp, buffer.value, buffer.length);
```
nicowilliams, line 164: Why not just pass the sp to export_oid_set()?  Why not just always `export_oid_set()` unconditionally?  If the OID set is empty or `GSS_C_NO_OID_SET` then `export_oid_set()` can just write a bit fat zero.

lib/gssapi/mech/gss_cred.c
lhoward: Because there is (frustratingly) an asymmetry between `gss_import_cred()` and gss_export_cred()`. As noted in the source, `gss_export_cred()` expects the mechanism's `gm_export_cred()` (which we are emulating with `export_oid_set()`) to return:

```
mech-len mech-data cred-len cred-data
```

to be written to directly to `sp`; `gss_import_cred()` however, peeks inside this and only passes `mech-data` to `gm_import_cred()`. A similar asymmetry is present when importing and exporting security contexts. So, we're just playing by the rules I'm afraid.

lib/gssapi/mech/gss_cred.c
```diff
@@ -179,7 +314,8 @@ gss_import_cred(OM_uint32 * minor_status,
 	    goto out;
 	}
 
-	if (m->gm_import_cred == NULL) {
+	if (m->gm_import_cred == NULL &&
+	    (m->gm_flags & GM_USE_MG_CRED) == 0) {
```
nicowilliams, line 318: I mean, `m->gm_import_cred` should be `NULL` for `GM_USE_MG_CRED`, no?  And should that be `||` instead of `&&`?

lib/gssapi/mech/gss_cred.c
```diff
@@ -195,6 +331,19 @@ gss_import_cred(OM_uint32 * minor_status,
 	buffer.value = data.data;
 	buffer.length = data.length;
 
+	if (m->gm_flags & GM_USE_MG_CRED) {
+	    if (gss_oid_equal(&m->gm_mech_oid, GSS_SPNEGO_MECHANISM))
```
nicowilliams, line 335: Given the above, do we reach this code ever?

lib/gssapi/mech/gss_cred.c
lhoward: Sure, I can pass `sp` to `export_oid_set()`; I was just trying to keep things close to the existing mechanism calls for simplicity (whilst I was understanding the code too). I'll fix that now. I don't think it's necessary to always export unconditionally though. Technically there's a difference between a null and empty OID set of neg mechs, although I don't know off hand how that plays out in the rest of the code...

lib/gssapi/mech/gss_cred.c
lhoward: If a mech does *not* set GM_USE_MG_CRED and does *not* implement `gm_import_cred()`, then we cannot import the cred. So we should fail.

lib/gssapi/mech/gss_cred.c
lhoward: Yes.

lib/gssapi/mech/gss_cred.c
lhoward: Conversely, if it does set GM_USE_MG_CRED we will handle it in the mechglue (only for SPNEGO right now). Perhaps, perhaps, we should allow a mechanism that supports GM_USE_MG_CRED to implement `gm_import_cred()` for some future pseudo-mechanism?

lhoward: I might change the logic to special-case SPNEGO specifically.

lib/gssapi/mech/gss_cred.c
nicowilliams: I expect we never see empty OID sets here.

nicowilliams:
This is really a clone of #694. I couldn't push to #695's branch after doing a bit of git rebase fu to put all changes to gss_duplicate_oid_set() in one commit, so I pushed to #694, which this PR was superseding.

PR #697, opened 2020-04-21
gss: pass mechanism error tokens through SPNEGO

lhoward:
Resolves issue #486 based on a patch by Nico Williams.

A GSS-API acceptor can return an error token to be sent to the initiator. Our SPNEGO implementation discarded these when sending a SPNEGO reject response. This patch fixes the SPNEGO acceptor to convey those in the SPNEGO response.

The SPNEGO initiator is also updated to not bail out early on receiving a SPNEGO reject response from the acceptor, but instead pass the response token (if any) to gss_init_sec_context(). A reject response with no response token will continue to return an error.

### DIFF
--- a/lib/gssapi/spnego/accept_sec_context.c
+++ b/lib/gssapi/spnego/accept_sec_context.c
@@ -35,10 +35,12 @@
 
 static OM_uint32
 send_reject (OM_uint32 *minor_status,
+	     gss_buffer_t mech_token,
 	     gss_buffer_t output_token)
 {
     NegotiationToken nt;
     size_t size;
+    heim_octet_string responseToken;
 
     nt.element = choice_NegotiationToken_negTokenResp;
 
@@ -50,6 +52,13 @@ send_reject (OM_uint32 *minor_status,
     *(nt.u.negTokenResp.negState)  = reject;
     nt.u.negTokenResp.supportedMech = NULL;
     nt.u.negTokenResp.responseToken = NULL;
+
+    if (mech_token != GSS_C_NO_BUFFER && mech_token->value != NULL) {
+	responseToken.length = mech_token->length;
+	responseToken.data   = mech_token->value;
+	nt.u.negTokenResp.responseToken = &responseToken;
+     } else
+	nt.u.negTokenResp.responseToken = NULL;
     nt.u.negTokenResp.mechListMIC   = NULL;
 
     ASN1_MALLOC_ENCODE(NegotiationToken,
@@ -513,7 +522,7 @@ acceptor_complete(OM_uint32 * minor_status,
 	    ret = _gss_spnego_verify_mechtypes_mic(minor_status, ctx, mic);
 	    if (ret) {
 		if (*get_mic)
-		    send_reject(minor_status, output_token);
+		    send_reject(minor_status, GSS_C_NO_BUFFER, output_token);
 		if (buf.value)
 		    free(buf.value);
 		return ret;
@@ -888,12 +897,11 @@ acceptor_continue
 			      input_chan_bindings,
 			      &obuf,
 			      delegated_cred_handle);
-	    if (ret == GSS_S_COMPLETE || ret == GSS_S_CONTINUE_NEEDED) {
-		mech_output_token = &obuf;
-	    }
+	    mech_output_token = &obuf;
 	    if (ret != GSS_S_COMPLETE && ret != GSS_S_CONTINUE_NEEDED) {
 		free_NegotiationToken(&nt);
-		send_reject(&junk, output_token);
+		send_reject(&junk, mech_output_token, output_token);
+		gss_release_buffer(&junk, mech_output_token);
 		HEIMDAL_MUTEX_unlock(&ctx->ctx_id_mutex);
 		return ret;
 	    }
--- a/lib/gssapi/spnego/init_sec_context.c
+++ b/lib/gssapi/spnego/init_sec_context.c
@@ -427,6 +427,7 @@ spnego_reply(OM_uint32 * minor_status,
     OM_uint32 ret, minor;
     NegotiationToken resp;
     gss_buffer_desc mech_output_token;
+    NegStateEnum negState;
 
     *minor_status = 0;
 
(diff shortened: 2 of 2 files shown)


lib/gssapi/spnego/accept_sec_context.c
```diff
@@ -888,12 +897,11 @@ acceptor_continue
...
 			      delegated_cred_handle);
-	    if (ret == GSS_S_COMPLETE || ret == GSS_S_CONTINUE_NEEDED) {
-		mech_output_token = &obuf;
-	    }
+	    mech_output_token = &obuf;
 	    if (ret != GSS_S_COMPLETE && ret != GSS_S_CONTINUE_NEEDED) {
 		free_NegotiationToken(&nt);
-		send_reject(&junk, output_token);
+		send_reject(&junk, mech_output_token, output_token);
+		gss_release_buffer(&junk, mech_output_token);
```
nicowilliams, line 904: This fixes a leak!

lib/gssapi/spnego/init_sec_context.c
```diff
@@ -567,9 +571,16 @@ spnego_reply(OM_uint32 * minor_status,
...
 				       &ctx->mech_flags,
 				       &ctx->mech_time_rec);
-	    if (GSS_ERROR(ret))
+	    if (GSS_ERROR(ret)) {
 		gss_mg_collect_error(ctx->selected_mech_type, ret, minor);
+	    }
 	}
+	/*
+	 * If the acceptor rejected, we're out even if the inner context is
+	 * now complete. Note that the rejection is not integrity-protected.
```
nicowilliams, line 580: Nice note.

lib/gssapi/spnego/accept_sec_context.c
lhoward: Yes, I sneaked that in. Probably should have been a separate PR.

lib/gssapi/spnego/init_sec_context.c
lhoward: I believe you wrote that note.

lib/gssapi/spnego/init_sec_context.c
```diff
@@ -567,9 +571,16 @@ spnego_reply(OM_uint32 * minor_status,
 				       &mech_output_token,
 				       &ctx->mech_flags,
 				       &ctx->mech_time_rec);
-	    if (GSS_ERROR(ret))
+	    if (GSS_ERROR(ret)) {
 		gss_mg_collect_error(ctx->selected_mech_type, ret, minor);
+	    }
```
nicowilliams, line 576: I think here you should `if (negState == incomplete && ret == GSS_S_COMPLETE && mech_output_token.length) negState = complete;`

lib/gssapi/spnego/init_sec_context.c
```diff
@@ -655,7 +677,7 @@ spnego_reply(OM_uint32 * minor_status,
 	}
     }
 
-    if (*resp.u.negTokenResp.negState != accept_completed ||
+    if (negState != accept_completed ||
```
nicowilliams, line 680: See above comment.  Just because `negState == incomplete` doesn't mean we need to send a reply.  If we set it to `incomplete` by inference, and we don't have a token to send back, and we're locally-complete, then we should not send back a reply!

lib/gssapi/spnego/init_sec_context.c
```diff
@@ -578,19 +589,30 @@ spnego_reply(OM_uint32 * minor_status,
...
+	/*
+	 * Note that the accept_completed isn't integrity-protected, but
+	 * ctx->maybe_open can only be true if the inner context is fully
+	 * established.
+	 */
 	if (ctx->flags.maybe_open)
 	    ctx->flags.open = 1;
 
 	if (!ctx->flags.open) {
+	    free_NegotiationToken(&resp);
```
nicowilliams, line 607: Ay, another leak you're fixing.  Nice.

lib/gssapi/spnego/accept_sec_context.c
nicowilliams: Nah, it's ok here.

lib/gssapi/spnego/init_sec_context.c
nicowilliams: We could probably use a test.  In `tests/gss/check-context.in` we could try to accept a token for a name for which we don't have a keytab entry.

lib/gssapi/spnego/init_sec_context.c
lhoward: Then we may trample on `negState = requestMic`, I think?

lib/gssapi/spnego/init_sec_context.c
lhoward: OK, let me think about this some more :)

lib/gssapi/spnego/init_sec_context.c
lhoward: Of course we won't, sorry.

lhoward: I've applied your patch without the output token length check, remind me why it's needed?

lib/gssapi/spnego/init_sec_context.c
nicowilliams: Oh, well.

lib/gssapi/spnego/init_sec_context.c
lhoward: See above, you're right :)

lhoward:
Hmm, OK, that most recent patch breaks the tests. Will investigate.

lhoward:
OK, with my most recent commit I see in `check-negoex`

```
test_context: init_sec_context: gss-code: 327680  An invalid status code was supplied -- mech-code: 22 SPNEGO called got ISC call one too many
```

lhoward:
which suggests to me that `make_reply()` is being called and returning `GSS_C_CONTINUE_NEEDED`; exactly why I am still investigating

lib/gssapi/spnego/init_sec_context.c
lhoward: It turns out (after discussion offline with Nico) that the original patch was correct. negState refers to the state as viewed by the acceptor upon entering the function, further code in `spnego_reply()` relies on this and it should not be updated to the *current* state after the mechanism ISC completes.

lib/gssapi/spnego/init_sec_context.c
lhoward: See above comments – original patch was correct.

PR #700, opened 2020-04-25
spnego: Also use mechglue names

nicowilliams:
Recently we switched to using mechglue creds as SPNEGO creds in SPNEGO by completing the `GM_US_MG_CRED` code in the mechglue and making SPNEGO use it.

This PR does the same for names.

Notes:

 - this makes `gss_localname()` work when using SPNEGO, whether a mech OID is given or not
 - a test of `gss_localname()` and `gss_userok()` is now included in `tests/gss/check-context` via `test_context`
 - all call sites for calls to `_gss_find_mn()` needed thought and/or changes
 - the changes to `gss_canonicalize_name()` might be overwrought -- maybe if you call it with a `GM_US_MG_CRED` we should just fail

### DIFF
--- a/lib/gssapi/test_acquire_cred.c
+++ b/lib/gssapi/test_acquire_cred.c
@@ -337,7 +337,8 @@ main(int argc, char **argv)
 	}
 	gss_release_cred(&min_stat, &cred);
     }
-
+    gss_release_oid_set(&min_stat, &oidset);
+    gss_release_name(&min_stat, &target);
 
     return 0;
 }
--- a/lib/gssapi/test_context.c
+++ b/lib/gssapi/test_context.c
@@ -254,7 +254,7 @@ loop(gss_OID mechoid,
 	    gss_release_buffer(&min_stat, &output_token);
 
 	if (maj_stat & GSS_S_CONTINUE_NEEDED)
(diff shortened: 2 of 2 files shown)


lhoward:
Fantastic! LGTM. Just checking with SAnon.

lib/gssapi/mech/context.h
```diff
@@ -31,6 +31,7 @@
 
 struct _gss_context {
 	gssapi_mech_interface	gc_mech;
+	gssapi_mech_interface	gc_mech_actual;
```
lhoward, line 34: `gc_mech_actual` doesn't appear to get used anywhere, I agree it might be nice to have but do we need it?

lhoward:
No problems with SAnon.

lib/gssapi/mech/context.h
nicowilliams: I can remove it.  My idea had been to also implement a `GM_USE_MG_CTX` for when SPNEGO returns `GSS_S_COMPLETE`: then the context returned by SPNEGO would replace the context output to the application.  In fact, we don't need `gc_mech_actual` for that anyways, so I should remove it.

PR #702, opened 2020-04-26
hcrypto: Fix warnings in LTM

nicowilliams:
This commit fixes all the warnings in `lib/hcrypto/rsa-ltm.c`, which were mostly use of deprecated libtommath functions and ignoring their return values.  One warning is introduced that is safe to ignore.

Also, this commit re-writes how RSA keys are generated.  We pick a random number, make it odd, then set the high nibbles of `p` and `q` to be from a nibble pair chosen at random from a list of 32 such pairs such that the product `pq`'s high nibble will be at least 128.  Then we test for primality, and loop if not.

### DIFF
--- a/lib/hcrypto/rsa-ltm.c
+++ b/lib/hcrypto/rsa-ltm.c
@@ -40,96 +40,122 @@
 
 #include "tommath.h"
 
-static int
+#define CHECK(f)                                                        \
+    do { if (ret == MP_OKAY && ((ret = f)) != MP_OKAY) { goto out; } } while (0)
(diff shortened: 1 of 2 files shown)


nicowilliams:
A side-by-side (split) [view](https://github.com/heimdal/heimdal/pull/702/commits/88bd7b1260d7a2f19346db6a7f13ee538b1b43e4?diff=split&w=1) ignoring whitespace changes really helps, and makes the warnings fixes clearly mechanical.  The new RSA keygen code was reviewed by Viktor.

PR #703, opened 2020-04-27
gss: update SAnon for draft-howard-gss-sanon-13

lhoward:
draft-howard-gss-sanon-13 will move extended (RFC4757) flags from the NegoEx metadata to an optional component of the initial context token

### DIFF
--- a/lib/gssapi/sanon/accept_sec_context.c
+++ b/lib/gssapi/sanon/accept_sec_context.c
@@ -48,8 +48,10 @@ _gss_sanon_accept_sec_context(OM_uint32 *minor,
     OM_uint32 major, tmp;
     sanon_ctx sc = (sanon_ctx)*context_handle;
     gss_buffer_desc mech_input_token = GSS_C_EMPTY_BUFFER;
+    gss_buffer_desc initiator_pk = GSS_C_EMPTY_BUFFER;
     gss_buffer_desc hok_mic = GSS_C_EMPTY_BUFFER;
     gss_buffer_desc session_key = GSS_C_EMPTY_BUFFER;
+    OM_uint32 req_flags = 0;
 
     if (output_token == GSS_C_NO_BUFFER) {
 	*minor = EINVAL;
@@ -62,7 +64,7 @@ _gss_sanon_accept_sec_context(OM_uint32 *minor,
     if (input_token == GSS_C_NO_BUFFER) {
 	major = GSS_S_DEFECTIVE_TOKEN;
 	goto out;
-    } else if (sc != NULL && sc->rfc4121 != GSS_C_NO_CONTEXT) {
+    } else if (sc != NULL) {
 	major = GSS_S_BAD_STATUS;
 	goto out;
     }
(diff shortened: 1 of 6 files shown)


lib/gssapi/sanon/accept_sec_context.c
```diff
@@ -73,30 +75,50 @@ _gss_sanon_accept_sec_context(OM_uint32 *minor,
...
     major = _gss_sanon_curve25519_base(minor, sc);
     if (major != GSS_S_COMPLETE)
 	goto out;
 
+    if (mech_input_token.length > crypto_scalarmult_curve25519_BYTES) {
+	/* extra flags */
+	uint8_t *p = (uint8_t *)mech_input_token.value + crypto_scalarmult_curve25519_BYTES;
+	uint32_t dummy;
+
+	_gss_mg_decode_be_uint32(p, &dummy); /* higher 64 bits not used yet */
```
nicowilliams, line 106: High _32_ bits, not 64.

lib/gssapi/sanon/accept_sec_context.c
```diff
@@ -73,30 +75,50 @@ _gss_sanon_accept_sec_context(OM_uint32 *minor,
...
+	_gss_mg_decode_be_uint32(&p[4], &req_flags);
+    }
+
+    req_flags &= SANON_PROTOCOL_FLAG_MASK; /* do not let initiator set any other flags */
```
nicowilliams, line 110: We need a bitmask here, yes, but that restrictive?

nicowilliams: I guess we'll always report `ret_flags` as having `GSS_C_REPLAY_FLAG | GSS_C_SEQUENCE_FLAG | GSS_C_CONF_FLAG | GSS_C_INTEG_FLAG | GSS_C_ANON_FLAG | GSS_C_TRANS_FLAG`, no?

lhoward: Flags that can be inferred, or that do not change the protocol behaviour, should not be present. e.g. I don't see the value in, say, the absence `GSS_C_INTEG_FLAG` meaning you can't use integrity; if the initiator doesn't want to use it, it can just not use it.

But, if there are other flags that should be included in the mask, I'm all ears.

lib/gssapi/sanon/accept_sec_context.c
```diff
@@ -73,30 +75,50 @@ _gss_sanon_accept_sec_context(OM_uint32 *minor,
...
+
     /* compute shared secret */
-    major = _gss_sanon_curve25519(minor, sc, &mech_input_token,
+    major = _gss_sanon_curve25519(minor, sc, &initiator_pk, req_flags,
 				  input_chan_bindings, &session_key);
     if (major != GSS_S_COMPLETE)
 	goto out;
 
-    sc->flags |= GSS_C_REPLAY_FLAG | GSS_C_SEQUENCE_FLAG | GSS_C_CONF_FLAG |
```
nicowilliams, line 96: Ahh, we no longer need `sc->flags` because they go into the RFC4121 context, got it.

lib/gssapi/sanon/accept_sec_context.c
lhoward: Yeah, good point. Probably replay/sequence should be optional. Which means, should we make the flags field mandatory? (And if not, what should the default behaviour be?)

nicowilliams: An implementation is free to output more and fewer than requested.

nicowilliams: No change needed.

lhoward: OK, cool.

PR #714, opened 2020-05-20
[libedit/configure.ac] Search libtermcap for tgetent

earlchew:
Also search libtermcap.

Signed-off-by: Earl Chew <earl_chew@yahoo.com>


jaltman:
Hi @earlchew, 

Thanks for the contribution.  Could you update the commit message with some details about the platforms on which this change is necessary?

Thanks.

earlchew:
I took a closer look at the actual messages in `config.log` and the underlying issue. It seems that although `-ltermcap` fixes the configure issue, I found that on the more recent branch commit 872222d also fixes the issue that I see.

With that in mind, I think it's better to skip my original proposed change, and just tidy up the error message as I'm suggesting in  180fc33.

PR #715, opened 2020-05-22
krb5.conf include issue

iboukris:
Trying to reproduce the krb5.conf issue I had in PR #712 as a separate test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ env:
 before_install:
     - if [ $TRAVIS_OS_NAME = linux ]; then sudo apt-get update -qq; fi
     - if [ $TRAVIS_OS_NAME = linux ]; then sudo apt-get install -qq bison comerr-dev flex libcap-ng-dev libdb-dev libedit-dev libjson-perl libldap2-dev libncurses5-dev libperl4-corelibs-perl libsqlite3-dev libkeyutils-dev pkg-config python ss-dev texinfo unzip netbase keyutils; fi
-    - if [ $TRAVIS_OS_NAME = linux ]; then sudo apt-get install -qq ldap-utils gdb; fi
+    - if [ $TRAVIS_OS_NAME = linux ]; then sudo apt-get install -qq ldap-utils gdb apport; fi
     - if [ $TRAVIS_OS_NAME = linux ]; then sudo apt-get install -qq libmicrohttpd-dev; fi
     - if [ $TRAVIS_OS_NAME = linux -a -n "$COVERAGE" ]; then sudo apt-get install -qq jq; fi
     - if [ $TRAVIS_OS_NAME = osx ]; then brew update; fi
@@ -55,7 +55,7 @@ after_script:
 
 after_failure:
     - find . -name test-suite.log -print0 | xargs -0 cat
-    - if [ $TRAVIS_OS_NAME = linux ]; then echo "thread apply all bt" > x; find . -name core.\* -print | while read gdb; do gdb -batch -x x `file "$core"|sed -e "s/^[^']*'//" -e "s/[ '].*$//"` "$core"; done; fi
+    - if [ $TRAVIS_OS_NAME = linux ]; then echo "thread apply all bt" > x; find . -name core -print | while read core; do gdb -batch -x x `file "$core"|sed -e "s/^[^']*'//" -e "s/[ '].*$//"` "$core"; done; fi
 
 compiler:
     - clang
--- a/tests/gss/krb5.conf.in
+++ b/tests/gss/krb5.conf.in
@@ -31,3 +31,5 @@ include @srcdirabs@/include-krb5.conf
 [logging]
 	kdc = 0-/FILE:@objdir@/messages.log
 	default = 0-/FILE:@objdir@/messages.log
+
+include @srcdirabs@/missing-krb5.conf


iboukris:
====== krb5.conf include issue.
Segmentation fault (core dumped)

iboukris:
With some clumsy debugging on travis, here's the backtrace. The script in .travis.yml to find cores and print the backtrace, needs fixing.

```
$ if [ $TRAVIS_OS_NAME = linux ]; then echo "thread apply all bt" > x; find . -name core -exec gdb -batch -x x /home/travis/build/heimdal/heimdal/ci-build/lib/gssapi/.libs/lt-test_context {} \; ; fi
[New LWP 16988]
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Core was generated by `/home/travis/build/heimdal/heimdal/ci-build/lib/gssapi/.libs/lt-test_context ho'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007f2d2ba05cc0 in vfprintf () from /lib/x86_64-linux-gnu/libc.so.6
Thread 1 (Thread 0x7f2d2ccf5740 (LWP 16988)):
#0  0x00007f2d2ba05cc0 in vfprintf () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x00007f2d2ba2d773 in vasprintf () from /lib/x86_64-linux-gnu/libc.so.6
#2  0x00007f2d2ae855d6 in heim_vset_error_message (context=<optimized out>, ret=<optimized out>, fmt=<optimized out>, args=<optimized out>) at /home/travis/build/heimdal/heimdal/lib/base/error_string.c:74
#3  heim_set_error_message (context=0x1107170, ret=2, fmt=0x7f2d2ae8c0d3 "%s:%u: %s") at /home/travis/build/heimdal/heimdal/lib/base/error_string.c:56
#4  0x00007f2d2ae7fdfc in heim_config_parse_file_multi (context=0x1107170, fname=0x11071c0 "../../tests/gss/new_clients_k5.conf", res=0x7ffcac11b778) at /home/travis/build/heimdal/heimdal/lib/base/config_file.c:674
#5  0x00007f2d2ae81f40 in heim_set_config_files (context=0x1107170, filenames=<optimized out>, res=0x7ffcac11b778) at /home/travis/build/heimdal/heimdal/lib/base/context.c:356
#6  0x00007f2d2c639c71 in krb5_set_config_files (context=<optimized out>, filenames=0x7ffcac11a458) at /home/travis/build/heimdal/heimdal/lib/krb5/context.c:693
#7  krb5_init_context (context=0x607888 <context>) at /home/travis/build/heimdal/heimdal/lib/krb5/context.c:460
#8  0x0000000000402255 in main (argc=2, argv=0x7ffcac11ba58) at /home/travis/build/heimdal/heimdal/lib/gssapi/test_context.c:726

**Comment by jaltman** [comment]
Please try modifying heim_config_parse_debug() to initialize 

  *lineno = 0;
  *error_message = "";

**Comment by jaltman** [comment]
I'm also not convinced that the heim_set_error_message() at lib/base/error_string.c:56 is the right thing to do since heim_config_parse_debug() might have already set its own error message that will be deleted and replaced.

**Comment by jaltman** [comment]
@iboukris good job with the reproduction!

**Comment by iboukris** [comment]
Initializing lineno and err_message does prevent the crash, but the test still fails, meaning the include didn't work. btw, disabling optimization with CFLAGS=-O0 also has prevented the crash.

**Comment by jaltman** [comment]
The crash is because 'str' in heim_config_parse_file_multi() is uninitialized when heim_set_error_message() is called at lib/base/error_string.c:56.    Disabling optimization prevents the crash because doing so initializes 'str'.  

The test doesn't work because a configuration file containing a line of `====== foo` is invalid input.  The parser finds the `=`' operator with no left hand side and potentially outside of a section.

**Comment by jaltman** [comment]
Indeed, `krb5_config_parse_dir_multi()` includes

`        ret = heim_config_parse_file_multi(context, path, res);
        free(path);
        if (ret == ENOMEM) {
            (void) closedir(d);
            return heim_enomem(context);
        }
        /* Ignore malformed config files so we don't lock out admins, etc... */`

which is great except for the fact that by throwing away the error and returning 0 we ignore the krb5 error message which we constructed to explain what was wrong with the config file.

Ignoring invalid include files was introduced as part of commit fe43be85587f834266623adb0ecf2793d212a7ca.

**Comment by jaltman** [comment]
Try adding

`diff --git a/lib/base/config_file.c b/lib/base/config_file.c`
`index c41f4d624..2e3d9c81f 100644`
`--- a/lib/base/config_file.c`
`+++ b/lib/base/config_file.c`
`@@ -521,12 +521,14 @@ heim_config_parse_dir_multi(heim_context context,`
`             return heim_enomem(context);`
`         }`
`         ret = heim_config_parse_file_multi(context, path, res);`
`+        /* Ignore malformed config files so we don't lock out admins, etc... */`
`+       if (ret && ret != ENOMEM)`
`+           heim_warnx(context, "Included file '%s' invalid", path);`
`         free(path);`
`         if (ret == ENOMEM) {`
`             (void) closedir(d);`
`             return heim_enomem(context);`
`         }`
`-        /* Ignore malformed config files so we don't lock out admins, etc... */`
`     }`
`     (void) closedir(d);`
`     return 0;`

**Comment by iboukris** [comment]
> It might be worth 'cat'ing the contents of new_clients_k5.conf and the path that include @srcdirabs@/krb5.conf resolves to.

Oh, I now remember being here before, I'll try to build locally with a different build dir like in travis, that was probably my error all along. Thanks!

> The crash is because 'str' in heim_config_parse_file_multi() is uninitialized

Would it make sense to initialize it in the callers like I did now?

> Try adding ... which should send the include file parsing error to stderr but permit execution to continue.

Done, but travis didn't pick it up, I'll re-push later.

**Comment by jaltman** [comment]
> Would it make sense to initialize it in the callers like I did now?

It can be initialized there as well but functions with output parameters should not complete without initializing the contents.  Otherwise every call site must initialize the parameters before the call and when a developer forgets to do so we end up with exactly this problem.

**Comment by iboukris** [comment]
> It can be initialized there as well but functions with output parameters should not complete without initializing the contents.

Got it, I'll change it back.

> Oh, I now remember being here before, I'll try to build locally with a different build dir like in travis, that was probably my error all along. 

Yeah it was it, and I was already here in my last comment on PR #656..

File: lib/base/config_file.c [hunk 1]
```diff
@@ -521,12 +521,14 @@ heim_config_parse_dir_multi(heim_context context,
             return heim_enomem(context);
         }
         ret = heim_config_parse_file_multi(context, path, res);
+	/* Ignore malformed config files so we don't lock out admins, etc... */
+	if (ret && ret != ENOMEM)
+	    heim_warnx(context, "Ignoring '%s'", path);
```
jaltman, line 526: Sorry, this should have been 

`heim_warnx(context, ret, "Ignoring '%s'', path);

jaltman:
That change certainly produced a large number of seg faults.

iboukris:
Here we can see the core dump printing is fixed, now I'll add the fix for the crash:
https://travis-ci.org/github/heimdal/heimdal/builds/690825040

jaltman:
@iboukris I've pushed https://github.com/heimdal/heimdal/pull/719 which has some overlap with this series.   In particular, it includes a better version of the malformed config warnings and it includes the initialize error output parameters change.

iboukris:
I removed the overlapping commits, left the travis fix and the test which if good could be moved to #719. Looks like Nico fixed the appveyor failure.

PR #723, opened 2020-05-28
Catch many compile warnings as errors in Travis CI with -Werror via --enable-developer

abartlet:
This should make it harder to introduce problem code into Heimdal and make it easier to import a new Heimdal version into Samba (where very strict warning options are in use).

Compiler warning options are very compiler version dependent so this list will need to be refined as the compiler used in Travis CI is upgraded.

### DIFF
--- a/lib/gssapi/mech/gss_krb5.c
+++ b/lib/gssapi/mech/gss_krb5.c
@@ -486,7 +486,7 @@ gss_krb5_ccache_name(OM_uint32 *minor_status,
 		     const char **out_name)
 {
     struct _gss_mech_switch *m;
-    gss_buffer_desc buffer;
+    gss_buffer_desc buffer = GSS_C_EMPTY_BUFFER;
     OM_uint32 junk;
 
     _gss_load_mech();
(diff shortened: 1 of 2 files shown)


abartlet:
The single warning in the test code for -Wempty-body would be worth fixing in a separate commit, but I'm trying to just land this for now.

lib/gssapi/mech/gss_krb5.c
```diff
@@ -494,11 +494,11 @@ gss_krb5_ccache_name(OM_uint32 *minor_status,
     if (out_name)
 	*out_name = NULL;
 
+    buffer.value = rk_UNCONST(name);
     if (name) {
-	buffer.value = rk_UNCONST(name);
 	buffer.length = strlen(name);
     } else {
-	_mg_buffer_zero(&buffer);
+	buffer.length = 0;
```
lhoward, line 501: this should probably be in the other comment, maybe just initialise buffer to `GSS_C_EMPTY_BUFFER` at the start too?

lhoward: and remove the else block

abartlet: Do you mean the other PR?  It is in it's own commit, needed to get the -Werror changes in.

I'll try the initialisation.

lhoward: Great, just wasn't sure how this commit was related to the subject "Catch many compile warnings as errors in Travis CI with -Werror via --enable-developer"

lhoward: Ah, it's a separate commit. All good! As you were.

PR #724, opened 2020-05-28
kdc: Provide flag to hint to KDC that this is a FAST key lookup

abartlet:
For Samba the FAST key is not stored in the replicated DB, so
this helps Samba find it in the Samba hdb module.

Likewise, the armor key might be either a server or a krbtgt
key so call this with the right flags.

In-tree databases in Heimdal do not use these flags.


abartlet:
Replaced by #789

PR #728, opened 2020-05-28
Fix more heimdal warnings

abartlet:
A few more warnings fixed and in one case we are able to remove a warning from the permitted list.

### DIFF
--- a/lib/hcrypto/rsa-ltm.c
+++ b/lib/hcrypto/rsa-ltm.c
@@ -132,7 +132,7 @@ ltm_rsa_private_calculate(mp_int * in, mp_int * p,  mp_int * q,
 {
     mp_err ret;
     mp_int vp, vq, u;
-    int where = 0; /* Ignore the set-but-unused warning from this */
+    int where HEIMDAL_UNUSED_ATTRIBUTE = 0;
 
     FIRST(mp_init_multi(&vp, &vq, &u, NULL));
 
@@ -479,7 +479,7 @@ gen_p(int bits, enum gen_pq_type pq_type, uint8_t nibble_pair, mp_int *p, mp_int
     size_t len = (bits + 7) / 8;
     int trials = mp_prime_rabin_miller_trials(bits);
     int counter = 0;
-    int where = 0; /* Ignore the set-but-unused warning from this */
+    int where HEIMDAL_UNUSED_ATTRIBUTE = 0;
 
 
     FIRST(mp_init_multi(&t1, &t2, NULL));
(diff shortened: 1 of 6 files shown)


lib/hcrypto/rsa-ltm.c
```diff
@@ -132,7 +132,7 @@ ltm_rsa_private_calculate(mp_int * in, mp_int * p,  mp_int * q,
 {
     mp_err ret;
     mp_int vp, vq, u;
-    int where = 0; /* Ignore the set-but-unused warning from this */
+    int where HEIMDAL_UNUSED_ATTRIBUTE = 0; /* Ignore the set-but-unused warning from this */
```
jaltman, line 135: Please remove the `Ignore the set-but-unused warning from this` comment.

lib/hcrypto/rsa-ltm.c
```diff
@@ -479,7 +479,7 @@ gen_p(int bits, enum gen_pq_type pq_type, uint8_t nibble_pair, mp_int *p, mp_int
     size_t len = (bits + 7) / 8;
     int trials = mp_prime_rabin_miller_trials(bits);
     int counter = 0;
-    int where = 0; /* Ignore the set-but-unused warning from this */
+    int where HEIMDAL_UNUSED_ATTRIBUTE = 0; /* Ignore the set-but-unused warning from this */
```
jaltman, line 482: same here.

lib/roken/getaddrinfo-test.c
```diff
@@ -129,7 +129,8 @@ main(int argc, char **argv)
 	return 0;
     }
 
-    rk_SOCK_INIT();
+    if (rk_SOCK_INIT())
```
jaltman, line 132: This will end up being

   if (0) errx(1, ....);

which will generate a warning of its own (at least in Coverity).  Most likely some variation of a dead code warning.   

My opinion is that if this change will be made that rk_SOCK_INIT() should become a proper function and not just a macro.

lib/hcrypto/rsa-ltm.c
abartlet: Done

lib/roken/getaddrinfo-test.c
abartlet: It is now a proper inline function.  Long term we can't really win against Coverity and still do the cross-platform case where it is a call that could fail on windows, but this may be a better compromise and makes a real-world improvement in the number of warnings we need to allow (where other cases might be real bugs)

abartlet:
I should have made clearer at the start, this allows building on Ubuntu 18.04 with the list of permitted warnings from the new .travis-ci.yml check plus only `-Wno-error=implicit-fallthrough=`.  This makes it easier to whittle down that list and stop them coming back.

PR #730, opened 2020-05-31
Fix Windows i386 test failures

jaltman:
These two commits fix transient failures of the in-tree tests on Windows i386.

For discussion only at this point.   The impact of the ASN1CALL change on 32-bit Windows consumers must be evaluated in more detail.


jaltman:
This change is safe for "master" because its a major version change and the fact that the DLLs are linked together as part of an assembly.   Therefore, a mismatch of calling conventions is not possible.

PR #735, opened 2020-06-28
hx509 context memory leaks

jaltman:
The implementation of hx509_context_free() fails to free the parsed configuration file data.

On Windows, no differentiation was made between KRB5_CONFIG and new configurations such as HX509_CONFIG when obtaining configuration file lists from the registry.


jaltman:
These changes have been confirmed to remove the memory leak.

nicowilliams:
FYI, LGTM.

PR #736, opened 2020-07-05
base: implement atomics using stdatomic

lhoward:
Where C11 stdatomic is available, use it to implement atomics in lib/base


lhoward:
Arguably we should just switch everything over to using `stdatomic` and provide an `libroken` implementation. ffmpeg has such an implementation (for GCC, Solaris and Windows) although it's not suitably licensed.

jaltman:
I would prefer to see these changes broken out into three (possibly four) separate commits 
0. move heim_base_atomics defines out of heimbase.h and into its own header, heimbase_atomics.h  [optional].   
1. implement HAVE_STDATOMIC_H definitions for the heim_base_atomic operations that are already supported and used in the tree
2. implement heim_base_atomic_init(), heim_base_atomic_load(), heim_base_atomic_store(), heim_base_atomic_barrier() and any other new operation for all platforms.   We need to explicitly call out the platforms where load/store ordering is different.  Note that stdatomics atomic_load provides a specific sequential ordering guarantee and we should aim to be consistent for all platforms.   
3. Re-implement HEIM_SLIST_ATOMIC_FOREACH(), heim_retain(), heim_release(), etc using the new functionality.

lhoward:
I need to double check on the ordering guarantees (I believe it's OK) but I have separated the commits.

jaltman:
@lhoward Thanks for breaking up the series.   One small nit, could you re-order "base: use heim_base_atomic_load in heimqueue.h" after "base: add atomic init, load and store macros"?   We can use something before it has been introduced.

Thanks again.

lhoward:
I think that might be GitHub weirdness, ordering by timestamp rather than commit, as I see the following when I do `git rebase -i master`

```pick 6e17477f7 base: move atomic macros into separate header
pick a13d938b4 base: heim_base_exchange_{32,64} for platforms without atomics
pick 56d6543ad base: use <stdatomic.h> for atomics, where present
pick 41af452a5 base: add atomic init, load and store macros
pick 643baa935 base: use heim_base_atomic_load in heimqueue.h
pick a7ed68c07 base: rename heim_base_atomic_{max,type} to ...integer_{max,type}
pick d51cf0d99 base: use atomic load/store in reference counting
```

jaltman:
@lhoward Yeah, https://github.com/heimdal/heimdal/compare/master...PADL:lukeh/stdatomic shows the right thing.

jaltman:
This series looks good to me BUT Travis has stopped building pull requests.   I would like to see this series build successfully on Linux and macOS before merging.

lhoward:
OK, so it builds on macOS OK (that's my development platform). I can try on Linux and Solaris.

jaltman:
macOS passed but there is a Xenial failure 
https://travis-ci.org/github/heimdal/heimdal/builds/707483707

lhoward:
Thanks, looking.

lhoward:
OK, I'll need to spin up a modern enough Linux VM to fix this. May take some time.

lhoward:
OK, please retry Travis now.

lhoward:
BTW, I wouldn't mind dropping support for platforms that don't have atomic intrinsics. The workaround we have for such platforms (global lock) is pretty ugly, and it is broken in master anyway (missing a couple of functions) and no one has complained yet. We have native compiler support for Solaris and AIX, as well as gcc and clang on (presumably) all platforms.

lhoward:
It failed on Travis macOS owing to an unrelated error, but the Linux build is fixed.

jaltman:
> 
> 
> It failed on Travis macOS owing to an unrelated error, but the Linux build is fixed.

The macOS error was a transient problem within Travis-CI infrastructure.    I restarted the build and it looks like it will complete this time.

jaltman:
> 
> 
> BTW, I wouldn't mind dropping support for platforms that don't have atomic intrinsics. The workaround we have for such platforms (global lock) is pretty ugly, and it is broken in master anyway (missing a couple of functions) and no one has complained yet. We have native compiler support for Solaris and AIX, as well as gcc and clang on (presumably) all platforms.

Debian indicated that they would have to drop Heimdal on all platforms if we break the build for i386.

lhoward:
Re: i386, let my investigate and get back to you.

lhoward:
Ah OK, so we have 64-bit atomic intrinsics on i686 but not i386. I wonder how the existing code is working on i386 in the first place. It’s late here, but sounds like this needs more investigation.

lhoward:
I don't think this PR is any worse on i386 than what we had before. I'm yet to find a good explanation of how `atomic_exchange()` and `__sync_swap()` work on i386 with 64-bit types.

Practically I believe we are safe: 64-bit atomic operations on 32-bit platforms (with 32-bit `time_t` at least) are only used in the keyring credential cache, to set the principal and cache ID in a single operation. I don't believe this is any racier code than, say, the filesystem credential cache (with multiple processes accessing the filesystem).

As such, I believe this patch is good to merge. However, care should be taken with future use of 64-bit atomic operations, whilst i386 still needs to be supported. Note these concerns to do not apply to i686 and above as they have  64-bit atomic instructions.

PR #738, opened 2020-07-10
gss/krb5: acquire_cred_with_password set opt default flags

jaltman:
acquire_cred_with_password() must call
krb5_get_init_creds_opt_set_default_flags() to initialize the
krb5_get_init_creds option flags to the values obtained from
the krb5_context.

Change-Id: Icd8c500dd0787a781c2382284f19cef277b1d30b


jborean93:
Sorry for the delay in testing this, just confirmed on my hosts that after these changes credentials acquired with `gss_acquire_cred_with_password` use the config settings that control the specific flags used when retrieving the credential.

Thanks @jaltman for looking into this and fixing the problem.

PR #726, opened 2020-05-28
32-bit time_t enhancements

nicowilliams:

PR #743, opened 2020-07-13
32-bit time_t enhancements

jaltman:
The intent of this change is to potentially handle Kerberos ticket lifetimes post-2038.

This PR replaces https://github.com/heimdal/heimdal/pull/726 which was initiated from a non-stable branch.


jaltman:
@nicowilliams I've updated the changes to build successfully on Windows.

jaltman:
@nicowilliams The most recent push corrects a mismatch of the prototype and declaration of heim_time2generalizetime that caused the macOS build to fail.

jaltman (commented):
This series is still not correct as we are passing krb5_timestamp pointers and KerberosTime pointers into functions that require time_t pointers.   Doing so is generating warnings but should be fatal.

Search for krb5_timestamp and KerberosTime in the output from https://api.travis-ci.org/v3/job/707961249/log.txt

jaltman:
A decision needs to be made regarding the time format for the hdb functions.   I think it should be KerberosTime.

In addition, we need to handle the restriction of the MIT Dump file format which is restricted to 32-bits.

Neustradamus:
Linked to:
- https://github.com/heimdal/heimdal/issues/533
- https://github.com/heimdal/heimdal/pull/726
- https://github.com/heimdal/heimdal/pull/743
- https://github.com/heimdal/heimdal/issues/1011
- https://github.com/heimdal/heimdal/issues/1231
- https://github.com/heimdal/heimdal/issues/1385

PR #744, opened 2020-07-14
base: require platform/compiler support for atomics

lhoward:
If there are no platform or compiler atomic intrinsics, we currently fallback to a single global mutex for all atomic operations. We anticipate that all modern platforms support atomic operations, and if we don't currently have
support, it would be better to add them rather than falling back to this unperformant workaround.

For the record, the current implementation supports C11, GCC, Solaris, AIX and Windows atomics (used in that order of preference).

### DIFF
--- a/lib/base/heimbase-atomics.h
+++ b/lib/base/heimbase-atomics.h
@@ -156,70 +156,7 @@ heim_base_exchange_64(uint64_t *p, uint64_t newval)
 #define heim_base_exchange_32(t,v)	((ULONG)InterlockedExchange((LONG volatile *)(t), (LONG)(v)))
 #define heim_base_exchange_64(t,v)	((ULONG64)InterlockedExchange64((LONG64 violatile *)(t), (LONG64)(v)))
 
-#else
-
-#include <heim_threads.h>
-
-#define HEIM_BASE_NEED_ATOMIC_MUTEX 1
-extern HEIMDAL_MUTEX _heim_base_mutex;
-
-#define heim_base_atomic_integer_type	unsigned int
-#define heim_base_atomic_integer_max	UINT_MAX
-
-static inline heim_base_atomic_integer_type
-heim_base_atomic_inc(heim_base_atomic_integer_type *x)
-{
-    heim_base_atomic_integer_type t;
-    HEIMDAL_MUTEX_lock(&_heim_base_mutex);
-    t = ++(*x);
-    HEIMDAL_MUTEX_unlock(&_heim_base_mutex);
-    return t;
-}
-
-static inline heim_base_atomic_integer_type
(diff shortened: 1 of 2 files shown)


lhoward:
I'm on the fence (pun!) about this one, but feedback appreciated.

jaltman:
I would rather not remove this because I want Heimdal to build on platforms where we do not have an implementation for intrinsic operations.   Although I would be supportive of requiring an explicit #define in order to enable it.

lhoward:
Maybe we can just put a #warning in instead.

jaltman:
Perhaps HEIMDAL_DEPRECATED could be used in some way.

jaltman:
Indirectly related but I will mention it here 
`/Users/travis/build/heimdal/heimdal/lib/sqlite/sqlite3.c:20839:17: warning: 'OSAtomicCompareAndSwapPtrBarrier' is deprecated: first deprecated in macOS 10.12 - Use atomic_compare_exchange_strong() from <stdatomic.h> instead [-Wdeprecated-declarations]

      success = OSAtomicCompareAndSwapPtrBarrier(NULL, newzone, 

                ^

/usr/include/libkern/OSAtomicDeprecated.h:547:6: note: 'OSAtomicCompareAndSwapPtrBarrier' has been explicitly marked deprecated here

bool    OSAtomicCompareAndSwapPtrBarrier( void *__oldValue, void *__newValue, void * volatile *__theValue );

        ^

1 warning generated.
`

lhoward:
I looked into this separately. Apparently Apple won’t really be deprecating it (sook), so sqlite left as-is. I’ll dig up the link tomorrow. Happy to fix, if you’re ok with diverging from upstream.

lhoward:
See [here](https://sqlite-users.sqlite.narkive.com/5uIHXIrZ/mac-os-sierra-deprecation-in-bindings). From an Sqlite maintainer:

> Engineers at Apple tell me that they are still using OSAtomicCompareAndSwapPtrBarrier() inside of essential OS-X components and that the routine will not be going away any time soon.
> 
> If you are still concerned, your work-around is to compile with -DSQLITE_WITHOUT_ZONEMALLOC. (I have to do that when compiling SQLite for OS 10.2, since that release predates the OSAtomicCompareAndSwapPtrBarrier().)

I will investigate a proper fix using Heimdal-atomics though.

lhoward:
Hmm, `stdatomic.h` is only available in macOS 10.12 and later. I think it's safe to set `SQLITE_WITHOUT_ZONEMALLOC` though, as the `OSAtomicCompareAndSwapPtrBarrier()` branch is only used with single-CPU systems which is pretty much no macOS system these days.

lhoward:
However, given there are a probably a bunch of other warnings, I vote to leave as-is.

PR #746, opened 2020-07-20
enable keepalive mode for kadmin sockets

jaltman:


lhoward:
Haven't looked closely (just did a search on GitHub), but does kadmind need to be told to ignore `SIGPIPE` or is it doing that already?

PR #751, opened 2020-08-29
krb5: consolidate default ccache type logic into get_default_cc_type()

lhoward:


lhoward:
`get_default_cc_type_win32()` is kinda broken though for API:krb5cc

lib/krb5/cache.c
```diff
@@ -252,6 +252,42 @@ krb5_cc_resolve(krb5_context context,
...
+    }
+
+    /*
+     * If the API:krb5cc ccache can be resolved,
+     * use it as the default.
+     */
+    ret = krb5_cc_resolve(context, "API:krb5cc", &id);
+    krb5_cc_close(context, id);
+    if (ret == 0) {
+	return "API"; /* XXX will this work? */
```
nicowilliams, line 284: Yes, it should work.  the `"API:krb5cc"` might not be needed -- `"API:"` alone should work fine, I think, since `acc_get_default_name()` does do something to discover a proper default ccache name (namely it calls the CCAPI's `get_default_ccache_name()` function).

PR #752, opened 2020-08-29
krb5: simplify krb5_cc_resolve_sub() logic

lhoward:
`krb5_cc_resolve_sub()` redundantly checks the collection name for a cctype, and also assumes that `cctype` can be NULL (which is not the case as `get_default_cc_type()` always returns non-NULL).

### DIFF
--- a/lib/krb5/cache.c
+++ b/lib/krb5/cache.c
@@ -100,6 +100,11 @@ main (int argc, char **argv)
 * @endcode
 */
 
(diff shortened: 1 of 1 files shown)


lhoward:
Good idea, Jeff!

jaltman:
I think there is a missing "static"

lhoward:
Thanks, I've fixed it, weird that it compiled for me.

PR #753, opened 2020-09-01
krb5: Fix API: ccache failure

nicowilliams:


lhoward:
LGTM, just put something in the commit message?

SimonWilkinson (commented):
Code looks good to me. Is more needed in the commit message?

For reference, this fixes an issue introduced by 7bf4d76e75e904dd65a0fbb90c9cad981245f714

lib/krb5/acache.c
```diff
@@ -557,6 +557,11 @@ acc_resolve_2(krb5_context context, krb5_ccache *id, const char *res, const char
     }
 
     error = (*a->context->func->open_ccache)(a->context, res, &a->ccache);
+    if (error == ccErrCCacheNotFound) {
+        a->ccache = NULL;
+        a->cache_name = NULL;
```
SimonWilkinson, line 562: Do we not need to also free(s); here?

SimonWilkinson (changes requested):
Sorry, there's also a missing free()

lib/krb5/acache.c
lhoward: for a->cache_name? a->ccache presumably is NULL in this case?

lhoward:
Also there's a #pragma(pack) change for ABI compatibility on ARM, right? Or is that a separate issue?

lib/krb5/acache.c
lhoward: I don't think so.` a->ccache` is NULL because `open_ccache()` failed. `a->cache_name` is set by `get_cc_name()` which is not yet called. Or did I misunderstand?

lhoward: Which makes me wonder if it's necessary to set either of these to NULL?

lhoward: Also, should we be calling `get_kdc_time_offset()` in the `ccErrCCacheNotFound` case? If so, this logic should be moved further up.

lhoward:
BTW, @SimonWilkinson have you tried building with `-DKCM_IS_API_CACHE=1`? I'm wondering if this should be the default given versions of macOS that don't use Heimdal are less common. When this is defined, the API cache is implemented directly in terms of KCM rather than CCAPI.

lhoward (changes requested):
See comments re: NULL, KDC time offset and whether we should use KCM

lib/krb5/acache.c
SimonWilkinson: s may have been allocated by the asprintf() in the if (sub) block above. If we return without freeing s, it will be leaked.

jaltman:
> See comments re: NULL, KDC time offset and whether we should use KCM

@lhoward, I believe the assignment of acache and acache_name to NULL is appropriate as documentation if nothing else.  I do not believe it is appropriate to return the cache offset since we are returning "No Cache".

PR #755, opened 2020-09-18
Hierarchical iprop

nicowilliams:


nicowilliams:
FYI, @vdukhovni reviewed and approved.  He did find several issues, including that iprop would block when the log is empty and not even send a complete dump.

PR #757, opened 2020-10-03
hdb: Remove default HDB backend footgun

nicowilliams:
The current state of the world is such that the default HDB is named `"heimdal"` in the `LOCALSTATEDIR`, so typically it's `/var/heimdal/heimdal`.  The actual backend for this is to be determined at run-time because a backend type prefix is not part of the default HDB name.

Well, the default backend type selection at run-time depends build configuration.  So if you have BDB installed but not LMDB, you'll get BDB as the default backend, but if you later install LMDB and rebuild Heimdal then default backend type selection will change, and if you install the resulting artifacts and restart a KDC that was running with the default HDB... the KDC will fail to start!

The fix is to be a bit smarter about default backend selection.

This PR includes two commits:

 - The first commit implements HDB "tasting" for file-based HDB backends so that if you have an HDB of type `foo:` and later add support for a new preferred backend `bar:` you won't stop being able to open the default HDB that exists.

 - The second commit uses an HDB type from the `./configure` run, which can be affected by the `./configure` `--with-db-type-preference=<list>` option.

### DIFF
--- a/lib/hdb/hdb.c
+++ b/lib/hdb/hdb.c
@@ -802,7 +802,19 @@ hdb_create(krb5_context context, HDB **db, const char *filename)
             *db = NULL;
         }
     }
+#ifdef HDB_DEFAULT_DB_TYPE
+    if (cb_ctx.h == NULL || cb_ctx.h->prefix == NULL) {
+        /*
+         * If still we've not picked a backend, use a build configuration time
+         * default.
+         */
+        for (cb_ctx.h = methods; cb_ctx.h->prefix != NULL; cb_ctx.h++)
+            if (strcmp(cb_ctx.h->prefix, HDB_DEFAULT_DB_TYPE) == 0)
+                break;
+    }
+#endif
     if (cb_ctx.h == NULL || cb_ctx.h->prefix == NULL)
+        /* Last resort default */
         cb_ctx.h = &default_dbmethod;
     if (cb_ctx.h == NULL || cb_ctx.h->prefix == NULL) {
         krb5_set_error_message(context, ENOTSUP,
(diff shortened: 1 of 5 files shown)


nicowilliams:
We should add a test that it is possible to iterate the contents of the HDB while performing write operations when using the LMDB or SQLite3 backends.

lib/hdb/hdb.c
```diff
@@ -693,47 +740,67 @@ hdb_create(krb5_context context, HDB **db, const char *filename)
...
          */
-        (*db)->virtual_hostbased_princ_svcs =
-            krb5_config_get_strings(context, NULL, "hdb",
-                                    "virtual_hostbased_princ_svcs", NULL);
-        /* Check for ENOMEM */
-        if ((*db)->virtual_hostbased_princ_svcs == NULL
-            && krb5_config_get_string(context, NULL, "hdb",
-                                      "virtual_hostbased_princ_svcs", NULL)) {
-            (*db)->hdb_destroy(context, *db);
+        if (!is_pathish(filename)) {
```
nicowilliams, line 757: Whoops, use the `pathish` variable.

lib/hdb/hdb.c
```diff
@@ -693,47 +740,67 @@ hdb_create(krb5_context context, HDB **db, const char *filename)
...
+    if (cb_ctx.h == NULL || cb_ctx.h->prefix == NULL)
+        cb_ctx.h = &default_dbmethod;
+    if (cb_ctx.h == NULL || cb_ctx.h->prefix == NULL) {
+        krb5_set_error_message(context, ENOTSUP,
+                               "Could not determine default DB backend for %s",
+                               filename);
+        return ENOTSUP;
+    }
+    if (!*db)
+        ret = (*cb_ctx.h->create)(context, db, cb_ctx.residual);
```
nicowilliams, line 797: Add comment?

vdukhovni (approved):

PR #758, opened 2020-10-06
hdb: Test LMDB and SQLite3 HDB concurrency

nicowilliams:
Test that we can dump an HDB w/o blocking write transactions.

This currently works for different processes (both backends), but not
for threads (for either backend), and it's not yet clear why.


hyc:
When you say "not working" what's actually happening? deadlock? invalid data?

PR #760, opened 2020-10-08
gss-token(1) improvements and fixes.

elric1:
We present a series of improvements and fixes to gss-token(1). We clean up and restructure the code a bit. We re-implement a couple of flags that were forgot when it was imported and changed to use getargs(). We make the code implement a proper GSS loop rather than assuming that there is only a single input and output token. We provide a mechanism to split up tokens to test new functionality that is being developed. And finally, we implement a way to ask GSS to use a particular mech on the initiator side.

### DIFF
--- a/lib/gssapi/gss-token.1
+++ b/lib/gssapi/gss-token.1
@@ -15,6 +15,7 @@
 .Fl r
 .Op Fl MNln
 .Op Fl C Ar ccache
+.Op Fl S Ar maxsize
 .Op Fl c count
 .Op Ar service@host
 .Sh DESCRIPTION
@@ -52,6 +53,11 @@ This can be used to load test the KDC.
 prepend
 .Dq Negotiate\ 
 to generated tokens and expect it on consumed tokens.
+.It Fl S Ar maxsize
+split each token that is generated into components of maximum
+size
+.Ar maxsize .
+Each token is base64 encoded and output separately.
 .It Fl c Ar count
 repeat the operation
 .Ar count
--- a/lib/gssapi/gss-token.1
+++ b/lib/gssapi/gss-token.1
@@ -20,7 +20,9 @@
 .Sh DESCRIPTION
 .Nm
 generates and consumes base64 encoded GSS tokens.
-It is mostly useful for testing.
+By default, it runs as an initiator and with the
+.Fl r
+flag it becomes an acceptor.
 .Pp
 .Nm
 supports the following options:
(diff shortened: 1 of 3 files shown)


lib/gssapi/gss-token.1
```diff
@@ -50,26 +54,54 @@ This can be used to load test the KDC.
...
-loop infinitely in read mode.
-This is to support a multiple round trip GSS mechanism.
+loop infinitely in acceptor mode.
+.It Fl m Ar mech
+specifies the GSS mechanism that will be used in initiator mode.
+If
+.Do ? Dc
+is specified, a list of supported mechanisms will be output and
+.Nm
+will exit.
```
vdukhovni, line 76:
```suggestion
If a mechanism name of
.Do ? Dc
is specified, a list of supported mechanisms will be output and
.Nm
will exit.
```

lib/gssapi/gss-token.1
```diff
@@ -50,26 +54,54 @@ This can be used to load test the KDC.
...
+and will output a token whenever the GSS mechanism provides one.
+Tokens are base64 encoded and terminated by either two successive
+newlines or one newline and EOF.
+The base64 encoding may be broken up by single newlines which will
+be ignored when read.  No extra whitespace will be ignored.
+.Sh EXAMPLES
+To test a simple GSS mechanism which doesn't require a round trip,
+a single
+.Pa /bin/sh
+pipeline will suffice:
```
vdukhovni, line 101: Should there be mention here of a keytab requirement on the acceptor side?

lib/gssapi/gss-token.c
```diff
@@ -164,45 +166,186 @@ gss_mk_err(OM_uint32 maj_stat, OM_uint32 min_stat, const char *preamble)
...
+	int		 first = 1;
+
+	if (nflag)
+		goto bail;
+
+	/*
+	 * According to RFC 2744 page 25, we simply don't output
+	 * zero length output tokens.
+	 */
+	if (out->length == 0)
```
vdukhovni, line 224:
```suggestion
	if (len == 0)
```

lib/gssapi/gss-token.c
```diff
@@ -164,45 +166,186 @@ gss_mk_err(OM_uint32 maj_stat, OM_uint32 min_stat, const char *preamble)
...
+		return -1;
+	}
+	ret = rk_base64_decode(tmp, in->value);
+	if (ret < 0) {
+		free(in->value);
+		if (errno == EOVERFLOW)
+			fprintf(stderr, "Token is too big\n");
+		else
+			fprintf(stderr, "Token encoding is not valid "
+			    "base64\n");
```
vdukhovni, line 295: Should these exit?  Set `in->value` back to NULL? ...

lib/gssapi/gss-token.c
```diff
@@ -529,6 +619,14 @@ main(int argc, char **argv)
 	argc -= optidx;
 	argv += optidx;
 
+	if (mech) {
+		if (mech[0] == '?' && mech[1] == '\0') {
+			print_all_mechs();
+			exit(0);
+		}
+		global_mech = gss_name_to_oid(mech);
```
vdukhovni, line 627: What happens when the lookup fails?

lib/gssapi/gss-token.c
elric1: Thanks for pointing this out. I noticed a few more problems in this code, actually. I'll push updates.

vdukhovni (approved):
Looks ok.

PR #767, opened 2020-12-29
rework PAC and AD-SIGNTICKET for S4U2Proxy support

iboukris:


kdc/kerberos5.c
```diff
@@ -2588,6 +2572,12 @@ _kdc_tkt_add_if_relevant_ad(krb5_context context,
 	    krb5_set_error_message(context, ret, "add AuthorizationData failed");
 	    return ret;
 	}
+
+	ade = tkt->authorization_data->val[tkt->authorization_data->len-1];
+	for (i = 0; i < tkt->authorization_data->len-1; i++) {
+		tkt->authorization_data->val[i+1] = tkt->authorization_data->val[i];
```
lhoward, line 2578: picky but `i + 1` (whitespace)

kdc/kerberos5.c
```diff
@@ -2588,6 +2572,12 @@ _kdc_tkt_add_if_relevant_ad(krb5_context context,
 	    krb5_set_error_message(context, ret, "add AuthorizationData failed");
 	    return ret;
 	}
+
+	ade = tkt->authorization_data->val[tkt->authorization_data->len-1];
```
lhoward, line 2576: this always makes me nervous, haven't looked at the full code path but perhaps an assert that `tkt->authorization_data->len != 0` could be nice

iboukris: Hi Like :)
The code just above adds one wrapped AD to the list, but sure we can assert it, I'll revise it.

lhoward: Maybe I'm being needlessly paranoid :) (Given there are probably a few overflow bugs lurking elsewhere as a @abartlet has mentioned)

iboukris:
I realized I can take some tests from PR #599, the second commit in there regarding AS-REQ may still be needed, although without ticket-signature it would still be unusable as evidence ticket in constrained-delegation, so it matters less.

kdc/krb5tgs.c
```diff
@@ -284,85 +89,64 @@ check_PAC(krb5_context context,
...
+	    ret = krb5_pac_verify(context, pac, 0, NULL, NULL,
+				  krbtgt_check_key);
+	    if (ret) {
 		krb5_pac_free(context, pac);
-
 		return ret;
 	    }
 	}
-	free_AuthorizationData(&child);
+	/* Discard the PAC if the pluging didn't handle it */
```
RobCrowston, line 137: "plugin", not pluging

kdc/krb5tgs.c
```diff
@@ -2433,36 +2154,14 @@ tgs_build_reply(astgs_request_t priv,
...
-	if (!ad_signedpath) {
+	if (mspac == NULL || !ad_signedticket) {
 	    ret = KRB5KDC_ERR_BADOPTION;
 	    kdc_log(context, config, 4,
-		    "Ticket not signed with PAC nor SignedPath service %s failed "
+		    "Ticket not signed with PAC, service %s failed "
 		    "for delegation to %s for client %s (%s)"
-		    "from %s",
-		    spn, tpn, dpn, cpn, from);
+		    "from %s,  mspac(%p), ad_signedticket(%s).",
```
RobCrowston, line 2162: want to leave this as a pointer? (%p)

kdc/krb5tgs.c
iboukris: Oh thanks, I had it right in the commit message ;)

kdc/krb5tgs.c
iboukris: Yeah, I just want to convey is it's NULL, I'll rework this log.

iboukris:
I think this is ready for review, thanks!

kdc/krb5tgs.c
```diff
@@ -59,85 +59,64 @@ check_PAC(krb5_context context,
...
+    /* Verify the KDC signatures. */
+    ret = _kdc_pac_verify(context, client_principal, delegated_proxy_principal,
+			  client, server, krbtgt, &pac);
+    if (ret == KRB5_PLUGIN_NO_HANDLE) {
+	/*
+	 * We can't verify the KDC signatures if the ticket was issued by
+	 * another realm's KDC.
+	 */
+	if (krb5_realm_compare(context, server->entry.principal,
+			       ticket_server->entry.principal)) {
```
abartlet, line 97: I'm concerned by how this might work or not on Samba where we, due to that silly NETBIOS realm thing, and due to allowing the wrong case etc, essentially support multiple realms.  

I guess it doesn't matter if Samba always handles it fully in KRB5_PLUGIN_NO_HANDLE.

kdc/krb5tgs.c
```diff
@@ -2229,6 +2215,20 @@ tgs_build_reply(astgs_request_t priv,
...
+     * Currently Heimdal only allow renewal of krbtgt anyway but that might
+     * change one day (see issue #763) so make sure to check for it.
+     */
+
+    if (kdc_issued &&
+	!krb5_principal_is_krbtgt(context, server->entry.principal))
+	add_ticket_sig = TRUE;
+
+    rodc_id = krbtgt_out->entry.kvno >> 16;
+
```
abartlet, line 2231: Does this match the MS behaviour?   @josephsutton1 do you know?

kdc/krb5tgs.c
iboukris: Yeah, samba should handle it in the plugin (although as i mentioned, that could have worked for samba too, as samba should always canonicalize the server realm).

kdc/krb5tgs.c
iboukris: As it says, heimdal doesn't allow renewal of non-tgt but Windows does (can be tested with MIT's kinit -R, when the cache only contains a service ticket for instance). The aim is in the future when service-ticket renewal is made to work in Heimdal, that we match Windows behavior, which is afair; is the ticket had no PAC renew without PAC, if it had a PAC but no ticket-signature (like currently tickets issued by Windows via AS request, kinit -S host/xyz), renew the ticket and keep the PAC but do *not* add a ticket signature (the PAC can be unchanged), if it had ticket-signature then renew the ticket and regenerate all three signatures.

lhoward: Probably not relevant but may be interesting: Windows won't convert a RODC TGT to a full TGT if the renew flag is set. I noticed this during [TktBridgeAP](http://github.com/PADL/TktBridgeAP) development.

lhoward:
Just a general comment, everyone agrees that removing KRB5SignedPath won't be an interoperability issue? I know we probably have been through this before but, we should document what the issues are, e.g. do all KDCs in a realm need to be upgraded at once, is this only relevant if they were issuing authorisation data, etc.

kdc/windc_plugin.h
```diff
@@ -43,8 +43,9 @@
  * krb5_pac_init and fill in the PAC structure for the principal using
  * krb5_pac_add_buffer.
  *
- * The PAC verify function should verify all components in the PAC
- * using krb5_pac_get_types and krb5_pac_get_buffer for all types.
+ * The PAC verify function should verify the PAC KDC signature by fetching the
```
lhoward, line 46: Picky, but reformat line wrap please :)

lib/krb5/authdata.c
```diff
@@ -0,0 +1,121 @@
...
+			    int type,
+			    const krb5_data *data)
+{
+    krb5_error_code ret;
+    size_t size = 0;
+
+    if (tkt->authorization_data == NULL) {
+	tkt->authorization_data = calloc(1, sizeof(*tkt->authorization_data));
+	if (tkt->authorization_data == NULL) {
+	    krb5_set_error_message(context, ENOMEM, "out of memory");
```
lhoward, line 54: You can use `krb5_enomem()` for this, saves typing.

lib/krb5/authdata.c
```diff
@@ -0,0 +1,121 @@
...
+	if (ret) {
+	    krb5_set_error_message(context, ret, "add AuthorizationData failed");
+	    return ret;
+	}
+    }
+
+    return 0;
+}
+
+/* Insert a PAC wrapped in AD-IF-RELEVANT container as the first AD element */
```
lhoward, line 98: Is adding it as a first element something necessary for Windows interop? Perhaps a note to this effect would be useful, as it seems unusual (as a casual reader) to not insert it at the end.

lib/krb5/pac.c
```diff
@@ -911,6 +932,21 @@ krb5_pac_verify(krb5_context context,
 			      privsvr);
 	if (ret)
 	    return ret;
+
+	if (pac->ticket_sign_data.length != 0) {
```
lhoward, line 936: Remind me, I'm not across the latest protocol revisions, but we're going to require a ticket checksum always, not just when S4U2Proxy is present, because this is the moral replacement for KRB5SignedPath?

lib/krb5/pac.c
```diff
@@ -1024,6 +1061,16 @@ _krb5_pac_sign(krb5_context context,
 				       N_("PAC have two logon names", ""));
 		goto out;
 	    }
+	} else if (p->pac->buffers[i].type == PAC_TICKET_CHECKSUM) {
+	    if (p->ticket_checksum == NULL) {
+		p->ticket_checksum = &p->pac->buffers[i];
+	    }
+	    if (p->ticket_checksum != &p->pac->buffers[i]) {
+		ret = EINVAL;
```
lhoward, line 1069: Wonder if there is a more Kerberos-specific error than EINVAL, eg. `KRB5KDC_ERR_BADOPTION` or `KRB5KRB_APERR_BAD_INTEGRITY`.

lib/krb5/pac.c
```diff
@@ -1201,3 +1306,223 @@ _krb5_pac_sign(krb5_context context,
...
+    size_t cksumsize, prefix;
+    uint32_t type = 0;
+
+    *cstype = 0;
+    *rodc_id = 0;
+
+    sig = pac->privsvr_checksum;
+    if (sig == NULL) {
+	krb5_set_error_message(context, EINVAL, "PAC missing kdc checksum");
+	return EINVAL;
```
lhoward, line 1328: Ditto re: error code

lib/krb5/pac.c
```diff
@@ -1201,3 +1306,223 @@ _krb5_pac_sign(krb5_context context,
...
+    sp = krb5_storage_from_mem((char *)pac->data.data + sig->offset_lo,
+			       sig->buffersize);
+    if (sp == NULL)
+	return krb5_enomem(context);
+
+    krb5_storage_set_flags(sp, KRB5_STORAGE_BYTEORDER_LE);
+
+    ret = krb5_ret_uint32(sp, &type);
+    if (ret) {
+	krb5_storage_free(sp);
```
lhoward, line 1340: Maybe a cleanup handler here? It's safe to call `krb5_storage_free(NULL)`.

lib/krb5/pac.c
```diff
@@ -1201,3 +1306,223 @@ _krb5_pac_sign(krb5_context context,
...
+	return ret;
+    }
+
+    *cstype = type;
+
+    prefix = krb5_storage_seek(sp, 0, SEEK_CUR);
+
+    if (cksumsize == (sig->buffersize - prefix)) {
```
lhoward, line 1354: Maybe instead of an early return, check for `cksumsize > (sig->buffersize - prefix)` and handle the RODC path in a new block (that checks for `cksumsize + 2 >= (sig->buffersize - prefix)` in case it ever gets further expanded).

lib/krb5/pac.c
```diff
@@ -1201,3 +1306,223 @@ _krb5_pac_sign(krb5_context context,
...
+				     pac_data.length,
+				     &pac);
+		if (ret) {
+		    free_AuthorizationData(&child);
+		    return ret;
+		}
+
+		if (pac->ticket_checksum == NULL) {
+		    free_AuthorizationData(&child);
+
```
lhoward, line 1428: could remove this newline

lib/krb5/pac.c
```diff
@@ -1201,3 +1306,223 @@ _krb5_pac_sign(krb5_context context,
...
+    *signedticket = FALSE;
+    *ppac = NULL;
+
+    if (ad == NULL || ad->len == 0)
+	return 0;
+
+    for (i = 0; i < ad->len; i++) {
+	AuthorizationData child;
+
+	if (ad->val[i].ad_type != KRB5_AUTHDATA_IF_RELEVANT)
```
lhoward, line 1395: I wonder if we should check if the PAC ever appears outside AD-IF-RELEVANT.

lhoward: And return an error.

lib/krb5/pac.c
```diff
@@ -1201,3 +1306,223 @@ _krb5_pac_sign(krb5_context context,
...
+		    free_AuthorizationData(&child);
+
+		    *ppac = pac;
+		    return 0;
+		}
+
+		child.val[j].ad_data.length = 1;
+		child.val[j].ad_data.data = &single_zero;
+
+		ASN1_MALLOC_ENCODE(AuthorizationData, recoded_adifr.data,
```
lhoward, line 1436: Maybe a little comment saying, re-encoded the ticket with the PAC replaced with a single NUL byte, to be used as input to the ticket signature.

lib/krb5/pac.c
```diff
@@ -1201,3 +1306,223 @@ _krb5_pac_sign(krb5_context context,
...
+	    krb5_abortx(context, "Internal error in ASN.1 encoder");
+	if (ret)
+	    return ret;
+
+	ret = remove_AuthorizationData(tkt->authorization_data, 0);
+	if (ret) {
+	    krb5_data_free(&tkt_data);
+	    return ret;
+	}
+
```
lhoward, line 1518: indentation

lhoward (changes requested):
LGTM, well done! Perhaps because of the sensitivity of these code paths it's worth getting @nicowilliams or @jaltman to review too.

lib/krb5/authdata.c
abartlet: I would prefer it if possible, given memories of early Samba that assumed the PAC needed to be the first one.

lhoward: I think Windows too, I managed to make W2K crash many years ago putting an experimental NFSv4 PAC in the first element. But, a note to that effect would be useful.

lib/krb5/pac.c
```diff
@@ -911,6 +932,21 @@ krb5_pac_verify(krb5_context context,
...
+	if (pac->ticket_sign_data.length != 0) {
+
+	    if (pac->ticket_checksum == NULL) {
+		krb5_set_error_message(context, EINVAL,
+				       "PAC missing ticket-checksum");
+		return EINVAL;
+	    }
+
+	    ret = verify_checksum(context, pac->ticket_checksum, &pac->data,
```
lhoward, line 944: BTW do we want to defense in depth and also check for keyed checksums in `create_checksum()`?

iboukris:
> Just a general comment, everyone agrees that removing KRB5SignedPath won't be an interoperability issue? I know we probably have been through this before but, we should document what the issues are, e.g. do all KDCs in a realm need to be upgraded at once, is this only relevant if they were issuing authorisation data, etc.

So iirc when we last discussed it with @greghudson we noticed that MIT and Heimdal do not currently interoperate anyway as the asn1 of KRB5SignedPath differs, while by moving to PAC we'd be able to interoperate with both MIT and Windows.

For S4U2Proxy request to work on an updated KDC the evidence ticket should be issued by an updated KDC as well, i'll add a note to the commit message.

lib/krb5/pac.c
abartlet: The defence in depth I really want is to remove unkeyed checksums as 'weak', but until then I would support such a change.

lib/krb5/pac.c
iboukris: Well, what's always? When else does a KDC get to look at a service ticket except S4U2Proxy and renewal/validation? We could make other uses of that, but I'm not aware of existing use case.

lib/krb5/pac.c
iboukris: Ins't it checked elsewhere? I'll look.
BTW, can't we get rid of unkeyed checksums altogether already? Wasn't it only used for DES enctypes?

lib/krb5/authdata.c
iboukris: Yeah, definitely Windows as well, tested with 2019 server some while ago. I'll expand the comment.

lhoward:
> > Just a general comment, everyone agrees that removing KRB5SignedPath won't be an interoperability issue? I know we probably have been through this before but, we should document what the issues are, e.g. do all KDCs in a realm need to be upgraded at once, is this only relevant if they were issuing authorisation data, etc.
> 
> So iirc when we last discussed it with @greghudson we noticed that MIT and Heimdal do not currently interoperate anyway as the asn1 of KRB5SignedPath differs, while by moving to PAC we'd be able to interoperate with both MIT and Windows.
> 
> For S4U2Proxy request to work on an updated KDC the evidence ticket should be issued by an updated KDC as well, i'll add a note to the commit message.

I guess, MIT and Heimdal KDCs were less likely to be expected to interoperate within the same realm, but Heimdal KDCs may be. I don't think you need to change anything, just perhaps a note, I'm sure there are other breaking changes (like the FAST cookie key derivation coming up in the `lukeh/gss-preauth-apple` branch).

lib/krb5/pac.c
iboukris: @abartlet I didn't see your comment while writing mine, but I'm not surprise we think alike on this :)

lib/krb5/pac.c
abartlet: user-to-user?

lib/krb5/pac.c
lhoward: Unkeyed SHA checksums are used by PKINIT (`build_auth_pack()`) and GSS pre-auth `pa_gss_decode_context_state()` so they can't be gotten rid of entirely. Let me discuss this off-list.

lib/krb5/pac.c
abartlet: (sorry, that is a TGT, I just still get confused by this feature)

lib/krb5/pac.c
iboukris: I'll go with KRB5KDC_ERR_BADOPTION as integrity sounds like checksum failure to me.

lib/krb5/pac.c
iboukris: Ok,I guess you meant '(sig->buffersize - prefix) >= cksumsize + 2' though.

lib/krb5/pac.c
iboukris: Done.

lib/krb5/pac.c
iboukris: Leaving this out given your related, more generic work.

iboukris:
I've added a note about interoperability in the commit message and I think I've addressed all the comments, I've also added a check for duplicate PACs while at it, please take a look.

lhoward:
Ready to merge? If so I'll just clean up some indentation and commit messages etc and do it manually.

iboukris:
Thanks Luke and all!

iboukris:
Hi @lhoward I noticed with my last changes _krb5_kdc_pac_ticket_parse() would rely on the caller to free the pac in case of an error, I don't like it so I'll try switch to a cleanup handler, I'll submit a PR later today hopefully.

lhoward:
Sure thing.

PR #771, opened 2021-03-17
autoconf-2.70 fix

Polynomial-C:
autoconf-2.70 and newer are more strict with quoting etc. and thus generate
a broken configure file:
```
  configure: 20855: Syntax error: ")" unexpected (expecting "fi")
```
Gentoo-bug: https://bugs.gentoo.org/776241
Signed-off-by: Lars Wendler <polynomial-c@gentoo.org>


Polynomial-C:
I have removed the changes to `AS_IF` because this code is inside of `AC_DEFUN`. The remaining change still is necessary to get a working `configure` file with autoconf-2.70 or newer.

jaltman:
Thank you.

PR #779, opened 2021-06-15
kcm: Fix SEGV in kcm_get_cache_next

fihuer:
The 'KCM_OP_GET_CACHE_BY_UUID' op may return 'KRB5_FCC_NOFILE' if
user access gets refused by KCM access control.
In this case, kcm_get_cache_next ignores it and may trigger a NULL pointer dereference
in "krb5_ret_stringz".

This patch fixes it by catching it like a 'KRB5_CC_END' error (no more ccaches).

### DIFF
--- a/lib/krb5/kcm.c
+++ b/lib/krb5/kcm.c
@@ -900,7 +900,7 @@ kcm_get_cache_next(krb5_context context, krb5_cc_cursor cursor, const krb5_cc_op
 
     ret = krb5_kcm_call(context, request, &response, &response_data);
     krb5_storage_free(request);
-    if (ret == KRB5_CC_END)
+    if (ret == KRB5_CC_END || ret == KRB5_FCC_NOFILE)
 	goto again;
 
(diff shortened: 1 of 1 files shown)


lhoward:
Hmm, yes, this code is a bit suspect but I am not sure this the right solution. I suspect the logic should be:

```c
    ret = krb5_kcm_call(context, request, &response, &response_data);
    krb5_storage_free(request);
    if (ret == KRB5_CC_END) {
        goto again;
    } else if (ret != 0) {
        return ret;
    }
```

lhoward:
I'll see if Apple ever fixed this...

lhoward:
Yes, the current Apple code returns directly if `ret != 0`, there is no need to free the response data as it will not be allocated on failure. I will apply Apple's patch now.

lhoward:
Fixed in 0452d8c601e599ba741b9560c4ac84fa7752e7fb

PR #782, opened 2021-06-24
Move out: label to inside #endif of HAVE_DLOPEN

abartlet:
Samba compiles Heimdal internally without HAVE_DLOPEN to keep
to internally supplied mechanisms and plugins.

Samba compiles with strict warning flags and on Ubuntu 20.04
with gcc version 9.3.0 (Ubuntu 9.3.0-17ubuntu1~20.04) we see:

../../source4/heimdal/lib/gssapi/mech/gss_mech_switch.c: In function ‘_gss_load_mech’:
../../source4/heimdal/lib/gssapi/mech/gss_mech_switch.c:462:1: error: label ‘out’ defined but not used [-Werror=unused-label]
  462 | out:
      | ^~~

### DIFF
--- a/lib/gssapi/mech/gss_mech_switch.c
+++ b/lib/gssapi/mech/gss_mech_switch.c
@@ -457,9 +457,10 @@ _gss_load_mech(void)
 		continue;
 	}
 	fclose(fp);
-#endif
 
 out:
+
+#endif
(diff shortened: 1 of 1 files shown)


lhoward:
LGTM

PR #780, opened 2021-06-22
Provide the correct principal name to verify_flags() for user2user tickets

abartlet:
This is a patch to help Samba merge in a current upstrem Heimdal and is selected out of a larger patch by @metze-samba 

It looks like this codepath is tested by server_uu / check-uu but I've not manually checked it.

When b->kdc_options.enc_tkt_in_skey is set, the variable 'spn'
provided to verify_flags() for the purpose of logging is not
yet set, it is still NULL from the initialiser.

On Samba we see, when compiling with -O3 and strict warning rules with
gcc version 9.3.0 (Ubuntu 9.3.0-17ubuntu1~20.04):

$ make -k
PYTHONHASHSEED=1 WAF_MAKE=1  ./buildtools/bin/waf build
Waf: Entering directory `/home/abartlet/heimdal-import/samba/bin/default'
	Selected embedded Heimdal build
[2822/4527] Compiling source4/heimdal/kdc/krb5tgs.c
In function ‘verify_flags’,
    inlined from ‘verify_flags’ at ../../source4/heimdal/kdc/krb5tgs.c:625:1,
    inlined from ‘tgs_build_reply.isra.0’ at ../../source4/heimdal/kdc/krb5tgs.c:1671:8:
../../source4/heimdal/kdc/krb5tgs.c:635:2: error: ‘%s’ directive argument is null [-Werror=format-overflow=]
  635 |  kdc_log(context, config, 4, "Ticket not valid (%s)", pstr);
      |  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In function ‘verify_flags’,
    inlined from ‘tgs_build_reply.isra.0’ at ../../source4/heimdal/kdc/krb5tgs.c:1671:8:
../../source4/heimdal/kdc/krb5tgs.c:631:2: error: ‘%s’ directive argument is null [-Werror=format-overflow=]
  631 |  kdc_log(context, config, 4, "Ticket expired (%s)", pstr);
      |  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors

Essentially, with an optimising compiler, the funciton is inlined
and the compiler notices that the kdc_log() argument is always NULL.

Selected from patches by Stefan Metzmacher <metze@samba.org>
in his master-heimdal branch at:
[...]

PR #783, opened 2021-06-30
Fix user2user principal (again)

abartlet:
73debbc1668adb4080165fcc1a7028ecae36e9b7 accidentially reverted
the critical part of 040a093654bfbb602ef0ab9807f51a5c83cf9596,
presumably during conflict resolution.

Found by a strict Samba compile during import of current Heimdal.

See #780


lhoward:
Thanks :)

PR #785, opened 2021-06-30
Initialise err in hdb_create()

abartlet:
On Ubuntu 20.04 with gcc version 9.3.0 during a strict
Samba build with -Werror=maybe-uninitialized we get:

../../source4/heimdal/lib/hdb/hdb.c: In function ‘hdb_create’:
../../source4/heimdal/lib/hdb/hdb.c:831:13: error: ‘ret’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
  831 |     if (*db && ret == 0)
      |         ~~~~^~~~~~~~~~~

### DIFF
--- a/lib/hdb/hdb.c
+++ b/lib/hdb/hdb.c
@@ -737,7 +737,7 @@ load_config(krb5_context context, HDB *db)
 krb5_error_code
 hdb_create(krb5_context context, HDB **db, const char *filename)
 {
-    krb5_error_code ret;
+    krb5_error_code ret = ENOTSUP;
     struct cb_s cb_ctx;
 
     *db = NULL;


lhoward:
Generally Heimdal prefers not to initialise at the start of the function, as it can mask the error not being set elsewhere but, this function is such a mess that your patch is the least worst option. Applying.

abartlet:
Thanks.  I also couldn't work out what a better fix would be, and couldn't follow the flow of what we would watch for with *db either.

lhoward:
That function is a bit WORN (write once, read never) but I guess it evolved over time. I'd rather not be responsible for changing it and potentially introducing a regression.

abartlet:
I can't see this merged manually, was this really merged?

PR #787, opened 2021-06-30
Allow KDC to always return the salt in the PA-ETYPE-INFO[2]

abartlet:
This is to match the Windows KDC behaviour, optionally.

Samba will use this mode, as Samba has tests that show
the difference and is keen for bug-for-bug/byte-for-byte
rather than strict RFC compliance where there is a
reasonable choice.

The behaviour was changed in:

commit de1f37a6aadb35fdce3254528d89bdac23fb120b
Author: Luke Howard <lukeh@padl.com>
Date:   Mon Jan 7 15:45:36 2019 +1100

    kdc: omit default salt from PA-ETYPE-INFO[2]

    If the salt for the AS-REP client key matches the default password salt for the
    client principal in the AS-REQ, then it can be omitted from the PA-ETYPE-INFO,
    PA-ETYPE-INFO2 (RFC4120) as the client will assume the default salt in its
    absence.

### DIFF
--- a/kdc/kdc.h
+++ b/kdc/kdc.h
@@ -62,6 +62,14 @@ typedef struct krb5_kdc_configuration {
 
     krb5_boolean encode_as_rep_as_tgs_rep; /* bug compatibility */
 
+    /*
+     * Windows 2019 (and earlier versions) always sends the salt
+     * and Samba has testsuites that check this behaviour, so a
+     * Samba AD DC will set this flag to match the AS-REP packet
+     * exactly.
+     */
+    krb5_boolean force_include_pa_etype_salt;
+
     krb5_boolean tgt_use_strongest_session_key;
     krb5_boolean preauth_use_strongest_session_key;
     krb5_boolean svc_use_strongest_session_key;
--- a/kdc/kerberos5.c
+++ b/kdc/kerberos5.c
@@ -1404,6 +1404,15 @@ get_pa_etype_info_both(krb5_context context,
     krb5_error_code ret;
 
     /*
+     * Windows 2019 (and earlier versions) always sends the salt
+     * and Samba has testsuites that check this behaviour, so a
+     * Samba AD DC will set this flag to match the AS-REP packet
+     * more closely.
+     */
+    if (config->force_include_pa_etype_salt)
+	include_salt = TRUE;
+
+    /*
      * RFC4120 requires:
      *   When the AS server is to include pre-authentication data in a
      *   KRB-ERROR or in an AS-REP, it MUST use PA-ETYPE-INFO2, not


lhoward:
I understand the desire for bug-for-bug compatibility (I went there once, too!) but, now that Windows protocols are documented it should be possible to find out whether Windows actually depends on this behaviour. If it does not, then I would prefer to omit this patch and I imagine @jaltman / @nicowilliams might feel similarly.

abartlet:
That's fine, we are very unlikely to ever use an exact Heimdal release so I'll just add this to the list of patches we carry in lorikeet-heimdal (currently https://gitlab.com/samba-team/devel/lorikeet-heimdal/-/commits/lorikeet-heimdal-202106300341) and mark it as not-for-upstream so we don't bother you.

However, I'll note that it isn't just windows clients that we are concerned with, there is a wide world of Kerberos clients out there and if our testsuite can tell the difference, and we can mitigate it, we (Samba) need to.

Also, while the MS docs are good, they don't say very much about the internal Kerberos processing on Windows, and while details can be clarified, there are limits and we would prefer not to need to ask if we can just be byte-for-byte.

abartlet:
While probably not triggered by this, we have had strange behaviour like 'bad password count updated twice as fast as expected' from Windows in the past when it didn't like some of our AS-REP error replies, and came back for an extra go.  Due to the time that kind of thing takes to debug, we would prefer to be byte-for-byte.

lhoward:
> While probably not triggered by this, we have had strange behaviour like 'bad password count updated twice as fast as expected' from Windows in the past when it didn't like some of our AS-REP error replies, and came back for an extra go. Due to the time that kind of thing takes to debug, we would prefer to be byte-for-byte.

Do you send `KERB-EXT-ERROR`?

abartlet:
Production Samba is still on 2011 code, it was around the ETYPE2 and ENC_TIMESTAMP replies. 

The details are on https://bugzilla.samba.org/show_bug.cgi?id=11539

Enough time was sunk there that I'm very keen to avoid a repeat.  That doesn't mean you need to take our patches however, but getting to run against pure upstream would be nice some day (perhaps in a mode where the NETBIOS realm stuff isn't working, but otherwise)

kdc/kdc.h
```diff
@@ -67,6 +67,8 @@ typedef struct krb5_kdc_configuration {
     krb5_boolean svc_use_strongest_session_key;
     krb5_boolean use_strongest_server_key;
 
+    krb5_boolean always_include_salt;
```
lhoward, line 70: I would move this to directly under `encode_as_rep_as_tgs_rep` and add a similar bug compatibility comment

kdc/kerberos5.c
```diff
@@ -1347,6 +1347,10 @@ get_pa_etype_info2(krb5_context context,
     if(pa.val == NULL)
 	return ENOMEM;
 
+    if (config->always_include_salt) {
+	include_salt = TRUE;
+    }
```
lhoward, line 1352: no braces necessary

lhoward (changes requested):

lhoward:
BTW, I assume there's no explicit krb5.conf configuration option for this because you set the configuration directly?

lhoward:
Also, the PR name doesn't match the behaviour, i.e. you're only including default salt in PA-ETYPE-INFO2, please either move the check to `get_pa_etype_info_both()` or change the configuration variable name to indicate it only applies to the INFO2. Suggest the former would be cleaner.

lhoward:
And maybe change it to `force_include_pa_etype_salt` or something

abartlet:
> BTW, I assume there's no explicit krb5.conf configuration option for this because you set the configuration directly?

Yes, we set it at runtime by supplying the configuration directly into the KDC which is called as a library.

kdc/kdc.h
abartlet: Thanks, will do

kdc/kerberos5.c
abartlet: Sorry about that, Samba coding conventions creeping in.

PR #788, opened 2021-06-30
Align locked out account behaviour with Windows

abartlet:
Windows does not check the password on an account that has been locked.

Heimdal does not implement locked_out, however the Samba hdb
backend does, and needs this checked before passwords (for bad
password lockout), not after in kdc_check_access().

Signed-off-by: Gary Lockyer <gary@catalyst.net.nz>

### DIFF
--- a/kdc/kerberos5.c
+++ b/kdc/kerberos5.c
@@ -580,6 +580,13 @@ pa_enc_chal_validate(astgs_request_t r, const PA_DATA *pa)
 	return ret;
     }
 
+    if (r->client->entry.flags.locked_out) {
+       ret = KRB5KDC_ERR_CLIENT_REVOKED;
+       kdc_log(r->context, r->config, 0,
+               "Client (%s) is locked out", r->cname);
+       return ret;
+    }
+
     ret = decode_EncryptedData(pa->padata_value.data,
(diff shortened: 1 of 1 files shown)


lhoward:
See `_kdc_check_access()`. If you have a WinDC plugin then you need to implement this logic inside there.

lhoward:
I've applied fdf5b402021c5fc141f51a16891f5a3f34bd62ff to correct the error codes, but I am going to close this otherwise, because we shouldn't be adding revocation checks to individual pre-authentication mechanisms (you've missed PKINIT, for example).

If there is no WinDC plugin, then the revocation error will be returned by `kdc_check_flags()`. If there is a WinDC plugin, then `kdc_check_flags()` will not be called by `_kdc_check_access()` and the plugin must implement this logic itself. If you think this is an error and `kdc_check_flags()` should be called unconditionally, please submit another PR.

abartlet:
The problem was the `_kdc_check_access()` is too late, being after the password is checked.  PKINIT isn't checked as it isn't password based, we have chosen to read `locked_out` as being _locked out for having tried to many passwords_.  Is that clearer?

lhoward:
OK, fair enough.

PR #797, opened 2021-08-04
kuser: --fast-armor-anon-pkinit option for kinit

lhoward:
Add the ` --pk-anon-fast-armor` option to kinit, which acquires a temporary anonymous PKINIT TGT to use as a FAST armor key.  The KDC certificate is not authenticated, per RFC 6113 5.4.1.1.

### DIFF
--- a/lib/hx509/hx509.h
+++ b/lib/hx509/hx509.h
@@ -182,6 +182,9 @@ typedef enum {
 #define HX509_CMS_VS_ALLOW_ZERO_SIGNER			0x04
 #define HX509_CMS_VS_NO_VALIDATE			0x08
 
+/* flags from hx509_cms_verify_signed_ext (out verify_flags) */
+#define HX509_CMS_VSE_VALIDATED				0x01
+
 /* selectors passed to hx509_crypto_select and hx509_crypto_available */
 #define HX509_SELECT_ALL 0
 #define HX509_SELECT_DIGEST 1
--- a/lib/krb5/pkinit.c
+++ b/lib/krb5/pkinit.c
@@ -787,7 +787,7 @@ _krb5_pk_mk_padata(krb5_context context,
 				     NULL);
     if (ic_flags & KRB5_INIT_CREDS_NO_C_NO_EKU_CHECK)
 	ctx->require_eku = 0;
-    if (ctx->id->flags & PKINIT_BTMM)
+    if (ctx->id->flags & (PKINIT_BTMM | PKINIT_NO_KDC_ANCHOR))
 	ctx->require_eku = 0;
 
     ctx->require_krbtgt_otherName =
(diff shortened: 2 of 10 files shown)


lib/hx509/hx509.h
```diff
@@ -182,6 +182,9 @@ typedef enum {
 #define HX509_CMS_VS_ALLOW_ZERO_SIGNER			0x04
 #define HX509_CMS_VS_NO_VALIDATE			0x08
 
+/* flags from hx509_cms_verify_signed_ext */
+#define HX509_CMS_VS_VALIDATED				0x01
```
nicowilliams, line 186: Did you mean to make this `0x10`, or did you mean to have this flag share a value with `HX509_CMS_VS_ALLOW_DATA_OID_MISMATCH`??

lhoward: It's an output flag. Perhaps a different prefix is in order.

nicowilliams: Oh, right, this is an _output_ flag here.  Maybe it should have a different prefix then?

lib/krb5/pkinit.c
```diff
@@ -829,33 +829,39 @@ pk_verify_sign(krb5_context context,
...
+				      &signer_certs,
+				      &verify_flags);
     if (ret) {
 	pk_copy_error(context, context->hx509ctx, ret,
 		      "CMS verify signed failed");
 	return ret;
     }
 
+    if ((verify_flags & HX509_CMS_VSE_VALIDATED) == 0)
+	goto out;
```
nicowilliams, line 863: But `ret == 0` here...  Don't we need to set it to something?  Isn't this an error condition?

nicowilliams: I think this needs fixing, otherwise this looks good!

nicowilliams:
Do we need a test?  How are you testing this?

lhoward:
Sorry I missed some of these comments before merging (Skype/Github synchronisation error!). I'll address any issues in a follow-up commit.

lib/krb5/pkinit.c
lhoward: It's valid to return 0 but no signing certificate if `HX509_CMS_VSE_VALIDATED` was set. See the caller assert below:

```c
    heim_assert(host || (ctx->id->flags & PKINIT_NO_KDC_ANCHOR),
                "KDC signature must be verified unless PKINIT_NO_KDC_ANCHOR set");
```

lhoward:
> Do we need a test? How are you testing this?

there is one test at present

kaduk:
> The KDC certificate is not authenticated, per RFC 6113 5.4.1.1.

It's late here, but I am seeing an option (2) in that section that involves anonymous pkinit that does authenticate the KDC as being preferred to option (3) that does not authenticate the KDC and is vulnerable to MITM.  I assume there is some more subtlety that you had in mind when writing this, but it seemed worth checking.

lhoward:
> > The KDC certificate is not authenticated, per RFC 6113 5.4.1.1.
> 
> It's late here, but I am seeing an option (2) in that section that involves anonymous pkinit that does authenticate the KDC as being preferred to option (3) that does not authenticate the KDC and is vulnerable to MITM. I assume there is some more subtlety that you had in mind when writing this, but it seemed worth checking.

The KDC will be authenticated if it can be, although obviously as there's no indication to the user whether this happened or not, that is kind of useless. Currently if you want a KDC-authenticated anonymous PKINIT FAST armour cache then you should retrieve one explicitly and store it in a credential cache.

Generally the MITM is not an issue because the inner exchange authenticates the KDC, e.g. by the KDC proving that it knows the user's password when using PA-ENC-TIMESTAMP. It _is_ an issue when using pre-authentication mechanisms that don't authenticate the KDC (such as GSS-API without `GSS_C_MUTUAL_FLAG`), and yes, those cases will explicitly check the KDC authentication state before proceeding.

If you like, we could display a warning when the KDC could not be authenticated.

kaduk:
Thanks for the more in-depth explanation.  I'm not requesting any behavior change, and just wanted to better understand how the intended change related to the RFC.  I have that now, so we should be all set.

lhoward:
> Thanks for the more in-depth explanation. I'm not requesting any behavior change, and just wanted to better understand how the intended change related to the RFC. I have that now, so we should be all set.

I actually originally added it because I thought using FX-COOKIE (which I wanted for GSS preauth) would require it, but turns out it didn't.

PR #796, opened 2021-08-01
Implement draft-perez-krb-wg-gss-preauth-02.txt

lhoward:
Creating PR to track GSS-API pre-authentication support

Note this protocol differs from the draft: GSS pre-auth contains the raw GSS-API token (not encapsulated in PA-GSS) and the state is managed by FAST FX-COOKIE.

PR #798, opened 2021-08-04
Support split gss tokens

elric1:
There are a number of other commits in this pull request to support the main change, but we present the meat of the commit message for said main change's commit:

```
    gss_accept_sec_context: support reassembling split tokens.

    Microsoft will sometimes split GSS tokens when they exceed a certain
    size in some protocols.  This is specified in

            [MS-SPNG]: Simple and Protected GSS-API Negotiation
            Mechanism (SPNEGO) Extension

    https://winprotocoldoc.blob.core.windows.net/productionwindowsarchives/MS-SPNG/%5bMS-SPNG%5d.pdf

    sections 3.1.5.4 to 3.1.5.9.

    We extend gss_accept_sec_context() to recognise partial tokens and
    to accumulate the fragments until an entire token is available to
    be processed.  If the entire token is not yet available,
    GSS_S_CONTINUE_NEEDED is returned with a zero length output token.
    This is specified in RFC2744 page 25-26 to indicate that no reply
    need be sent.

    We include updates to the test framework to test split tokens when
    using SPNEGO.
```


lhoward:
It looks like no changes are required to the SPNEGO mech, is that correct? I ask as I have in #796 added support for exporting partially established SPNEGO contexts (to use with GSS pre-authentication on the KDC) and wonder what the interaction would be with this.

elric1:
@lukeh: Thanks for asking! I need to fix SPNEGO, but just before I did it, @nicowilliams suggested talking to you about it first. So, I'd like the export of partially established SPNEGO contexts to work, but I didn't implement it. And so there should be no conflicts with your work.

lib/gssapi/mech/gss_accept_sec_context.c
```diff
@@ -168,37 +266,35 @@ gss_accept_sec_context(OM_uint32 *minor_status,
...
+        }
+
+        major_status = accumulate_token(ctx, input_token);
+        if (major_status != GSS_S_COMPLETE)
+                return major_status;
+
+        /*
+         * If we get here, then we have a complete token.  Please note
+         * that we may have a major_status of GSS_S_DEFECTIVE_TOKEN.  This
+         *
```
lhoward, line 286: Is the comment finished?

lhoward (commented):
LGTM!

lib/gssapi/mech/gss_accept_sec_context.c
```diff
@@ -28,87 +28,148 @@
...
+                ctx->gc_input.value  = p;
+
+                if (len == 0)
+                        return GSS_S_COMPLETE;
+
+                /* Not our DER w/ a length */
+                if (*p != 0x60)
+                        return GSS_S_COMPLETE;
+
+                if (der_get_length(p+1, len-1, &ctx->gc_target_len, &l) != 0)
```
lhoward, line 92: If I was being picky, whitespace around p+1 but...

lib/krb5/store.c
```diff
@@ -335,17 +336,62 @@ krb5_storage_to_data(krb5_storage *sp, krb5_data *data)
     return 0;
 }
 
+static size_t
+pack_int(void *buffer, uint64_t value)
+{
```
lhoward, line 341: Two issues here: firstly, 32-bit negative numbers (e.g. ~0UL) are sign extended which makes the encoding less efficient. We could add a size argument to `pack_int()` and mask `value` off, or perhaps there is another approach.

Secondly, encoding ~0ULL is not possible – the decoder parses it as two integers. We will need to expand the number of bits used to encode the length. If the length can never be zero, then we can encode one less than the length, but we still need to account for `length - 1 == 8` for ~0ULL.

Cc @nicowilliams

lib/krb5/store.c
```diff
@@ -449,8 +495,20 @@ krb5_ret_int(krb5_storage *sp,
...
     unsigned char v[8];
-    uint64_t w;
+    uint64_t w = 0;
     *value = 0; /* quiets warnings */
+    if (BYTEORDER_IS_PACKED(sp)) {
+        ret = sp->fetch(sp, v, 1);
+        if (ret < 0)
+            return errno;
+        len = unpack_int(v, &w, 0);
+        ret = sp->fetch(sp, v + 1, len - 1);
```
lhoward, line 505: We need to check `len > 0` otherwise we have an overflow.

lhoward (changes requested):

lhoward:
If you resolve the issues with a separate commit I can cherry-pick, I have a local branch that resolves the conflicts in `test_context.c`. Although I'm sure you can do that yourself too!

lhoward:
Fixed, merged.

PR #799, opened 2021-08-04
gss: allow partial accept context export in SPNEGO

lhoward:
Support for exporting partially established acceptor context tokens. With this, an acceptor can send the initiator an encrypted state cookie containing the exported context token.

(The concrete mechanism, of course, must either require a single round trip or support partial context export itself. Kerberos and GSS EAP would work, but Kerberos with `GSS_C_DCE_STYLE` would not, as currently implemented.)

Partial context export is not permitted for initiators.

### DIFF
--- a/lib/gssapi/spnego/context_storage.c
+++ b/lib/gssapi/spnego/context_storage.c
@@ -1,0 +1,584 @@
+/*
+ * Copyright (C) 2021, PADL Software Pty Ltd.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ * * Redistributions of source code must retain the above copyright
+ *   notice, this list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright
+ *   notice, this list of conditions and the following disclaimer in
+ *   the documentation and/or other materials provided with the
+ *   distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+ * FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ * COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT,
+ * INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+ * STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+ * OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "spnego_locl.h"
+#include "mech/mech_locl.h"
+
+#define SC_MECH_TYPES               0x0001
+#define SC_PREFERRED_MECH_TYPE      0x0002
+#define SC_SELECTED_MECH_TYPE       0x0004
+#define SC_NEGOTIATED_MECH_TYPE     0x0008
+#define SC_NEGOTIATED_CTX_ID        0x0010
+#define SC_MECH_FLAGS               0x0020
+#define SC_MECH_TIME_REC            0x0040
+#define SC_MECH_SRC_NAME            0x0080
+#define SC_TARGET_NAME              0x0100
+#define SC_NEGOEX                   0x0200
+
+#define SNC_OID                     0x01
+#define SNC_MECH_CONTEXT            0x02
+#define SNC_METADATA                0x04
+
+static krb5_error_code
+ret_spnego_context(krb5_storage *sp, gssspnego_ctx *ctxp);
+static krb5_error_code
+store_spnego_context(krb5_storage *sp, gssspnego_ctx ctx);
+
+static krb5_error_code
+ret_negoex_auth_mech(krb5_storage *sp, struct negoex_auth_mech **mechp);
+static krb5_error_code
+store_negoex_auth_mech(krb5_storage *sp, struct negoex_auth_mech *mech);
+
+static krb5_error_code
+ret_gss_oid(krb5_storage *sp, gss_OID *oidp);
+static krb5_error_code
+store_gss_oid(krb5_storage *sp, gss_OID oid);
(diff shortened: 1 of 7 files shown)


lib/gssapi/spnego/spnego_locl.h
```diff
@@ -147,4 +149,23 @@ struct gssspnego_optimistic_ctx {
...
 
+static inline int
+gssspnego_ctx_complete_p(gssspnego_ctx ctx)
+{
+    return ctx->flags.open &&
+	    (ctx->flags.safe_omit || (ctx->flags.sent_mic && ctx->flags.verified_mic));
+}
+
+static inline krb5_error_code
+store_bytes(krb5_storage *sp, const void *bytes, size_t length)
```
lhoward, line 160: @elric1 will replace this with `krb5_store_datalen()` when your PR gets merged

lib/gssapi/spnego/context_stubs.c
```diff
@@ -319,25 +321,30 @@ OM_uint32 GSSAPI_CALLCONV _gss_spnego_export_sec_context (
 
     HEIMDAL_MUTEX_lock(&ctx->ctx_id_mutex);
 
-    if (ctx->negotiated_ctx_id == GSS_C_NO_CONTEXT) {
-	HEIMDAL_MUTEX_unlock(&ctx->ctx_id_mutex);
-	return GSS_S_NO_CONTEXT;
+    /*
+     * Partial context export is only supported on the acceptor side, as we
+     * cannot represent the initiator function pointer state in an exported
+     * token, and also because it is only useful for acceptors which need
```
elric1, line 327: Very minor nit: I'd say it is "mostly" useful for acceptors, because who knows what we'll think of in the future.

elric1 (commented):
In my code, I tried to add `--export-import-context` to `tests/gss/check-spnego.in` in these lines:
```
echo "spnego"
${context} \
    --client-ccache="${cache}" \
    --mech-type=spnego \
    --ret-mech-type=krb5 \
    --name-type=hostbased-service \
    host@host.test.h5l.se || \
    { exitcode=1 ; echo test failed; }

echo "spnego (split tokens)"
${context} \
    --token-split=128 \
    --client-ccache="${cache}" \
    --mech-type=spnego \
    --ret-mech-type=krb5 \
    --name-type=hostbased-service \
    host@host.test.h5l.se || \
    { exitcode=1 ; echo test failed; }
```
Your code here will enable this---and this should be a good test case for you to add in this commit.
Hmmm, actually, I added `--export-import-context` to that test case in `lib/gssapi/test_context.c`. But, I think that it'll provide a good test case for your code.

lhoward:
`--export-import-context` in SPNEGO should already be changed by `check-negoex.in`, I think.

lhoward:
But I guess yes, should check for export import of partial contexts, but that may require changes to `test_context`?

lhoward:
> In my code, I tried to add `--export-import-context` to `tests/gss/check-spnego.in` in these lines:
> 
> ```
> echo "spnego"
> ${context} \
>     --client-ccache="${cache}" \
>     --mech-type=spnego \
>     --ret-mech-type=krb5 \
>     --name-type=hostbased-service \
>     host@host.test.h5l.se || \
>     { exitcode=1 ; echo test failed; }
> 
> echo "spnego (split tokens)"
> ${context} \
>     --token-split=128 \
>     --client-ccache="${cache}" \
>     --mech-type=spnego \
>     --ret-mech-type=krb5 \
>     --name-type=hostbased-service \
>     host@host.test.h5l.se || \
>     { exitcode=1 ; echo test failed; }
> ```
> 
> Your code here will enable this---and this should be a good test case for you to add in this commit.
> Hmmm, actually, I added `--export-import-context` to that test case in `lib/gssapi/test_context.c`. But, I think that it'll provide a good test case for your code.

Passing `--token-split=128 --export-import-context` in `test_context` calls `gss_export_sec_context()` before the first call from the mechglue to the mechanism's `gss_accept_sec_context()`. So unless the mechglue itself supports exporting partial contexts I'm not sure how this can ever work?

lhoward:
Ah, I see the mechglue itself does support exporting partial contexts.

`accumulate_token()` is returning `GSS_C_CONTINUE_NEEDED` because `gci->length != ctx->gc_target_len`, before the first call to the mechanism `gss_accept_sec_context()`.

Thus `gss_export_sec_context()` bails because `ctx->gc_ctx == NULL` (line 59).

lhoward:
OK, I didn't review this closely enough before integrating. There are some use after frees etc. I will fix and merge.

lhoward:
Fixed. Note that `test_context` can only split tokens where the first byte is `0x60`, otherwise `gss_accept_sec_context()` doesn't have enough information to piece it together.

`gss_import_sec_context()` also failed to allocate a buffer large enough to accumulate up to the target length.

I have also added the tests you suggested.

Fixed in 06232dfccee57aa0f49ca94c6e0428ec94ef15ca.

lhoward:
Obviously changing the exported context token format is a breaking ABI change but, I don't think we've cared too much about this in the past. As long as the libraries get a version number bump.

lhoward:
Sorry all those comments should be been directed to #798 not this PR.

lib/gssapi/spnego/context_storage.c
```diff
@@ -0,0 +1,584 @@
...
+    if (snc_flags & SNC_OID)
+        CHECK(ret, store_gss_oid(sp, mech->oid));
+
+    CHECK(ret, krb5_store_bytes(sp, mech->scheme, GUID_LENGTH));
+
+    if (snc_flags & SNC_MECH_CONTEXT) {
+        CHECK(major, gss_export_sec_context(&minor, &mech->mech_context,
+                                            &buf));
+        CHECK(ret, store_gss_buffer(sp, &buf));
+    }
```
elric1, line 439: In the rest of your code, I think that you would have put a `gss_release_buffer(&minor, &buf)` here even though it is `free(3)`d at the end. It's not a bug, but it's a slightly different idiom than what you've been doing above.

lib/gssapi/spnego/context_storage.c
```diff
@@ -0,0 +1,584 @@
...
+    if (oid) {
+        data.length = oid->length;
+        data.data = oid->elements;
+    }
+
+    return krb5_store_data(sp, data);
+}
+
+static krb5_error_code
+ret_gss_buffer(krb5_storage *sp, gss_buffer_t buffer)
```
elric1, line 495: I wonder if `{ret,store}_gss_{oid,buffer}()` might be of more general use and might want to have a life where they can be shared.

lib/gssapi/spnego/context_storage.c
```diff
@@ -0,0 +1,584 @@
...
+
+    if (buffer) {
+        data.length = buffer->length;
+        data.data = buffer->value;
+    }
+
+    return krb5_store_data(sp, data);
+}
+
+static uint16_t
```
elric1, line 530: It might be an idea here and lines 365 and 413 to name these constants.

lib/gssapi/spnego/context_storage.c
lhoward: True, there's only one use of it but it would be more future-proof to free it.

lib/gssapi/spnego/context_storage.c
lhoward: `_gss_mg_{ret,store}_gss_{oid,buffer}()` makes sense. Question is whether we should introduce those now or wait until we have a use?

lib/gssapi/spnego/context_storage.c
lhoward: I felt if we used constants, we should use the names everywhere (which would be more intrusive but could do that in a separate commit).

lhoward: As in, make flags in the context an integer rather than bit field.

PR #804, opened 2021-08-09
kdc: make it possible to disable the principal based referral detection

abartlet:
This, from @metze-samba and @cryptomilk is the KDC side of Samba BUG: https://bugzilla.samba.org/show_bug.cgi?id=12554 and I suspect is significantly so that we can both match Windows and trigger the behaviour we see there in the tests of our client code.

Either way, I'm trying to upstream as much as we can of our lorikeet-heimdal patches, at least for comment.

I know this will collide/conflict with #787 but I'll happily rebase that if this lands first, but this was the location that was requested for weird compatibility flags for Samba.

### DIFF
--- a/kdc/default_config.c
+++ b/kdc/default_config.c
@@ -92,6 +92,7 @@ krb5_kdc_get_config(krb5_context context, krb5_kdc_configuration **config)
     c->preauth_use_strongest_session_key = FALSE;
     c->svc_use_strongest_session_key = FALSE;
     c->use_strongest_server_key = TRUE;
+    c->autodetect_referrals = TRUE;
     c->check_ticket_addresses = TRUE;
     c->warn_ticket_addresses = FALSE;
     c->allow_null_ticket_addresses = TRUE;
--- a/kdc/kdc.h
+++ b/kdc/kdc.h
@@ -62,6 +62,12 @@ typedef struct krb5_kdc_configuration {
 
     krb5_boolean encode_as_rep_as_tgs_rep; /* bug compatibility */
 
+    /*
+     * For Samba's AD DC, which will set to FALSE
+     * to turn off principal base referral auto-detection
+     */
+    krb5_boolean autodetect_referrals;
+
     krb5_boolean tgt_use_strongest_session_key;
     krb5_boolean preauth_use_strongest_session_key;
     krb5_boolean svc_use_strongest_session_key;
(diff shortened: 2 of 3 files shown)


kdc/kdc.h
```diff
@@ -62,6 +62,12 @@ typedef struct krb5_kdc_configuration {
 
     krb5_boolean encode_as_rep_as_tgs_rep; /* bug compatibility */
 
+    /*
+     * For Samba's AD DC, which will set to FALSE
+     * to turn off principal base referral auto-detection
```
nicowilliams, line 67: Missing 'd' in `base`.

kdc/default_config.c
```diff
@@ -92,6 +92,7 @@ krb5_kdc_get_config(krb5_context context, krb5_kdc_configuration **config)
     c->preauth_use_strongest_session_key = FALSE;
     c->svc_use_strongest_session_key = FALSE;
     c->use_strongest_server_key = TRUE;
+    c->autodetect_referrals = TRUE;
```
nicowilliams, line 95: Maybe a different name?  How about `enable_referrals`?

lhoward:
I'm a bit iffy about this one. Rather than yet another knob that disables a chunk of KDC-side functionality, I think it would be cleaner to perhaps have a callback interface Samba could implement (the default implementation for which would be the existing KDC one). If the Samba one wants to be a NOOP, fine.

lhoward:
Also, does this mean this code can be removed?

{code}
    else if (server->name.name_string.len == 3) {
        /*
          This is used to give referrals for the
          E3514235-4B06-11D1-AB04-00C04FC2DCD2/NTDSGUID/DNSDOMAIN
          SPN form, which is used for inter-domain communication in AD
         */
        name = server->name.name_string.val[2];
        kdc_log(context, config, 4, "Giving 3 part referral for %s", name);
        *realms = malloc(sizeof(char *)*2);
        if (*realms == NULL) {
            krb5_set_error_message(context, ENOMEM, N_("malloc: out of memory", ""));
            return FALSE;
        }
        (*realms)[0] = strdup(name);
        (*realms)[1] = NULL;
        return TRUE;
{code}

lhoward:
Closing: I believe #917 will be a better solution. You can implement the referral policy callback to simply return zero.

PR #805, opened 2021-08-09
Import AS/TGS client from Apple Heimdal-597.121.1

lhoward:
This PR extends #796 with a merge of Apple's Heimdal open source drop 597.121.1.

Clearly it would have been better to build GSS pre-authentication on top of this, but, here we are. I would suggest reviewing and integrating #796 and #797 first.

The tests pass.

There is a lot of churn but, I have minimised it somewhat by not aggressively integrating some of the Apple features (e.g. completely rewritten TGS client). I am in the process of carefully comparing with the original to check for regressions.

The advantages of integrating this are:

- support for PA-ENC-CHALLENGE
- FAST reply key strengthening
- FAST support in TGS-REQ path (including for errors which the Apple code did not implement), both in libkrb5 and KDC
- refactored pre-authentication client that makes adding new mechanisms easy
- allow KDC hostname and AD site name to be pinned in `get_[_init]creds()` APIs

I've also fixed a bunch of bugs in the Apple code, e.g. it would not optimistic PA mechanisms, it could crash with Win2K PKINIT responses, etc.

### DIFF
--- a/kdc/fast.c
+++ b/kdc/fast.c
@@ -334,7 +334,6 @@ _kdc_fast_mk_error(astgs_request_t r,
 		   time_t *csec, int *cusec,
 		   krb5_data *error_msg)
 {
-    krb5_context context = r->context;
     krb5_error_code ret;
     krb5_data e_data;
     size_t size;
@@ -365,7 +364,7 @@ _kdc_fast_mk_error(astgs_request_t r,
 
 	/* first add the KRB-ERROR to the fast errors */
 
-	ret = krb5_mk_error(context,
+	ret = krb5_mk_error(r->context,
 			    outer_error,
 			    e_text,
 			    NULL,
@@ -377,7 +376,7 @@ _kdc_fast_mk_error(astgs_request_t r,
 	if (ret)
 	    return ret;
 
-	ret = krb5_padata_add(context, error_method,
+	ret = krb5_padata_add(r->context, error_method,
 			      KRB5_PADATA_FX_ERROR,
 			      e_data.data, e_data.length);
 	if (ret) {
@@ -394,14 +393,14 @@ _kdc_fast_mk_error(astgs_request_t r,
 	csec = 0;
 	cusec = 0;
 
-	ret = _kdc_fast_mk_response(context, armor_crypto,
+	ret = _kdc_fast_mk_response(r->context, armor_crypto,
 				    error_method, NULL, NULL,
 				    req_body->nonce, &e_data);
 	free_METHOD_DATA(error_method);
 	if (ret)
 	    return ret;
 
-	ret = krb5_padata_add(context, error_method,
+	ret = krb5_padata_add(r->context, error_method,
 			      KRB5_PADATA_FX_FAST,
 			      e_data.data, e_data.length);
 	if (ret)
@@ -416,7 +415,7 @@ _kdc_fast_mk_error(astgs_request_t r,
 	heim_assert(size == e_data.length, "internal asn.1 encoder error");
     }
 
-    ret = krb5_mk_error(context,
+    ret = krb5_mk_error(r->context,
 			outer_error,
 			e_text,
 			(e_data.length ? &e_data : NULL),
--- a/kdc/fast.c
+++ b/kdc/fast.c
@@ -46,6 +46,7 @@ get_fastuser_crypto(astgs_request_t r, krb5_enctype enctype,
 
     *crypto = NULL;
 
+    /* TODO: salt cookie key with client name and realm */
     ret = krb5_make_principal(r->context, &fast_princ,
 			      KRB5_WELLKNOWN_ORG_H5L_REALM,
 			      KRB5_WELLKNOWN_NAME, "org.h5l.fast-cookie", NULL);
@@ -116,7 +117,7 @@ fast_parse_cookie(astgs_request_t r, const PA_DATA *pa)
 	goto out;
 
     if (r->fast.expiration < kdc_time) {
-	kdc_log(r->context, r->config, 2, "fast cookie expired");
+	kdc_log(r->context, r->config, 2, "FAST cookie expired");
 	ret = KRB5KDC_ERR_POLICY;
 	goto out;
     }
@@ -140,15 +141,18 @@ fast_add_cookie(astgs_request_t r, METHOD_DATA *method_data)
 
     r->fast.expiration = kdc_time + FAST_EXPIRATION_TIME;
 
-    ASN1_MALLOC_ENCODE(KDCFastState, data.data, data.length, 
+    ASN1_MALLOC_ENCODE(KDCFastState, data.data, data.length,
 		       &r->fast, &size, ret);
     if (ret)
 	return ret;
-    heim_assert(size == data.length, "internal asn1 encoder error");
+    heim_assert(size == data.length, "internal asn.1 encoder error");
 
     ret = get_fastuser_crypto(r, KRB5_ENCTYPE_NULL, &crypto);
-    if (ret)
+    if (ret) {
+	kdc_log(r->context, r->config, 0,
+		"Failed to find FAST principal for cookie encryption: %d", ret);
 	goto out;
+    }
 
     ret = krb5_encrypt_EncryptedData(r->context, crypto,
 				     KRB5_KU_H5L_COOKIE,
@@ -157,24 +161,26 @@ fast_add_cookie(astgs_request_t r, METHOD_DATA *method_data)
     krb5_crypto_destroy(r->context, crypto);
     if (ret)
 	goto out;
-    
-    free(data.data);
+
+    krb5_data_free(&data);
 
     shell.version = "H5L1";
 
-    ASN1_MALLOC_ENCODE(KDCFastCookie, data.data, data.length, 
+    ASN1_MALLOC_ENCODE(KDCFastCookie, data.data, data.length,
 		       &shell, &size, ret);
     free_EncryptedData(&shell.cookie);
     if (ret)
 	goto out;
-    heim_assert(size == data.length, "internal asn1 encoder error");
-    
+    heim_assert(size == data.length, "internal asn.1 encoder error");
+
     ret = krb5_padata_add(r->context, method_data,
 			  KRB5_PADATA_FX_COOKIE,
 			  data.data, data.length);
+    if (ret == 0)
+	krb5_data_zero(&data);
+
  out:
-    if (ret)
-	free(data.data);
+    krb5_data_free(&data);
     return ret;
 }
 
@@ -209,9 +215,8 @@ _kdc_fast_mk_response(krb5_context context,
 		       &fastrep, &size, ret);
     if (ret)
 	return ret;
-    if (buf.length != size)
-	krb5_abortx(context, "internal asn.1 error");
-    
+    heim_assert(size == buf.length, "internal asn.1 encoder error");
+
     fxfastrep.element = choice_PA_FX_FAST_REPLY_armored_data;
 
     ret = krb5_encrypt_EncryptedData(context,
@@ -230,9 +235,8 @@ _kdc_fast_mk_response(krb5_context context,
     free_PA_FX_FAST_REPLY(&fxfastrep);
     if (ret)
 	return ret;
-    if (data->length != size)
-	krb5_abortx(context, "internal asn.1 error");
-    
+    heim_assert(size == data->length, "internal asn.1 encoder error");
+
     return 0;
 }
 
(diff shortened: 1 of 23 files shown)


lhoward:
TODO: Apple code uses separate fast cookie key per-realm, which prevents cross-realm cookie replays on KDCs that host multiple realms. Perhaps a simpler solution (that would not require upgrading HDBs) would be to salt the cookie key with the realm. Or, given they should not change in an AS-REQ, the client and server name, and realm.

lhoward:
It will likely be easier to review lib/krb5/{init_creds_pw.c,get_cred.c} themselves rather than the diffs.

lhoward:
@abartlet if you're adventurous you could check this branch against Samba. It gets you FAST TGS support.

abartlet:
Thanks.  I do recall grabbing some of Apple's FAST code when I was doing FAST work for Heimdal/Samba in 2018, so it is good to see a bigger sync up.

abartlet:
@josephsutton1 we should run our new FAST test against this branch and check how it matches up.

lhoward:
That would be great. I have only tested against Apple's own implementation (that ships with macOS). Not yet with MIT.

abartlet:
As someone involved in a similar exercise between Heimdal and Samba I
would minimise diffs were possible.

Andrew
-- 
Andrew Bartlett (he/him)       https://samba.org/~abartlet/
Samba Team Member (since 2001) https://samba.org
Samba Team Lead, Catalyst IT   https://catalyst.net.nz/services/samba

Samba Development and Support, Catalyst IT - Expert Open Source
Solutions

kdc/krb5tgs.c
```diff
@@ -2416,7 +2416,8 @@ _kdc_tgs_rep(astgs_request_t r)
 				 r->armor_crypto,
 				 &req->req_body,
 				 ret, r->e_text,
-				 NULL, NULL,
+				 ticket->client,
+				 ticket->server,
```
josephsutton1, line 2420: `ticket` may be NULL here, so you may wish to avoid dereferencing it in that case. Something like:
```
ticket ? ticket->client : NULL,
ticket ? ticket->server : NULL,
```

josephsutton1:
I've been running Samba's FAST tests against these changes and other than a few minor differences with error codes and padata in error replies, the results have been good so far.

kdc/krb5tgs.c
lhoward: ah thank you, will fix

lhoward:
> I've been running Samba's FAST tests against these changes and other than a few minor differences with error codes and padata in error replies, the results have been good so far.

Good to hear! If there are error code changes you think we should revert, let me know. I guess Love probably did test this against Windows, it's the classic issue where features have been sitting in the Apple code for years but remained un-merged (thanks to the large divergence since 2010).

abartlet:
Just wondering, what more do you think is needed before this is merged?

The reason is that I'm likely to include this in Samba regardless, so want to know if there are outstanding issues other than time.

lhoward:
@nicowilliams was going to review it but if he doesn’t have time, do you want @abartlet?

abartlet:
> @nicowilliams was going to review it but if he doesn’t have time, do you want @abartlet?

Sure, I'll have a go with @josephsutton1

abartlet:
Thanks for fixing the conflicts.  Sorry for the double-work, I had that already in our lorikeet-heimdal here: https://gitlab.com/samba-team/devel/lorikeet-heimdal/-/commits/lorikeet-heimdal-202111290112-patched/ but couldn't push to your branch and didn't get around to pushing somewhere else.

I still want to review this but it is a bit overwhelming!

lib/krb5/init_creds_pw.c
```diff
@@ -2524,18 +2968,35 @@ init_creds_step(krb5_context context,
...
+
     krb5_data_zero(out);
+    krb5_data_zero(&checksum_data);
 
     if (ctx->as_req.req_body.cname == NULL) {
 	ret = init_as_req(context, ctx->flags, &ctx->cred,
 			  ctx->addrs, ctx->etypes, &ctx->as_req);
-	if (ret)
+	if (ret) {
+	    free_init_creds_ctx(context, ctx);
```
abartlet, line 2987: This looks wrong.  None of the other error paths in this function call free_init_creds_ctx().

abartlet:
It there some value in showing that the code is the same or similar to the Apple code, as a code review technique, or did it have to change too much to import?

lhoward:
I did do some diffs to the Apple code for the files that were changed, when I reviewed it myself. But that was a while ago, it may have diverged a little since then.

kdc/fast.c
```diff
@@ -358,133 +446,141 @@ fast_unwrap_request(astgs_request_t r)
...
+	    free_AP_REQ(&ap_req);
+	    goto out;
+	}
 
-    ret = hdb_enctype2key(r->context, &armor_user->entry, NULL,
-			  ap_req.ticket.enc_part.etype,
-			  &armor_key);
-    if (ret) {
+	ret = _kdc_db_fetch(r->context, r->config, armor_server,
+			    HDB_F_GET_SERVER, NULL, NULL, &armor_user);
```
abartlet, line 526: the armor_server lookup changed from HDB_F_GET_KRBTGT | HDB_F_DELAY_NEW_KEYS to HDB_F_GET_SERVER, was that delibrate.  What principals are permissible here?  (Oddly, this appears to work regardless, given this passes Samba's tests and Samba is the only user of those flags).

kdc/kerberos5.c
```diff
@@ -695,68 +650,60 @@ pa_enc_chal_validate(astgs_request_t r, const PA_DATA *pa)
...
-	ret = krb5_copy_keyblock_contents(r->context, &k->key, &r->reply_key);
-	if (ret)
-	    goto out;
 
-	/*
-	 * Success
-	 */
-	if (r->clientdb->hdb_auth_status)
-	    r->clientdb->hdb_auth_status(r->context, r->clientdb, r->client,
-					 HDB_AUTH_SUCCESS);
```
abartlet, line 752: Why is this lost?

abartlet (commented):
Just a partial read over from the top for now.

lib/krb5/krbhst.c
```diff
@@ -65,7 +65,7 @@ is_invalid_tld_srv_target(const char *target)
 
 static krb5_error_code
 srv_find_realm(krb5_context context, krb5_krbhst_info ***res, int *count,
-	       const char *realm, const char *dns_type,
+	       const char *realm, const char *dns_type, const char *sitename,
```
abartlet, line 68: Interesting that we get 'sites' support with this.

lib/krb5/krb5.h
```diff
@@ -1008,6 +1008,7 @@ typedef struct krb5_name_canon_iterator_data *krb5_name_canon_iterator;
  *
  */
 
+struct hx509_cert_data;
```
abartlet, line 1011: I'm unclear why this is needed.

abartlet (approved):
I'm not sure my approval is worth the electrons they are send on, but given this passes our tests and I can't see much (other than as noted) really wrong, I guess this is OK.  

Mostly this was a read over and a diff with Apple where that was possible.

This is a 'there is nothing really strange going on' level of review, my Kerberos and Heimdal foo isn't strong enough to catch a subtle mistake in a change of this scale.

lhoward:
Thanks @abartlet! I'll have a look at your comments over the next few days :)

kdc/kerberos5.c
lhoward: Fixed, although isn't that also called from `_kdc_as_rep()`? And, should errors be reported from the PKINIT and GSS PA code paths as well? Can this be subsumed into the auditing functions?

kdc/krb5tgs.c
```diff
@@ -898,10 +920,7 @@ tgs_parse_request(astgs_request_t r,
 		  const struct sockaddr *from_addr,
 		  time_t **csec,
 		  int **cusec,
-		  AuthorizationData **auth_data,
-		  krb5_keyblock **replykey,
-		  Key **header_key,
-		  int *rk_is_subkey)
+		  AuthorizationData **auth_data)
 {
     krb5_context context = r->context;
```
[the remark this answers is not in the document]
lhoward: Good idea

kdc/kerberos5.c
abartlet: Yes, see https://gitlab.com/samba-team/devel/lorikeet-heimdal/-/commit/b306eecbc9126b2afa5f83dd197296384a6506b9

It may well be that we remove this again in landing that.

abartlet: Yeah, I agree it looks like a duplicate now, sorry for not digging into it further.  Anyway, once this is in I'll clean up the patch to do this properly in auditing and ensure we do these things exactly once.

lib/krb5/init_creds_pw.c
```diff
@@ -2758,62 +3309,102 @@ init_creds_step(krb5_context context,
...
+
+	    } else {
+		if (ctx->fast_state.flags & KRB5_FAST_OPTIMISTIC) {
+		    _krb5_debug(context, 10,
+				"Some other error %d failed with Optimistic FAST, trying w/o FAST", ret);
+
+		    ctx->fast_state.flags &= ~KRB5_FAST_OPTIMISTIC;
+		    ctx->fast_state.flags |= KRB5_FAST_DISABLED;
+		    pa_restart(context, ctx);
+
```
josephsutton1, line 3362: Should there be an `} else {` here? I noticed this was causing this function to keep looping when it received an error from the KDC that wasn't handled in one of the cases above.

abartlet:
See https://github.com/josephsutton1/heimdal/commits/gss-preauth-apple for a version of this that fixes the current conflicts.

lib/krb5/init_creds_pw.c
lhoward: Yes, how about this:

```diff
diff --git a/lib/krb5/init_creds_pw.c b/lib/krb5/init_creds_pw.c
index de6b13a53..b04b6ebab 100644
--- a/lib/krb5/init_creds_pw.c
+++ b/lib/krb5/init_creds_pw.c
@@ -3351,18 +3351,16 @@ init_creds_step(krb5_context context,
            retry:
                pa_restart(context, ctx);
 
-           } else {
-               if (ctx->fast_state.flags & KRB5_FAST_OPTIMISTIC) {
-                   _krb5_debug(context, 10,
-                               "Some other error %d failed with Optimistic FAST, trying w/o FAST", ret);
-
-                   ctx->fast_state.flags &= ~KRB5_FAST_OPTIMISTIC;
-                   ctx->fast_state.flags |= KRB5_FAST_DISABLED;
-                   pa_restart(context, ctx);
+           } else if (ctx->fast_state.flags & KRB5_FAST_OPTIMISTIC) {
+               _krb5_debug(context, 10,
+                           "Some other error %d failed with Optimistic FAST, trying w/o FAST", ret);
 
-                   /* some other error code from the KDC, lets return it to the user */
-                   goto out;
-               }
+               ctx->fast_state.flags &= ~KRB5_FAST_OPTIMISTIC;
+               ctx->fast_state.flags |= KRB5_FAST_DISABLED;
+               pa_restart(context, ctx);
+           } else {
+               /* some other error code from the KDC, lets return it to the user */
+               goto out;
            }
        }
     }
```

lhoward:
> See https://github.com/josephsutton1/heimdal/commits/gss-preauth-apple for a version of this that fixes the current conflicts.

I'm a bit pressed for time, so I'm just going to check that it's conflict free against master for now.

abartlet:
Thanks so much.  Let us know if there is anything more that can be done to help this land.

lhoward:
> Thanks so much. Let us know if there is anything more that can be done to help this land.

Still just waiting on @nicowilliams to have a look...

lhoward:
@nicowilliams will review this PR in the coming days

jaltman:
@abartlet what this PR needs more than anything else are tests.   Can the Samba team write some for the Heimdal test suite?

abartlet:
I've been thinking about this after some comments from the RH folks involved in our MIT work also. 

It would be really awesome to see the great work by @josephsutton1 reused and helping Heimdal ensure this continues to work long-term. 

Sadly our extensive testsuite is currently quite dependent on the Samba test infrastructure (python libs) and AD behaviour.  

The closest we come is that many of our tests have a mode that uses an existing account specified by environment variables (rather than being auto-created or downloaded over eg DRSUAPI), but it still assumes this is an AD account it can query over LDAP.

The best we can offer right now is a documented procedure for rebasing the lorikeet-heimdal tree on Heimdal master and importing it into Samba, so you can run our full CI (which has passed against a tree essentially similar to this), and is running against exactly this tree here: https://gitlab.com/samba-team/devel/samba/-/pipelines/425690907

Sorry!

abartlet:
@josephsutton1 confirms that current master as a client (eg `kinit --fast-armor-cache`) fails when use against a Windows DC supporting fast but succeeds when using this PR.

lhoward:
> @josephsutton1 confirms that current master as a client (eg `kinit --fast-armor-cache`) fails when use against a Windows DC supporting fast but succeeds when using this PR.

Right, if we define ”correctly implementing RFC 6113” to mean disallowing `PA-ENC-TIMESTAMP` inside FAST then, it's no surprise master fails because it doesn't support `PA-ENCRYPTED-CHALLENGE` at all :)

But, we don't want to break our existing clients if we can avoid it...

nicowilliams (approved):

PR #809, opened 2021-08-18
Check CLANG_FORMAT is executable.

RobCrowston:
If clang-format is not available, AC_CHECK_PROG sets CLANG_FORMAT=no,
not empty string. This results in an error at build time for those
without clang-format installed. Instead check if the program is
executable.

### DIFF
--- a/cf/Makefile.am.common
+++ b/cf/Makefile.am.common
@@ -147,7 +147,7 @@ SUFFIXES += .x .z .hx
 # clang-format styles is to sort includes, but in many cases in-tree we really
 # don't want to do that.
 .x.c:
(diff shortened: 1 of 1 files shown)


RobCrowston:
Error is of the form
```
Making all in asn1
/bin/sh: no: not found
*** Error code 127

Stop.
```
caused by [configure.ac](https://github.com/heimdal/heimdal/blob/9dcab76724b417140b4e475701118a01d2892e7c/configure.ac#L21) falling back to "no" if `clang-format` isn't found.

lhoward:
Ah, that's what was causing all those annoying messages. Thank you!

PR #810, opened 2021-08-20
Only #include <malloc.h> if it is available.

RobCrowston:
e.g., OpenBSD does not provide this header.


RobCrowston:
It's not entirely clear to me why we need this header since we don't seem to reference any non-standard `malloc()`-related functions in the source code -- and `malloc(3)` should be provided by `<stdlib.h>`, [per POSIX](https://pubs.opengroup.org/onlinepubs/009695399/functions/malloc.html). This is the least-invasive fix for building on OpenBSD.

PR #812, opened 2021-08-26
kdc: allow TGS requests from synthetic clients

lhoward:
Previously the TGS allowed synthetic clients (that is, clients that do not exist in the HDB) only if PKINIT was used. Since the introduction of support for GSS-API pre-authentication, synthetic clients can be used without PKINIT.

To support these in the TGS, use a new authorization data element to indicate the AS validated the usage of a synthetic client.

Note: untested.

### DIFF
--- a/lib/asn1/krb5.asn1
+++ b/lib/asn1/krb5.asn1
@@ -219,11 +219,12 @@ AUTHDATA-TYPE ::= INTEGER {
         -- N.B. these assignments have not been confirmed yet.
         --
         -- DO NOT USE in production yet!
+	KRB5-AUTHDATA-SYNTHETIC-CLIENT(-18),  -- client synthesized by KDC
 	KRB5-AUTHDATA-ON-BEHALF-OF(580),      -- UTF8String princ name
 	KRB5-AUTHDATA-BEARER-TOKEN-JWT(581),  -- JWT token
 	KRB5-AUTHDATA-BEARER-TOKEN-SAML(582), -- SAML token
 	KRB5-AUTHDATA-BEARER-TOKEN-OIDC(583), -- OIDC token
-	KRB5-AUTHDATA-CSR-AUTHORIZED(584),     -- Proxy has authorized client
+	KRB5-AUTHDATA-CSR-AUTHORIZED(584),    -- Proxy has authorized client
                                               -- to requested exts in CSR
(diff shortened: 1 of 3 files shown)


lib/asn1/krb5.asn1
```diff
@@ -223,8 +223,9 @@ AUTHDATA-TYPE ::= INTEGER {
 	KRB5-AUTHDATA-BEARER-TOKEN-JWT(581),  -- JWT token
 	KRB5-AUTHDATA-BEARER-TOKEN-SAML(582), -- SAML token
 	KRB5-AUTHDATA-BEARER-TOKEN-OIDC(583), -- OIDC token
-	KRB5-AUTHDATA-CSR-AUTHORIZED(584),     -- Proxy has authorized client
+	KRB5-AUTHDATA-CSR-AUTHORIZED(584),    -- Proxy has authorized client
                                               -- to requested exts in CSR
+	KRB5-AUTHDATA-SYNTHETIC-CLIENT(585),  -- client synthesized by KDC
 	KRB5-AUTHDATA-GSS-COMPOSITE-NAME(655) -- gss_export_name_composite
```
kaduk, line 229: What's the plan  for getting registered authdata type values for these "DO NOT USE in production yet!" uses?

lhoward: I can't speak to @nicowilliams' types, but I'm working on an update to draft-perez-krb-wg-gss-preauth.

lhoward: Oh sorry, this was re: SYNTHETIC-CLIENT. I don't think this one needs to be documented because its only consumer is the Heimdal KDC. But, we do need to check there's a private use number we can safely use.

kaduk: Sorry, I should have been more clear about the question (but you figured it out, fortunately).
I think these are not in an IANA registry, though, but rather something that tlyu had maintained for a long time and maybe has transitioned to ghudson.  https://datatracker.ietf.org/doc/html/draft-ietf-kitten-kerberos-iana-registries-04 stalled at the IETF for lack of energy but suggested that negative values would be suitable for local use (which meshes with some fragmentary memories I have of some negative values in a different codepoint space seeing common use because Microsoft caused them to be de facto standards).

lhoward:
Thanks @RobCrowston, I'll fix it.

PR #813, opened 2021-08-27
krb5: KRB5_KRBHST_TKTBRIDGEAP

lhoward:
Add a new krbhst type, KRB5_KRBHST_TKTBRIDGEAP to krb5_krbhst_init_flags(), that looks for the DNS SRV record kerberos-tkt-bridge. This is to support a new PADL project.


lhoward:
Merging it now TktBridgeAP is open source.

PR #814, opened 2021-09-01
Windows 10 SDK build fixes

lhoward:
Creating a PR to track Windows 10 build fixes

* update for Windows 10 SDK
* build without Win32.mak

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -42,7 +42,7 @@ install:
 
 build_script:
   - set PSDKDir=C:\Program Files\Microsoft SDKs\Windows\v7.1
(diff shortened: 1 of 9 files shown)


lhoward:
Failing in:

```
lex.obj : error LNK2019: unresolved external symbol strtoll referenced in function yylex
876C:\projects\heimdal\lib\asn1\..\..\out\dest_AMD64\bin\asn1_compile.exe : fatal error LNK1120: 1 unresolved externals
877
```

Clearly checking `_MSC_VER` is not the right way to test if we have UCRT.

PR #819, opened 2021-09-08
Update bits.c to fix memory leak

Hyxl1017:
p will not longer be use add after line 140,so add "free(q);" at line 140 to fix memory leak.

### DIFF
--- a/include/bits.c
+++ b/include/bits.c
@@ -137,6 +137,7 @@ int main(int argc, char **argv)
 	    if(!isalnum((unsigned char)*p))
 		*p = '_';
 	}
+	free(q);
 	f = fopen(argv[1], "w");
     }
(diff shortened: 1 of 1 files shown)


lhoward:
Did you check this actually compiles? `q` is not defined anywhere.

lhoward:
To fix this properly, you're going to need a non-const alias of the `p` base pointer that can be freed at the *end* of the program, as `hb` is used throughout. Please resubmit.

lhoward:
Or you could just make `hb` non-const, and check `argc >= 2` at the end of the program before calling `free(hb)`.

lhoward:
Closing, awaiting resubmission.

Hyxl1017:
thanks for advice.

PR #835, opened 2021-09-17
krb5: default to requiring keyed checksums

lhoward:
Make keyed checksums required when generating and verifying checksums, with the following exceptions:

* the checksum is being generated or verified as part of encrypting data for  a legacy (DES) encryption type

* the KRB5_CRYPTO_FLAG_ALLOW_UNKEYED_CHECKSUM flag was set on the crypto context, used to allow unkeyed checksums on krb5 authenticators

By making unkeyed checksums opt-in, we eliminate a class of potential vulnerabilities where callers could pass unkeyed checksums.

Any code that uses the mandatory checksum type for a given non-legacy encryption type should not be affected by this change. It could potentially break, say, a client trying to do FAST with DES keys but, that should not be supported (because FAST KDCs also support AES).

### DIFF
--- a/kdc/krb5tgs.c
+++ b/kdc/krb5tgs.c
@@ -791,6 +791,24 @@ out:
     return ret;
 }
 
+krb5_error_code
+_kdc_verify_checksum(krb5_context context,
+		     krb5_crypto crypto,
+		     krb5_key_usage usage,
+		     const krb5_data *data,
+		     Checksum *cksum)
+{
+    krb5_error_code ret;
+
+    ret = krb5_verify_checksum(context, crypto, usage,
+			       data->data, data->length,
+			       cksum);
+    if (ret == KRB5_PROG_SUMTYPE_NOSUPP)
+	ret = KRB5KDC_ERR_SUMTYPE_NOSUPP;
+
+    return ret;
+}
+
 static krb5_error_code
 tgs_check_authenticator(krb5_context context,
 			krb5_kdc_configuration *config,
@@ -830,11 +848,10 @@ tgs_check_authenticator(krb5_context context,
      * not require it to be keyed (as the authenticator is encrypted).
      */
     _krb5_crypto_set_flags(context, crypto, KRB5_CRYPTO_FLAG_ALLOW_UNKEYED_CHECKSUM);
-    ret = krb5_verify_checksum(context,
+    ret = _kdc_verify_checksum(context,
 			       crypto,
 			       KRB5_KU_TGS_REQ_AUTH_CKSUM,
-			       b->_save.data,
-			       b->_save.length,
+			       &b->_save,
 			       auth->cksum);
     krb5_crypto_destroy(context, crypto);
     if(ret){
@@ -1862,11 +1879,10 @@ server_lookup:
 		    ret = KRB5KRB_AP_ERR_BAD_INTEGRITY;
 	    }
 	    else {
-		ret = krb5_verify_checksum(context,
+		ret = _kdc_verify_checksum(context,
 					   crypto,
 					   KRB5_KU_OTHER_CKSUM,
-					   datack.data,
-					   datack.length,
+					   &datack,
 					   &self.cksum);
 	    }
 	    krb5_data_free(&datack);
--- a/kdc/krb5tgs.c
+++ b/kdc/krb5tgs.c
@@ -809,16 +809,8 @@ tgs_check_authenticator(krb5_context context,
 	ret = KRB5KRB_AP_ERR_INAPP_CKSUM;
 	goto out;
     }
-    /*
-     * according to RFC1510 it doesn't need to be keyed,
-     * but according to the latest draft it needs to.
-     */
-    if (
-#if 0
-!krb5_checksum_is_keyed(context, auth->cksum->cksumtype)
-	||
-#endif
- !krb5_checksum_is_collision_proof(context, auth->cksum->cksumtype)) {
+
+    if (!krb5_checksum_is_collision_proof(context, auth->cksum->cksumtype)) {
 	kdc_log(context, config, 4, "Bad checksum type in authenticator: %d",
 		auth->cksum->cksumtype);
 	ret =  KRB5KRB_AP_ERR_INAPP_CKSUM;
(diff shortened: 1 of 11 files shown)


kdc/krb5tgs.c
```diff
@@ -1113,6 +1105,12 @@ tgs_check_authenticator(krb5_context context,
 	krb5_free_error_message(context, msg);
 	goto out;
     }
+
+    /*
+     * according to RFC1510 it doesn't need to be keyed,
+     * but according to the latest draft it needs to.
+     */
```
lhoward, line 1112: obviously this comment didn't make sense and still doesn't. RFC4120 doesn't say it needs to be keyed, just collision proof FWIW.

lib/krb5/crypto.c
```diff
@@ -485,6 +486,8 @@ create_checksum_iov(krb5_context context,
 	ret = get_checksum_key(context, crypto, usage, ct, &dkey);
 	if (ret)
 	    return ret;
+    } else if ((flags & KRB5_CRYPTO_FLAG_ALLOW_UNKEYED_CHECKSUM) == 0) {
+	return EINVAL;
```
iboukris, line 490: Maybe KRB5KRB_AP_ERR_INAPP_CKSUM?

lhoward: For `create_checksum()`, it's a programming error. For `verify_checksum()`, it's a protocol error. Hence the different error codes :)

iboukris: Got it.

lib/krb5/crypto.c
```diff
@@ -472,6 +472,7 @@ create_checksum_iov(krb5_context context,
 		    unsigned usage,
 		    struct krb5_crypto_iov *iov,
 		    int niov,
+		    krb5_flags flags,
```
iboukris, line 475: Looking at these static create/verify functions, they a,ll seem to be handed the crypto object, do we really need to change their signature, can't they just use crypto_flags(crypto)?

kdc/krb5tgs.c
iboukris: Make sense since it is encrypted anyway, perhaps worth noting what RFC4120 says, the comment might have predated it, who knows :)

lib/krb5/crypto.c
lhoward: We always want to require keyed checksums when `krb5_encrypt()` (and friends) are used, at least for non-DES encryption types. `KRB5_CRYPTO_FLAG_ALLOW_UNKEYED_CHECKSUM` should only apply to `krb5_{create, verify}_checksum()`, not the encryption/decryption functions.

lhoward: i.e. `crypto_flags(crypto)` is honoured only when we're dealing with the checksum API functions (not the encryption ones)

kdc/krb5tgs.c
lhoward: Yeah, I was a bit confused by that. I checked with @nicowilliams and he said the authenticator checksum was more channel bindings, i.e. an extensibility point for protecting additional data (e.g. as used by GSS-API).

lib/krb5/crypto.c
iboukris: Oh I see, that explain the changes in the encryption functions, starting to get it.

kdc/krb5tgs.c
iboukris: That's when used in AP request to a server, in AP request to KDC aka TGS-REQ it uses regular keyed checksum afair.

lhoward: Right, but RFC4120 doesn't say it has to be keyed, only collision-proof. Section 3.3.2.

iboukris: Yeah, again make sense since it is in encrypted envelop, not sure what's the rational to have it to begin with, i'll look at the RFC..

lhoward: I'm actually going to introduce a new `_kdc_verify_checksum()` function that maps `KRB5_PROG_SUMTYPE_NOSUPP` to `KRB5KDC_ERR_SUMTYPE_NOSUPP` to increase our RFC4120 compliance.

lhoward: RFC4120 3.3.2 says the checksum is application specific. I don't even think MIT checks it tbh.

lhoward: Skimming the MIT code, it uses it only for GSS, and that's all within the Kerberos GSS mechanism itself (notwithstanding the machinery for `krb5_auth_con_set_checksum_func()`).

lib/krb5/crypto.c
lhoward: So, `encrypt_internal()` is used (as far as I can tell!) by DES only, so we explicitly set `KRB5_CRYPTO_FLAG_ALLOW_UNKEYED_CHECKSUM` there. Otherwise, `encrypt_internal_derived()` is used for 3DES and AES-SHA1, `encrypt_internal_special()` for RC4, and `encrypt_internal_enc_then_cksum()` for AES-SHA2. These latter functions do not set `KRB5_CRYPTO_FLAG_ALLOW_UNKEYED_CHECKSUM`, although that was never a vulnerability because they explicitly use `et->keyed_checksum` (i.e. the mandatory keyed checksum type is used for the encryption type). Not setting `KRB5_CRYPTO_FLAG_ALLOW_UNKEYED_CHECKSUM` is just good hygiene / defense in depth.

iboukris: Got it, fwiw it all looks good to me, thanks!

lhoward: Thanks for the review! Merged :)

abartlet:
@josephsutton1 has found a possible regression in gssapi CKSUMTYPE_RSA_MD5 as seen here: https://gitlab.com/samba-team/devel/samba/-/commit/2c888792e86f75f89ec4c22783e8c17e7ddaf960

lhoward:
> @josephsutton1 has found a possible regression in gssapi CKSUMTYPE_RSA_MD5 as seen here: https://gitlab.com/samba-team/devel/samba/-/commit/2c888792e86f75f89ec4c22783e8c17e7ddaf960

Fixed in eb293680a8aaaa5dd49c3a32d6177b64e5037d56

PR #839, opened 2021-09-22
Update bits.c to free memory properly

ash1852:
correctly free memory no longer needed in the end of program.


jaltman:
This change is unnecessary.   The program is about to terminate.   There is no need to free this memory that will be released two lines later when the process exits.

PR #840, opened 2021-09-23
Follow-up PAC related tests

iboukris:
CC @lhoward


lhoward:
Principal inside the PAC, do you mean inside the ticket? Or does this comment refer to the recent commit where we verify ticket cname matches logon info?

iboukris:
> Principal inside the PAC, do you mean inside the ticket? Or does this comment refer to the recent commit where we verify ticket cname matches logon info?

The latter, if you revert that commit the test will fail.

iboukris:
Recall that test_ap_req now verifies the PAC by default.

lhoward:
Ah, excellent. Merging!

Issue #816, opened 2021-09-05
kgetcred does not obtain S4U2Proxy ticket if cache already has service ticket for a different client


RobCrowston:
**Describe the bug**
`kgetcred` will not obtain a second service ticket (for the same service but for a different client) using S4U2Proxy impersonation for the second ticket.

**Reproduction**
The realm is configured with three principals:
* client
* service1
* service2
service1 is configured to delegate (via S4U2Proxy) to service2.

We can first obtain a normal (non-delegated) ticket as service1 to service2:

```
$ kinit -c /tmp/service service1
$ kgetcred -c /tmp/service service2   # first obtain ticket for service2 as service1
$ klist -c /tmp/service -v                                                                                           
Credentials cache: FILE:/tmp/service                                                                                                                                                 
        Principal: service1@REALM                                                                                                                                   
    Cache version: 4                                                                                                                                                                 
                                                                                                                                                                                     
Server: krbtgt/REALM@REALM                                                                                                                                   
Client: service1@REALM                                                                                                                                              
Ticket etype: aes256-cts-hmac-sha1-96, kvno 1                                                                                                                                        
Session key: aes128-cts-hmac-sha1-96                                                                                                                                                 
Ticket length: 441                                                                                                                                                                   
Auth time:  Sep  5 14:38:55 2021                                                                                                                                                     
End time:   Sep  6 14:38:55 2021                                                                                                                                                     
Ticket flags: enc-pa-rep, pre-authent, initial, forwardable                                                                                                                          
Addresses: addressless                                                                                                                                                               
                                                                                                                                                                                     
Server: service2@REALM   
Client: service1@REALM
Ticket etype: aes128-cts-hmac-sha1-96, kvno 1                                                                                                                                        
Ticket length: 446                                                                                                                                                                   
Auth time:  Sep  5 14:38:55 2021                                                          
Start time: Sep  5 14:39:05 2021
End time:   Sep  6 14:38:55 2021
Ticket flags: transited-policy-checked, pre-authent, forwardable
Addresses: addressless
```

Then try to obtain the delegated credential as client@REALM.

```
$ kinit client
$ kgetcred --out-cache=/tmp/evidence service1 # obtain ticket as client to service1
$ kgetcred -c /tmp/service --delegation-credential-cache=/tmp/evidence service2 # service1 can use the evidence ticket to obtain a delegated ticket as client to service2
$ klist -c /tmp/service -v                                                                                         
[...]

PR #841, opened 2021-09-27
krb5: skip cache check in S4U2Proxy requests

iboukris:
Fix for #816, turns out there was one more place that needed changing, as the tests show they both needed.


lhoward:
at least it's failing for me on macOS...

iboukris:
Ok, it runs fine on linux, maybe I made some wrong assumptions on the output of klist, mind mailing the whole log, trying to figure  where it fails exactly.

PR #846, opened 2021-10-14
Add placeholder for gss_acquire_cred_impersonate_name().

RobCrowston:
Add placeholder and prototype for `gss_acquire_cred_impersonate_name()`.

The prototype for `gss_acquire_cred_impersonate_name()` is patterned off the [MIT implementation](https://web.mit.edu/kerberos/krb5-devel/doc/appdev/gssapi.html#constrained-delegation-s4u). No functional changes are intended in this patch.

 See issue #4.

### DIFF
--- a/lib/gssapi/Makefile.am
+++ b/lib/gssapi/Makefile.am
@@ -85,6 +85,7 @@ mechsrc = \
 	mech/gss_accept_sec_context.c \
 	mech/gss_acquire_cred.c \
 	mech/gss_acquire_cred_from.c \
+	mech/gss_acquire_cred_impersonate_name.c \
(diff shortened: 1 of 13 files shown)


RobCrowston:
Updated, thanks.

lhoward:
I think you'll need to implement a stub for `gss_acquire_cred_impersonate_name()` to get it to build because `OPTSYM` references the mechglue symbol (loopback detection for mechanisms that do not implement it and re-export the mechglue).

Suggest a stub implementation that returns `GSS_S_NO_CRED`, don't forget to add it to the exports table too for both Windows and Unix.

PR #854, opened 2021-11-15
hdb: Namespace referrals

nicowilliams:


lib/hdb/common.c
```diff
@@ -1307,6 +1306,138 @@ make_namespace_princ(krb5_context context,
...
 }
 
+static int
+is_namespace_princ_p(krb5_context context,
+                     krb5_const_principal princ)
+{
+    return
+        krb5_principal_get_num_comp(context, princ) >= 4
+        && strcmp(krb5_principal_get_comp_string(context, princ, 0),
+                  "WELLKNOWN") == 0
```
lhoward, line 1316: If used in more than one place, perhaps a macro in `krb5.h` would be better.

nicowilliams: I mean, it will never change...

lib/hdb/common.c
```diff
@@ -1307,6 +1306,138 @@ make_namespace_princ(krb5_context context,
...
+                 krb5_const_principal ns_princ,
+                 krb5_const_principal found_ns_princ,
+                 char **s)
+{
+    const char *ns_host_part, *wanted_host_part, *found_host_part;
+    const char *p, *r;
+    size_t ns_host_part_len, wanted_host_part_len;
+
+    wanted_host_part = krb5_principal_get_comp_string(context, wanted_princ, 1);
+    wanted_host_part_len = strlen(wanted_host_part);
```
lhoward, line 1334: we know for sure wanted_princ has at least two components?

lhoward: I mean, four

lib/hdb/common.c
```diff
@@ -1307,6 +1306,138 @@ make_namespace_princ(krb5_context context,
...
+               krb5_const_principal nsprinc,
+               hdb_entry_ex *h)
+{
+    krb5_error_code ret = 0;
+    char *s = NULL;
+
+    if (!nsprinc)
+        return 0;
+    if (krb5_principal_get_num_comp(context, princ) < 2)
+        return HDB_ERR_NOENTRY;
```
lhoward, line 1380: ah, got it.

PR #855, opened 2021-11-15
Always cmp cmp (GCC bug 95189)

nicowilliams:
Because of:

http://r6.ca/blog/20200929T023701Z.html
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95189

this is bad:

```
    if (strcmp(a, b) || !strcmp(c, d)) ...
```

so this patch changes that to:

```
    if (strcmp(a, b) != 0 || strcmp(c, d)) == 0
```

for all C library cmp functions and related:

 - strcmp(), strncmp()
 - strcasecmp(), strncasecmp()
 - stricmp(), strnicmp()
 - memcmp()

### DIFF
--- a/appl/dceutils/k5dcecon.c
+++ b/appl/dceutils/k5dcecon.c
@@ -163,8 +163,8 @@ int k5dcesession(luid, pname, tgt, ppag, tflags)
 /*
  * (but root has the ffffffff which we are not interested in)
  */
-    if (!strncmp(direntp->d_name,"dcecred_",8)
-         && (strlen(direntp->d_name) == 16)) {
+    if (strncmp(direntp->d_name,"dcecred_",8) == 0 &&
+        (strlen(direntp->d_name) == 16)) {
 
       /* looks like a cache name, lets do the stat, etc */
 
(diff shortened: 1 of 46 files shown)


jaltman:
Thanks for doing this.  This change is going to require thorough review to ensure it handles all of the cases.   Including those where the result of memcmp is returned from a function which might be inline and the result is used without comparing to zero.

nicowilliams:
FYI I checked all the `cmp` returning functions's in-tree callers.

mistotebe:
If I understand [this](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95189) correctly, it's only memcmp calls that are affected (they start to behave like strcmp instead), so the strcmp() calls don't need any changes. Additionally only memcmp()s with a string that is known to contain a `\0` in it at compile time?

So the patch looks like a massive overreaction, or am I missing something?

jaltman:
The change is addressing two things at once.  First, is GCC bug 95189.   The other is the general style issue of treating cmp functions as inverted booleans.   Addressing GCC 95189 is a small subset and could/should be broken out separately.

jaltman:
On macOS Apple Silicon tests fail with 5c45be4554154d5643988e049ab52e5a93df08a9 but pass with 327bb6ff39bfcbc50c2d3c844ea7e9197e3bed9a

I'm reviewing 5c45be4554154d5643988e049ab52e5a93df08a9 for mistakes.

jaltman:
All tests pass on 32-bit Windows.   On macOS Apple Silicon all tests except check-iprop pass.    Commit 0e2cc5fd075b2541b9675e57ecba52baa7e2eeeb.

jaltman:
@nicowilliams reports that all tests pass on Linux

PR #863, opened 2021-11-28
s4/heimdal/lib/krb5/pac.c: Align PAC buffers to match Windows

abartlet:
BUG: https://bugzilla.samba.org/show_bug.cgi?id=14642

Signed-off-by: Joseph Sutton <josephsutton@catalyst.net.nz>
Reviewed-by: Andrew Bartlett <abartlet@samba.org>
(cherry picked from Samba commit 28a5a586c8e9cd155d676dcfcb81a2587ace99d1)

### DIFF
--- a/lib/krb5/pac.c
+++ b/lib/krb5/pac.c
@@ -62,10 +62,12 @@ struct krb5_pac_data {
 #define PACTYPE_SIZE			8
 #define PAC_INFO_BUFFER_SIZE		16
 
+#define PAC_LOGON_INFO			1
 #define PAC_SERVER_CHECKSUM		6
 #define PAC_PRIVSVR_CHECKSUM		7
 #define PAC_LOGON_NAME			10
 #define PAC_CONSTRAINED_DELEGATION	11
+#define PAC_UPN_DNS_INFO		12
 #define PAC_TICKET_CHECKSUM		16
 
 #define CHECK(r,f,l)						\
@@ -1200,7 +1202,17 @@ _krb5_pac_sign(krb5_context context,
 		ret = krb5_enomem(context);
 		goto out;
 	    }
-	    /* XXX if not aligned, fill_zeros */
+
+	    if (p->pac->buffers[i].type == PAC_LOGON_INFO
+		|| p->pac->buffers[i].type == PAC_UPN_DNS_INFO)
+	    {
+		uint32_t rounded = (len + PAC_ALIGNMENT - 1) / PAC_ALIGNMENT
+		    * PAC_ALIGNMENT;
+		uint32_t remaining = rounded - len;
+		CHECK(ret, fill_zeros(context, spdata, remaining), out);
+
+		len = rounded;
+	    }
 	}
 
 	/* write header */


lib/krb5/pac.c
```diff
@@ -1200,7 +1202,17 @@ _krb5_pac_sign(krb5_context context,
 		ret = krb5_enomem(context);
 		goto out;
 	    }
-	    /* XXX if not aligned, fill_zeros */
+
+	    if (p->pac->buffers[i].type == PAC_LOGON_INFO
```
lhoward, line 1206: Is there a reason the alignment is only for `PAC_LOGON_INFO` and `PAC_UPN_DNS_INFO`? If so, could you put a reference to the protocol docs in a comment?

josephsutton1: I based this on this bug report, which mentions those buffer types explicitly: [https://bugzilla.samba.org/show_bug.cgi?id=13134](https://bugzilla.samba.org/show_bug.cgi?id=13134)

However, it may be the case that Windows aligns the other PAC buffers as well.

abartlet:
I think it is likely aligned for everything but the other buffers have
natural alignment.

lib/krb5/pac.c
lhoward: Perhaps check with dochelp and, if it turns out they are all aligned, then make this a general case?

metze-samba: As far as I can see this is about padding not alignment.

MS-PAC 2.4 PAC_INFO_BUFFER says:
...
cbBufferSize (4 bytes): A 32-bit unsigned integer in little-endian format that contains the size, in
bytes, of the buffer in the PAC located at Offset.
...
Offset (8 bytes): A 64-bit unsigned integer in little-endian format that contains the offset to the beginning of the buffer, in bytes, from the beginning of the PACTYPE structure (section 2.3). The data offset MUST be a multiple of eight. The following sections specify the format of each type of element.
...


So Offset contains the alignment and that was already aligned to 8 bytes, this change is about
cbBufferSize and I think asking dochelp would be good, with having additional tests (with the python krb5
infrastructure in Samba), we should test with different sAMAccountName length values or account relevant fields
on the account object.

I'm also not sure if this is really the correct place for that change?

Shouldn't p->pac->buffers[i].buffersize already have the padded length?
Which would mean the Samba part should add the padding...

metze-samba: Looking at it I'm wondering if the "/* advance data endpointer and align */" section below is really correct,
I'd guess the alignment part of it should be moved to the top of the loop body, so that we don't add unused padding
at the end of the last loop iteration.

lhoward: Avoiding unneeded padding is good (given ticket size limits). First thing to fix though would be removing special cases for padding.

josephsutton1: I ran tests against Windows and this change seems to match what Windows is doing. LOGON_INFO and UPN_DNS_INFO are padded to a multiple of 8 bytes, while CREDENTIAL_INFO, the checksum types, LOGON_NAME, and REQUESTER_SID are not. The CONSTRAINED_DELEGATION and ATTRIBUTES_INFO blobs seem to already be multiples of 8 bytes.

lhoward: Ah ok, thanks for checking!

metze-samba: > I ran tests against Windows and this change seems to match what Windows is doing. LOGON_INFO and 
> UPN_DNS_INFO are padded to a multiple of 8 bytes, while CREDENTIAL_INFO, the checksum types, LOGON_NAME, 
> and REQUESTER_SID are not. The CONSTRAINED_DELEGATION and ATTRIBUTES_INFO blobs seem to already 
> be multiples of 8 bytes.

Why didn't you change the Samba code to add the padding? I think that would be the correct place, which means we'd also have it for the MIT KDC case...

abartlet: Yeah, I think this is the right spot actually.  Sorry for the stuffing about @lhoward

lhoward:
Do you want me to back it out?

abartlet:
I wasn't going to make a fuss as it at least solved it for now, but that (a revert) would probably be for the best.  

It isn't harmful but you won't want to be maintaining the list of things to pad on the Heimdal side, whereas we can pad in Samba and have good infrastructure for testing that kind of thing.

lhoward:
Or Heimdal could pessimistically pad everything to 8 bytes.

metze-samba:
> Or Heimdal could pessimistically pad everything to 8 bytes.

I'd like to avoid such magic, I think cbBufferSize should not be changed

metze-samba:
I think I found the root cause MS-RPCE 2.2.6.3 Primitive Type Serialization says:
```
For any top-level NDR primitive type, there MUST NOT be any private header preceding the actual type. 
The type MUST be aligned on an 8-byte boundary. If the size of the primitive type is not an integral
multiple of 8 bytes, the data MUST be padded at the end to ensure that its total length is an
integral multiple of 8 bytes.
```

And in Samba we have the related logic in ndr_push_subcontext_end() for case 0xFFFFFC01,
so our PAC_INFO union should be fine for PAC_TYPE_LOGON_INFO as it uses
subcontext(0xFFFFFC01), but PAC_UPN_DNS_INFO uses handmade marshalling.

metze-samba:
I also found _ndr_size_PAC_INFO() in Samba's PAC_BUFFER, which uses NDR_ROUND(size,8) for
LOGON_INFO and UPN_DNS_INFO together with 
[relative,switch_is(type),subcontext(0),subcontext_size(NDR_ROUND(_ndr_size,8)),flag(NDR_ALIGN8)] PAC_INFO *info;
which always adds padding to 8 bytes at the end for all elements (even the last one). The problem seems to
be that Samba's source4/kdc/pac-glue.c uses PAC_INFO only for the individual elements and the logic
of PAC_BUFFER doesn't apply.

See https://git.samba.org/?p=samba.git;a=blob;f=librpc/idl/krb5pac.idl;h=bbe4a253e3a24e7f8ad9950237c9ef058017b4c4;hb=e2b7a2f78117e20739aa4f895ce68825e160d451#l156 for more context.
and
https://git.samba.org/?p=samba.git;a=blob;f=source4/kdc/pac-glue.c;h=2a96a683cd9516fab7376d4bfa517094581b5156;hb=e2b7a2f78117e20739aa4f895ce68825e160d451#l104

I guess we have can have simple function that padds a talloc()'ed DATA_BLOB to a given boundary
and calls that on 'upn_data' after:
```
 135         ndr_err = ndr_push_union_blob(upn_data, mem_ctx, &pac_upn,
 136                                       PAC_TYPE_UPN_DNS_INFO,
 137                                       (ndr_push_flags_fn_t)ndr_push_PAC_INFO);
```

metze-samba:
So the code in heimdal was correct before...

lhoward:
Reverted.

PR #864, opened 2021-11-28
kdc: add krb5plugin_windc_pac_pk_generate() hook

abartlet:
This allows PAC_CRENDENTIAL_INFO to be added to the PAC
when using PKINIT. In that case PAC_CRENDENTIAL_INFO contains
an encrypted PAC_CRENDENTIAL_DATA.

BUG: https://bugzilla.samba.org/show_bug.cgi?id=11441

(similar to Samba commit 0022ea9efb0e7809fa2d060b294320eb0479cdd2)

### DIFF
--- a/kdc/kdc_locl.h
+++ b/kdc/kdc_locl.h
@@ -79,6 +79,7 @@ struct astgs_request_desc {
     EncKDCRepPart ek;
 
     /* PA methods can affect both the reply key and the session key (pkinit) */
+    krb5_preauthtype validated_pa_type;
     krb5_enctype sessionetype;
     krb5_keyblock reply_key;
     krb5_keyblock session_key;
--- a/kdc/kerberos5.c
+++ b/kdc/kerberos5.c
@@ -1787,28 +1787,28 @@ _kdc_check_anon_policy(astgs_request_t r)
  */
 
 static krb5_boolean
-send_pac_p(krb5_context context, KDC_REQ *req)
+check_pa_pac_request(krb5_context context, KDC_REQ *req, krb5_boolean *include_pac)
 {
     krb5_error_code ret;
     PA_PAC_REQUEST pacreq;
     const PA_DATA *pa;
     int i = 0;
 
+    *include_pac = TRUE;
+
     pa = _kdc_find_padata(req, &i, KRB5_PADATA_PA_PAC_REQUEST);
     if (pa == NULL)
-	return TRUE;
+	return KRB5KDC_ERR_PADATA_TYPE_NOSUPP;
 
     ret = decode_PA_PAC_REQUEST(pa->padata_value.data,
 				pa->padata_value.length,
 				&pacreq,
 				NULL);
     if (ret)
-	return TRUE;
-    i = pacreq.include_pac;
+	return KRB5KDC_ERR_PADATA_TYPE_NOSUPP;
+    *include_pac = pacreq.include_pac;
     free_PA_PAC_REQUEST(&pacreq);
-    if (i == 0)
-	return FALSE;
-    return TRUE;
+    return 0;
 }
 
 /*
@@ -1819,12 +1819,31 @@ static krb5_error_code
 generate_pac(astgs_request_t r, const Key *skey, const Key *tkey)
 {
     krb5_error_code ret;
+    const krb5_keyblock *pk_reply_key = NULL;
     krb5_pac p = NULL;
     krb5_data data;
     uint16_t rodc_id;
     krb5_principal client;
+    krb5_boolean sent_pac_request;
+    krb5_boolean pac_request;
 
-    ret = _kdc_pac_generate(r->context, r->client, &p);
+    ret = check_pa_pac_request(r->context, &r->req, &pac_request);
+
+    sent_pac_request = (ret == 0);
+
+    switch (r->validated_pa_type) {
+    case KRB5_PADATA_PK_AS_REQ:
+    case KRB5_PADATA_PK_AS_REQ_WIN:
+	pk_reply_key = &r->reply_key;
+	break;
+    }
+
+    ret = _kdc_pac_generate(r->context,
+			    r->client,
+			    r->server,
+			    pk_reply_key,
+			    sent_pac_request ? &pac_request : NULL,
+			    &p);
     if (ret) {
 	_kdc_r_log(r, 4, "PAC generation failed for -- %s",
 		   r->cname);
@@ -1986,7 +2005,6 @@ _kdc_as_rep(astgs_request_t r)
     krb5_enctype setype;
     krb5_error_code ret = 0;
     Key *skey;
-    int found_pa = 0;
     int i, flags = HDB_F_FOR_AS_REQ;
     METHOD_DATA error_method;
     const PA_DATA *pa;
@@ -2141,7 +2159,7 @@ _kdc_as_rep(astgs_request_t r)
 
 	/* Check if preauth matching */
 
-	for (n = 0; !found_pa && n < sizeof(pat) / sizeof(pat[0]); n++) {
+	for (n = 0; r->validated_pa_type == KRB5_PADATA_NONE && n < sizeof(pat) / sizeof(pat[0]); n++) {
 	    if (pat[n].validate == NULL)
 		continue;
 	    if (r->armor_crypto == NULL && (pat[n].flags & PA_REQ_FAST))
@@ -2183,13 +2201,13 @@ _kdc_as_rep(astgs_request_t r)
 		kdc_log(context, config, 4,
 			"%s pre-authentication succeeded -- %s",
 			pat[n].name, r->cname);
-		found_pa = 1;
 		r->et.flags.pre_authent = 1;
+		r->validated_pa_type = pat[n].type;
 	    }
 	}
     }
 
-    if (found_pa == 0) {
+    if (r->validated_pa_type == KRB5_PADATA_NONE) {
 	Key *ckey = NULL;
 	size_t n;
 	krb5_boolean default_salt;
@@ -2561,7 +2579,7 @@ _kdc_as_rep(astgs_request_t r)
     }
 
     /* Add the PAC */
-    if (send_pac_p(context, req) && !r->et.flags.anonymous) {
+    if (!r->et.flags.anonymous) {
 	generate_pac(r, skey, krbtgt_key);
     }
 
--- a/kdc/windc.c
+++ b/kdc/windc.c
@@ -71,7 +71,10 @@ krb5_kdc_windc_init(krb5_context context)
 
 struct generate_uc {
     hdb_entry_ex *client;
+    hdb_entry_ex *server;
+    const krb5_keyblock *pk_reply_key;
     krb5_pac *pac;
+    const krb5_boolean *pac_request;
 };
 
 static krb5_error_code KRB5_LIB_CALL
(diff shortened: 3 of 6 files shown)


kdc/kdc_locl.h
```diff
@@ -79,6 +79,7 @@ struct astgs_request_desc {
     EncKDCRepPart ek;
 
     /* PA methods can affect both the reply key and the session key (pkinit) */
+    int validated_pa_type;
```
lhoward, line 82: Can you make this

`unsigned int validated_pa_type : 1`

and put it immediately above `armor_crypto`, this will be more consistent when we merge lukeh/gss-preauth-apple

lhoward: Ignore that, sorry, I didn't realise this was not a boolean. In which case, this should be `krb5_preauthtype pa_type_validated`.

kdc/kerberos5.c
```diff
@@ -2185,6 +2194,7 @@ _kdc_as_rep(astgs_request_t r)
 			pat[n].name, r->cname);
 		found_pa = 1;
 		r->et.flags.pre_authent = 1;
+		r->validated_pa_type = pat[n].type;
```
lhoward, line 2197: If `found_pa` is now equivalent to `pa_type_validated != KRB5_PADATA_NONE`, then it should be removed. DRY.

kdc/windc.c
```diff
@@ -82,13 +83,20 @@ generate(krb5_context context, const void *plug, void *plugctx, void *userctx)
 
     if (ft->pac_generate == NULL)
 	return KRB5_PLUGIN_NO_HANDLE;
+
+    if (ft->pac_pk_generate != NULL && uc->pk_reply_key != NULL) {
+	return ft->pac_pk_generate((void *)plug, context,
+				   uc->client, uc->pk_reply_key, uc->pac);
+    }
+
```
lhoward, line 91: Can't we just update the signature of `pac_generate` and bump the plugin interface version number? Seems cleaner.

kdc/windc_plugin.h
```diff
@@ -74,7 +81,7 @@ typedef krb5_error_code
 	KDC_REQ *, METHOD_DATA *);
 
 
-#define KRB5_WINDC_PLUGIN_MINOR			6
+#define KRB5_WINDC_PLUGIN_MINOR			7
```
lhoward, line 84: Right, you've changed the ABI version so, you can change the ABI, right?

abartlet:
Thanks @lhoward for the feedback.

I'll leave this open for now so I don't re-upload this again without addressing these.  The patch is one of many I've been slowly trying to upstream, but also one of the few that isn't tangled up in the FAST changes.   I'll have some more to triage soon once we make some headway on reviewing that.

lhoward:
I am being picky but I reckon it's also good to keep this as clean as possible, particularly given the code doesn't get well tested outside Samba.

abartlet:
I totally agree.  No hard feelings, just trying to merge what is easy to accept now and then I'll try and clean house again next year. 

I have a deadline to upgrade our Heimdal fork by the end of this year (ideally) and the less we leave in lorikeet-heimdal the less I have to keep rebasing, but I also can't put much time into polishing each patch right now, particularly when they are the work of others that I've just been carrying forward.

abartlet:
I've built a new commit with all our changes to this PAC API in one commit, so while I'm hesitant to change the pkinit logic which I barely understand I can fix up the API as requested. I'll push once I've tested it in a Samba build.

jaltman:
I have not looked at this code carefully but the calling convention for all of the plug-in functions must be KRB5_CALLCONV.

abartlet:
Thanks @lhoward for pushing back on this.  On Samba's side we were literally calling the same function anyway, as you might have expected: https://gitlab.com/samba-team/samba/-/blob/master/source4/kdc/wdc-samba4.c#L106

Likely this was a hold-over from a fantasy world where Samba supported multiple system Heimdal versions (we only ever really supported our in-tree vendored copy)

kdc/kerberos5.c
```diff
@@ -1787,27 +1787,30 @@ _kdc_check_anon_policy(astgs_request_t r)
  */
 
 static krb5_boolean
-send_pac_p(krb5_context context, KDC_REQ *req)
+send_pac_p(krb5_context context, KDC_REQ *req, krb5_boolean *pac_request)
 {
```
lhoward, line 1791: Sorry to be picky, but two booleans is confusing. Perhaps something like this:

```c
static int
check_pa_pac_request(krb5_context context,
                     KDC_REQ *req,
                     krb5_boolean *include_pac)
{
    krb5_error_code ret;
    PA_PAC_REQUEST pacreq;
    const PA_DATA *pa;
    int i = 0;

    *include_pac = TRUE;

    pa = _kdc_find_padata(req, &i, KRB5_PADATA_PA_PAC_REQUEST);
    if (pa == NULL)
        return KRB5KDC_ERR_PADATA_TYPE_NOSUPP;

    ret = decode_PA_PAC_REQUEST(pa->padata_value.data,
                                pa->padata_value.length,
                                &pacreq,
                                NULL);
    if (ret)
        return KRB5KDC_ERR_PADATA_TYPE_NOSUPP;
    *include_pac = pacreq.include_pac;
    free_PA_PAC_REQUEST(&pacreq);
    return 0;
}
```

kdc/windc_plugin.h
abartlet: Done.

kdc/kdc_locl.h
abartlet: I've changed the type

kdc/kerberos5.c
abartlet: Thanks @josephsutton1 for fixing this up.

abartlet:
> I have not looked at this code carefully but the calling convention for all of the plug-in functions must be KRB5_CALLCONV.

There is no change to calling conventions, they use `KRB5_CALLCONV`

abartlet:
I think I've addressed the concerns but it might make more sense to land #805 first.  We have a set of these based on that branch available when that lands.

abartlet:
This version, after the fix to change the type of `r->validated_pa_type` gives:

```
../../source4/heimdal/kdc/kerberos5.c: In function ‘generate_pac’:
../../source4/heimdal/kdc/kerberos5.c:1854:5: error: enumeration value ‘KRB5_PADATA_NONE’ not handled in switch [-Werror=switch]
 1854 |     switch (r->validated_pa_type) {
      |     ^~~~~~
```
So best if we land #805 first if at all possible as our tested patches are on top of that (due conflicts, particularly with the requested `context` -> `r->context` change)

kdc/kerberos5.c
```diff
@@ -1787,28 +1787,28 @@ _kdc_check_anon_policy(astgs_request_t r)
  */
 
 static krb5_boolean
```
lhoward, line 1789: OK, now you also need to change this type to `krb5_error_code` :)

kdc/kerberos5.c
```diff
@@ -1819,12 +1819,31 @@ static krb5_error_code
...
+    ret = check_pa_pac_request(r->context, &r->req, &pac_request);
+
+    sent_pac_request = (ret == 0);
+
+    switch (r->validated_pa_type) {
+    case KRB5_PADATA_PK_AS_REQ:
+    case KRB5_PADATA_PK_AS_REQ_WIN:
+	pk_reply_key = &r->reply_key;
+	break;
+    }
```
lhoward, line 1839: This is for including NTLM keys when the client doesn't authenticate with long-term keys, right? In this case I think we actually want a flag on the PA mechanism that indicates a long-term key was used, e.g. `PA_LONG_TERM_KEY`. This should be set on `ENC-TS` and `ENC-CHAL`. Then lets store a flag in the request:

```c
unsigned int pa_validated_long_term_key : 1;
```

where `found_pa` would have been set, and pass `r->pa_validated_long_term_key ? NULL : &r->reply_key` as an argument to `_kdc_pac_generate()`.

Eliminate the temporary `pk_reply_key` variable and `validated_pa_type`.

tests/plugin/windc.c
```diff
@@ -20,11 +20,20 @@ windc_fini(void *ctx)
...
+	     struct hdb_entry_ex *server,
+	     const krb5_keyblock *pk_replykey,
+	     const krb5_boolean *pac_request,
+	     krb5_pac *pac)
 {
     krb5_error_code ret;
     krb5_data data;
 
+    if (pac_request != NULL && *pac_request == FALSE) {
+	    *pac = NULL;
```
lhoward, line 33: just checking indentation here

kdc/kerberos5.c
lhoward: Or, if you want, invert the logic and make the flag `PA_REPLACE_REPLY_KEY` (on PKINIT and GSS), and propagate when validated to `r->pa_replaced_reply_key`. Then, if *that* is set, you ask the backend to include the symmetries keys.

lhoward: `PA_REPLACE_REPLY_KEY` is a little more semantically consistent with RFC6113.

abartlet:
Is it OK if I push this back on top of #805 or will that make the review horrible?  Just trying to ensure I push up the same code I test, which will be in Samba on top of that.

lhoward:
I think if it's on top of it, it should be fine, because @nicowilliams will hopefully have that merged soon and you can just rebase back onto master

kdc/kerberos5.c
abartlet: The variable is communicating two things:
 * the key to encrypt the NTLM hash with
 * that an NTLM hash should be returned

A better design might be to always pass in the reply_key and the PA data and let the HDB module choose what PA types should trigger the NTLM reply.  Would that be OK?

lhoward: I think either is fine. The advantage of doing it in the front end is that it's resilient to new PA types and, perhaps we want to flag which PA types replace the reply key anyway, for some other unspecified future reason. But I'm also inclined to take the patch that changes the fewest lines of code ;)

abartlet: It isn't replacing the reply key, it is using it.

lhoward: See [RFC 6113 Section 3.3](https://datatracker.ietf.org/doc/html/rfc6113.html#section-3.3) and [RFC 4456 Section 3.2.3.2](https://datatracker.ietf.org/doc/html/rfc4556#section-3.2.3.2).

PKINIT and GSS are examples of two pre-authentication mechanisms that replace the reply key.

Mechanisms that replace the reply key by definition do not encrypt the reply in the client's long-term key. For downlevel compatibility Windows needs to include the NTLM OWFs for NTLM compatibility (of course, this case the OWFs can be randomly generated, as the user may never have a password set).

abartlet: Thanks, now I get it.  Now I get why a flag up at the PA types level makes sense here.

lhoward: Another point just for posterity: the KDC is implicitly authenticated to the client when the reply key is *not* replaced, because it’s proved knowledge of the client’s long term key. When a pre-authentication mechanism replaces the reply key, the client needs another way of authenticating the KDC: validating its certificate, checking GSS_C_MUTUAL_FLAG, etc.

lhoward: Do you want me to prepare a patch for this based on the proposed solution?

lhoward: I have a patch, will file a separate PR

lhoward:
@abartlet closing this as duplicate of #878, please review latter

PR #865, opened 2021-11-29
Use UTF-8 in KTH copyright notice

abartlet:
Samba is starting to protect against bi-di attacks and the starting point
is to require that input files be fully UTF-8.  In 2021 this is a reasonable
starting point anyway.

### DIFF
--- a/cf/maybe-valgrind.sh
+++ b/cf/maybe-valgrind.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 #
-# Copyright (c) 2006 - 2007 Kungliga Tekniska Högskolan
+# Copyright (c) 2006 - 2007 Kungliga Tekniska HÃ¶gskolan
 # (Royal Institute of Technology, Stockholm, Sweden). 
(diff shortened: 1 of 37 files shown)


lhoward:
LGTM

lhoward:
This broke `test_cms`, my bad for approving it without checking tests ran, I've reverted this commit on `lib/hx509/data/static-file`; feel free to propose a better fix.

abartlet:
Sorry about that.  I feel very lost without CI.  

I've since worked out that test_nist{,2} isn't spinning, it is just
slow, so one way or another I need to get 'make test' running in a CI
service.

I really didn't expect what looks like a shell script to be a
fixture...

Anyway, a revert is fine, I work around this here:
https://gitlab.com/samba-team/devel/samba/-/commit/de8e5835e64b990bf3d4931c8cf5e893ed688979

Andrew Bartlett

jaltman:
Reverted by 8a540962667a0674e95b9cf8be8df1c66b060538

PR #868, opened 2021-12-03
Upstream Heimdal-side helpers for Samba's Nov 2021 security release

abartlet:
This proposes the patches that were made to Samba to address the November 2021 Security issues 

See https://www.samba.org/samba/security/CVE-2020-25719.html

We made changes to our (older) copy of Heimdal to address these issues and @josephsutton1 has now forward-ported those to modern Heimdal. 

(I've written the new longer commit messages)

### DIFF
--- a/kdc/krb5tgs.c
+++ b/kdc/krb5tgs.c
@@ -349,17 +349,23 @@ check_constrained_delegation(krb5_context context,
 /*
  * Determine if s4u2self is allowed from this client to this server
  *
+ * also:
+ *
+ * Check that the client (user2user TGT, enc-tkt-in-skey) hosts the
+ * service given by the client.
+ *
  * For example, regardless of the principal being impersonated, if the
- * 'client' and 'server' (target) are the same, then it's safe.
+ * 'client' and 'server' (target) are the same, or server is an SPN
+ * alias of client, then it's safe.
  */
 
 static krb5_error_code
-check_s4u2self(krb5_context context,
-	       krb5_kdc_configuration *config,
-	       HDB *clientdb,
-	       hdb_entry_ex *client,
-	       hdb_entry_ex *target_server,
-	       krb5_const_principal target_server_principal)
+check_client_matches_target_service(krb5_context context,
+				    krb5_kdc_configuration *config,
+				    HDB *clientdb,
+				    hdb_entry_ex *client,
+				    hdb_entry_ex *target_server,
+				    krb5_const_principal target_server_principal)
 {
     krb5_error_code ret;
 
@@ -368,11 +374,11 @@ check_s4u2self(krb5_context context,
      * policy or audit check and can look into the DB records
      * directly
      */
-    if (clientdb->hdb_check_s4u2self) {
-	ret = clientdb->hdb_check_s4u2self(context,
-					   clientdb,
-					   client,
-					   target_server);
+    if (clientdb->hdb_check_client_matches_target_service) {
+	ret = clientdb->hdb_check_client_matches_target_service(context,
+								clientdb,
+								client,
+								target_server);
 	if (ret == 0)
 	    return 0;
     } else if (krb5_principal_compare(context,
@@ -1314,6 +1320,10 @@ eout:
     return ENOMEM;
 }
 
+/*
+ * This function is intended to be used when failure to find the client is
+ * acceptable.
+ */
 static krb5_error_code
 db_fetch_client(krb5_context context,
 		krb5_kdc_configuration *config,
@@ -1382,9 +1392,12 @@ tgs_build_reply(astgs_request_t priv,
     krb5_error_code ret, ret2;
     krb5_principal cp = NULL, sp = NULL, rsp = NULL, tp = NULL, dp = NULL;
     krb5_principal krbtgt_out_principal = NULL;
+    krb5_principal user2user_princ = NULL;
     char *spn = NULL, *cpn = NULL, *tpn = NULL, *dpn = NULL, *krbtgt_out_n = NULL;
+    char *user2user_name = NULL;
     hdb_entry_ex *server = NULL, *client = NULL, *s4u2self_impersonated_client = NULL;
     HDB *clientdb, *s4u2self_impersonated_clientdb;
+    HDB *serverdb = NULL;
     krb5_realm ref_realm = NULL;
     EncTicketPart *tgt = &ticket->ticket;
     const EncryptionKey *ekey;
@@ -1467,7 +1480,7 @@ server_lookup:
     server = NULL;
     ret = _kdc_db_fetch(context, config, sp,
                         HDB_F_GET_SERVER | HDB_F_DELAY_NEW_KEYS | flags,
-			NULL, NULL, &server);
+			NULL, &serverdb, &server);
     priv->server = server;
     if (ret == HDB_ERR_NOT_FOUND_HERE) {
 	kdc_log(context, config, 5, "target %s does not have secrets at this KDC, need to proxy", spn);
@@ -1646,6 +1659,7 @@ server_lookup:
 	    krb5uint32 second_kvno = 0;
 	    krb5uint32 *kvno_ptr = NULL;
 	    size_t i;
+	    hdb_entry_ex *user2user_client = NULL;
 
 	    if(b->additional_tickets == NULL ||
 	       b->additional_tickets->len == 0){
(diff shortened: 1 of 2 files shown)


lib/hdb/hdb.h
```diff
@@ -294,7 +294,7 @@ typedef struct HDB {
     /**
      * Check if s4u2self is allowed from this client to this server
      */
-    krb5_error_code (*hdb_check_s4u2self)(krb5_context, struct HDB *, hdb_entry_ex *, krb5_const_principal);
+    krb5_error_code (*hdb_check_s4u2self)(krb5_context, struct HDB *, hdb_entry_ex *, hdb_entry_ex *);
```
lhoward, line 297: Check 5aaf12351aa30ae2e9c657080b8566397c0da6bd hasn't made its way into any releases, otherwise you'll also need to bump the HDB interface version.

abartlet:
Thanks for thinking about interface versions (not something we had to
worry about on Samba's side so I totally forgot).  I can't see that
commit in any release branches.

abartlet:
Is there anything more we need to do on this?

kdc/krb5tgs.c
```diff
@@ -1379,78 +1394,7 @@ tgs_build_reply(astgs_request_t priv,
...
-	if (ret) {
-            _kdc_audit_addreason((kdc_request_t)priv,
-                                 "User-to-user TGT expired or invalid");
-	    goto out;
-        }
-
-	s = &adtkt.cname;
-	r = adtkt.crealm;
-    } else if (s == NULL) {
+    if (s == NULL) {
```
lhoward, line 1397: Isn't an empty server allowed for U2U?

kdc/krb5tgs.c
```diff
@@ -1620,7 +1604,144 @@ tgs_build_reply(astgs_request_t priv,
 	krb5_enctype etype;
 
 	if(b->kdc_options.enc_tkt_in_skey) {
+	    Ticket *t;
```
lhoward, line 1607: You just moved this code around, rather than changing it, right?

abartlet: That was the intention, yes.

kdc/krb5tgs.c
abartlet: I agree the draft at https://datatracker.ietf.org/doc/html/draft-ietf-cat-user2user-02 says it is optional however on Windows  our tests show it gives KDC_ERR_GENERIC  or  KDC_ERR_S_PRINCIPAL_UNKNOWN (probably GENERIC, MS is not good with error codes). https://gitlab.com/samba-team/samba/-/commit/26480ba2aa9834a24f1ea11ae3f8e2d7ed0ccfd8

@josephsutton1 may be able to shed some light on the tests as they are his tests.

josephsutton1: Yes, it gives a GENERIC error on Windows. The commit 'Check name in request against name in user-to-user TGT' states that the reason to require the server name is so it can be checked against the client name in the ticket.

lhoward: OK, it looks like MIT also requires the server principal to be non-NULL in all cases.

lhoward (approved):
LGTM (could cleanup the grammar in the commit message but, whatever)

lhoward (changes requested):
Can you just remove extraneous brace and check indentation so it matches rest of code

lhoward (approved):

lhoward (commented):
RFC4120 is the spec, the I-D you linked to is the proposed GSS mechanism.

kdc/krb5tgs.c
```diff
@@ -1671,6 +1675,54 @@ tgs_build_reply(astgs_request_t priv,
 		goto out;
 	    }
 
+	    /* Fetch the name from the TGT. */
+	    ret = _krb5_principalname2krb5_principal(context, &user2user_princ,
+						     adtkt.cname, adtkt.crealm);
+	    if (ret) {
+		goto out;
+	    }
```
lhoward, line 1683: remove extraneous braces

kdc/krb5tgs.c
```diff
@@ -1671,6 +1675,54 @@ tgs_build_reply(astgs_request_t priv,
...
+
+	    ret = krb5_unparse_name(context, user2user_princ, &user2user_name);
+	    if (ret) {
```
lhoward, line 1686: remove extraneous braces

kdc/krb5tgs.c
```diff
@@ -1671,6 +1675,54 @@ tgs_build_reply(astgs_request_t priv,
...
+		goto out;
+	    }
+
+	    /* Look up the name given in the TGT in the database. */
+	    ret = db_fetch_client(context, config, flags, user2user_princ, user2user_name,
+				  krb5_principal_get_realm(context, krbtgt_out->entry.principal),
+				  NULL, &user2user_client);
+	    if (ret) {
```
lhoward, line 1694: remove extraneous braces

kdc/krb5tgs.c
abartlet: sure, I can do that.

kdc/krb5tgs.c
```diff
@@ -1671,6 +1675,54 @@ tgs_build_reply(astgs_request_t priv,
...
+		if (ret) {
+		    _kdc_free_ent(context, user2user_client);
+		    goto out;
+		}
+
+		/*
+		 * Also check that the account is the same one specified in the
+		 * request.
+		 */
+		ret = check_s4u2self(context, config, serverdb, server, user2user_client, user2user_princ);
```
lhoward, line 1716: calling this `check_s4u2self` even if it does the ”right thing” is pretty confusing... can you rename the function and document what it's actually doing?

kdc/krb5tgs.c
```diff
@@ -2274,6 +2327,8 @@ tgs_build_reply(astgs_request_t priv,
     if(s4u2self_impersonated_client)
 	_kdc_free_ent(context, s4u2self_impersonated_client);
 
+    if (user2user_princ)
+        krb5_free_principal(context, user2user_princ);
```
lhoward, line 2352: it's safe to call `krb5_free_principal(NULL)` so you can remove this if

kdc/krb5tgs.c
```diff
@@ -1671,6 +1675,54 @@ tgs_build_reply(astgs_request_t priv,
...
+				  krb5_principal_get_realm(context, krbtgt_out->entry.principal),
+				  NULL, &user2user_client);
+	    if (ret) {
+		goto out;
+	    }
+
+	    if (user2user_client != NULL) {
+		/*
+		 * If the account is present in the database, check the account
+		 * flags.
```
lhoward, line 1701: if `user2user_client == NULL`, what happens? I think we need to check `adtkt->cname` matches the request server.

kdc/krb5tgs.c
lhoward: After discussing with @nicowilliams, yes, we do need to check that the additional ticket clients matches the server name in all cases.

lhoward (approved):

lhoward:
LGTM except for the comments as noted, @nicowilliams is going to have a look tomorrow as well, because of the nature of the changes

kdc/krb5tgs.c
abartlet: Other code rightly assumes that if `ret == 0` then the equivalent of `user2user_client` is non-NULL so I'll remove the check.

kdc/krb5tgs.c
abartlet: Sure.

abartlet:
NOTE:  This triggers a build failure if #805 is rebased on top.  I'm assuming #805 is landing shortly so I'll rebase on that and re-submit today, unless you want the other way around.

lhoward:
#805 merged

abartlet:
Rebased on master now that #805 is in.

kdc/krb5tgs.c
abartlet: Actually I stuffed up, we really don't want a case where `user2user_client == NULL` and so we have changed to not use the wrapper that allows `user2user_client == NULL`.  Thanks for the eagle eye!

lhoward:
LGTM again but, let's get @nicowilliams to look at.

josephsutton1:
Rebased on master.

abartlet:
I'll get the Samba CVE- prefix stripped here too.

kdc/krb5tgs.c
```diff
@@ -1449,7 +1449,10 @@ tgs_build_reply(astgs_request_t priv,
 	    ret = KRB5KDC_ERR_POLICY;
 	    goto out;
 	}
-	_krb5_principalname2krb5_principal(context, &p, t->sname, t->realm);
+	ret = _krb5_principalname2krb5_principal(context, &p, t->sname, t->realm);
+	if (ret) {
+	    goto out;
```
lhoward: Braces not required

kdc/krb5tgs.c
```diff
@@ -1777,7 +1777,27 @@ tgs_build_reply(astgs_request_t priv,
 		goto out;
 	    }
 
+	    /* Verify the PAC of the TGT. */
+	    ret = check_PAC(context, config, user2user_princ, NULL,
```
lhoward, line 1781: Does need to be behind require_pac as well?

abartlet: `check_PAC()` (now `_kdc_check_pac()`) does that.

josephsutton1: The PAC will still always be required for user-to-user due to the `user2user_pac == NULL || !user2user_kdc_issued` check. For the FAST armor ticket check I changed a similar check to `(r->config->require_pac && !mspac) || (mspac && !ad_kdc_issued)`.

abartlet:
Thanks so much!

PR #869, opened 2021-12-03
Address undefined behaviour

abartlet:
While most of the things fixed here have a long history of consistent behaviour, we all know how much the C compiler authors love to jump on any technically undefined behaviour.

Thankfully tools exist to find such things, and @josephsutton1 spent some time with such tools on Heimdal, resulting in these patches.

### DIFF
--- a/kdc/krb5tgs.c
+++ b/kdc/krb5tgs.c
@@ -787,7 +787,9 @@ out:
 	free(et.authorization_data);
     }
     free_LastReq(&ek.last_req);
-    memset(et.key.keyvalue.data, 0, et.key.keyvalue.length);
+    if (et.key.keyvalue.data != NULL) {
+	memset(et.key.keyvalue.data, 0, et.key.keyvalue.length);
+    }
     free_EncryptionKey(&et.key);
     return ret;
 }
--- a/kdc/krb5tgs.c
+++ b/kdc/krb5tgs.c
@@ -788,7 +788,7 @@ out:
     }
     free_LastReq(&ek.last_req);
     if (et.key.keyvalue.data != NULL) {
-	memset(et.key.keyvalue.data, 0, et.key.keyvalue.length);
+	memset_s(et.key.keyvalue.data, 0, et.key.keyvalue.length);
     }
     free_EncryptionKey(&et.key);
     return ret;
--- a/lib/asn1/der_copy.c
+++ b/lib/asn1/der_copy.c
@@ -103,7 +103,9 @@ der_copy_printable_string (const heim_printable_string *from,
     to->data   = malloc(to->length + 1);
     if(to->data == NULL)
 	return ENOMEM;
-    memcpy(to->data, from->data, to->length);
+    if (from->data != NULL) {
+	memcpy(to->data, from->data, to->length);
+    }
     ((char *)to->data)[to->length] = '\0';
     return 0;
 }
@@ -122,7 +124,9 @@ der_copy_bmp_string (const heim_bmp_string *from, heim_bmp_string *to)
     to->data   = malloc(to->length * sizeof(to->data[0]));
     if(to->length != 0 && to->data == NULL)
 	return ENOMEM;
-    memcpy(to->data, from->data, to->length * sizeof(to->data[0]));
+    if (to->data != NULL && from->data != NULL) {
+	memcpy(to->data, from->data, to->length * sizeof(to->data[0]));
+    }
     return 0;
 }
 
@@ -134,7 +138,9 @@ der_copy_universal_string (const heim_universal_string *from,
     to->data   = malloc(to->length * sizeof(to->data[0]));
     if(to->length != 0 && to->data == NULL)
 	return ENOMEM;
-    memcpy(to->data, from->data, to->length * sizeof(to->data[0]));
+    if (to->data != NULL && from->data != NULL) {
+	memcpy(to->data, from->data, to->length * sizeof(to->data[0]));
+    }
     return 0;
 }
 
@@ -152,7 +158,9 @@ der_copy_octet_string (const heim_octet_string *from, heim_octet_string *to)
     to->data   = malloc(to->length);
     if(to->length != 0 && to->data == NULL)
 	return ENOMEM;
-    memcpy(to->data, from->data, to->length);
+    if (to->data != NULL && from->data != NULL) {
+	memcpy(to->data, from->data, to->length);
+    }
     return 0;
 }
 
@@ -163,7 +171,9 @@ der_copy_heim_integer (const heim_integer *from, heim_integer *to)
     to->data   = malloc(to->length);
     if(to->length != 0 && to->data == NULL)
 	return ENOMEM;
-    memcpy(to->data, from->data, to->length);
+    if (to->data != NULL && from->data != NULL) {
+	memcpy(to->data, from->data, to->length);
+    }
     to->negative = from->negative;
     return 0;
 }
@@ -175,8 +185,10 @@ der_copy_oid (const heim_oid *from, heim_oid *to)
     to->components = malloc(to->length * sizeof(*to->components));
     if (to->length != 0 && to->components == NULL)
 	return ENOMEM;
-    memcpy(to->components, from->components,
-	   to->length * sizeof(*to->components));
+    if (to->components != NULL && from->components != NULL) != 0) {
+	memcpy(to->components, from->components,
+	       to->length * sizeof(*to->components));
+    }
     return 0;
 }
 
@@ -190,6 +202,8 @@ der_copy_bit_string (const heim_bit_string *from, heim_bit_string *to)
     to->data   = malloc(len);
     if(len != 0 && to->data == NULL)
 	return ENOMEM;
-    memcpy(to->data, from->data, len);
+    if (to->data != NULL && from->data != NULL) {
+	memcpy(to->data, from->data, len);
+    }
     return 0;
 }
--- a/lib/asn1/der_get.c
+++ b/lib/asn1/der_get.c
@@ -326,7 +326,9 @@ der_get_octet_string (const unsigned char *p, size_t len,
     data->data = malloc(len);
     if (data->data == NULL && data->length != 0)
 	return ENOMEM;
-    memcpy (data->data, p, len);
+    if (data->data != NULL) {
+	memcpy (data->data, p, len);
+    }
     if(size) *size = len;
     return 0;
 }
--- a/lib/asn1/der_put.c
+++ b/lib/asn1/der_put.c
@@ -237,7 +237,9 @@ der_put_general_string (unsigned char *p, size_t len,
     if (len < slen)
 	return ASN1_OVERFLOW;
     p -= slen;
-    memcpy (p+1, *str, slen);
+    if (*str != NULL) {
+	memcpy (p+1, *str, slen);
+    }
     *size = slen;
     return 0;
 }
@@ -313,7 +315,9 @@ der_put_octet_string (unsigned char *p, size_t len,
     if (len < data->length)
 	return ASN1_OVERFLOW;
     p -= data->length;
-    memcpy (p+1, data->data, data->length);
+    if (data->data) {
+	memcpy (p+1, data->data, data->length);
+    }
     *size = data->length;
     return 0;
 }
@@ -633,7 +637,9 @@ der_put_bit_string (unsigned char *p, size_t len,
 	return ASN1_OVERFLOW;
     p -= data_size + 1;
 
-    memcpy (p+2, data->data, data_size);
+    if (data->data != NULL) {
+	memcpy (p+2, data->data, data_size);
+    }
     if (data->length && (data->length % 8) != 0)
 	p[1] = 8 - (data->length % 8);
     else
--- a/lib/base/heimbase.c
+++ b/lib/base/heimbase.c
@@ -86,10 +86,12 @@ struct heim_auto_release {
 void *
 heim_retain(void *ptr)
 {
-    struct heim_base *p = PTR2BASE(ptr);
+    struct heim_base *p = NULL;
 
     if (ptr == NULL || heim_base_is_tagged(ptr))
 	return ptr;
+
+    p = PTR2BASE(ptr);
 
     if (heim_base_atomic_load(&p->ref_cnt) == heim_base_atomic_integer_max)
 	return ptr;
@@ -109,10 +111,12 @@ void
 heim_release(void *ptr)
 {
     heim_base_atomic_integer_type old;
-    struct heim_base *p = PTR2BASE(ptr);
+    struct heim_base *p = NULL;
 
     if (ptr == NULL || heim_base_is_tagged(ptr))
 	return;
+
+    p = PTR2BASE(ptr);
 
     if (heim_base_atomic_load(&p->ref_cnt) == heim_base_atomic_integer_max)
 	return;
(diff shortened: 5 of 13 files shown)


kdc/krb5tgs.c
```diff
@@ -787,7 +787,9 @@ tgs_make_reply(astgs_request_t r,
 	free(et.authorization_data);
     }
     free_LastReq(&ek.last_req);
-    memset(et.key.keyvalue.data, 0, et.key.keyvalue.length);
+    if (et.key.keyvalue.data != NULL) {
+	memset(et.key.keyvalue.data, 0, et.key.keyvalue.length);
```
jaltman, line 791: The protection against use of NULL provided that the length is not known to be zero.   
Unrelated, this call should be to memset_s() not memset() which is likely to be optimized away because 'et' is a stack variable going out of scope.

lib/asn1/der_copy.c
```diff
@@ -103,7 +103,9 @@ der_copy_printable_string (const heim_printable_string *from,
     to->data   = malloc(to->length + 1);
     if(to->data == NULL)
 	return ENOMEM;
-    memcpy(to->data, from->data, to->length);
+    if (to->length != 0) {
```
jaltman, line 106: unnecessary.   a length of zero will copy nothing

lib/asn1/der_copy.c
```diff
@@ -122,7 +124,9 @@ der_copy_bmp_string (const heim_bmp_string *from, heim_bmp_string *to)
     to->data   = malloc(to->length * sizeof(to->data[0]));
     if(to->length != 0 && to->data == NULL)
 	return ENOMEM;
-    memcpy(to->data, from->data, to->length * sizeof(to->data[0]));
+    if (to->length * sizeof(to->data[0]) != 0) {
```
jaltman, line 127: unnecessary.  a length of zero will copy nothing.

lib/asn1/der_copy.c
```diff
@@ -134,7 +138,9 @@ der_copy_universal_string (const heim_universal_string *from,
...
```
jaltman, line 141: unnecessary.  a length of zero will copy nothing.

lib/asn1/der_copy.c
```diff
@@ -152,7 +158,9 @@ der_copy_octet_string (const heim_octet_string *from, heim_octet_string *to)
     to->data   = malloc(to->length);
     if(to->length != 0 && to->data == NULL)
 	return ENOMEM;
-    memcpy(to->data, from->data, to->length);
+    if (to->length != 0) {
```
jaltman, line 161: a length of zero will copy nothing.

lib/asn1/der_copy.c
```diff
@@ -163,7 +171,9 @@ der_copy_heim_integer (const heim_integer *from, heim_integer *to)
...
```
jaltman, line 174: a length of zero will copy nothing

lib/asn1/der_copy.c
```diff
@@ -175,8 +185,10 @@ der_copy_oid (const heim_oid *from, heim_oid *to)
     to->components = malloc(to->length * sizeof(*to->components));
     if (to->length != 0 && to->components == NULL)
 	return ENOMEM;
-    memcpy(to->components, from->components,
-	   to->length * sizeof(*to->components));
+    if (to->length * sizeof(*to->components) != 0) {
```
jaltman, line 188: a length of zero will copy nothing.

lib/asn1/der_copy.c
```diff
@@ -190,6 +202,8 @@ der_copy_bit_string (const heim_bit_string *from, heim_bit_string *to)
     to->data   = malloc(len);
     if(len != 0 && to->data == NULL)
 	return ENOMEM;
-    memcpy(to->data, from->data, len);
+    if (len != 0) {
```
jaltman, line 205: a length of zero will copy nothing.

lib/asn1/der_get.c
```diff
@@ -326,7 +326,9 @@ der_get_octet_string (const unsigned char *p, size_t len,
     data->data = malloc(len);
     if (data->data == NULL && data->length != 0)
 	return ENOMEM;
-    memcpy (data->data, p, len);
+    if (len != 0) {
```
jaltman, line 329: a length of zero will copy nothing

lib/asn1/der_put.c
```diff
@@ -237,7 +237,9 @@ der_put_general_string (unsigned char *p, size_t len,
     if (len < slen)
 	return ASN1_OVERFLOW;
     p -= slen;
-    memcpy (p+1, *str, slen);
+    if (slen != 0) {
```
jaltman, line 240: a length of zero will copy nothing.

lib/asn1/der_put.c
```diff
@@ -313,7 +315,9 @@ der_put_octet_string (unsigned char *p, size_t len,
     if (len < data->length)
 	return ASN1_OVERFLOW;
     p -= data->length;
-    memcpy (p+1, data->data, data->length);
+    if (data->length != 0) {
```
jaltman, line 318: a length of zero will copy nothing.

lib/asn1/der_put.c
```diff
@@ -479,7 +483,9 @@ der_replace_tag(const unsigned char *p, size_t len,
      */
     if ((*out = malloc(*outlen = tag_len + len_len + payload_len)) == NULL)
         return ENOMEM;
-    memcpy(*out + tag_len, p, len_len + payload_len);
+    if (len_len + payload_len != 0) {
```
jaltman, line 486: a length of zero will copy nothing.

lib/asn1/der_put.c
```diff
@@ -532,7 +538,9 @@ der_encode_implicit(unsigned char *p, size_t len,
     if (e == 0) {
         (*size) -= ttaglen;
         (*size) += itaglen;
-        memcpy(p - *size, p2 + ttaglen - itaglen, *size);
+        if (*size != 0) {
```
jaltman, line 541: a length of zero will copy nothing.

lib/asn1/der_put.c
```diff
@@ -633,7 +641,9 @@ der_put_bit_string (unsigned char *p, size_t len,
 	return ASN1_OVERFLOW;
     p -= data_size + 1;
 
-    memcpy (p+2, data->data, data_size);
+    if (data_size != 0) {
```
jaltman, line 644: a length of zero will copy nothing.

lib/krb5/store_emem.c
```diff
@@ -47,7 +47,9 @@ emem_fetch(krb5_storage *sp, void *data, size_t size)
     emem_storage *s = (emem_storage*)sp->data;
     if((size_t)(s->base + s->len - s->ptr) < size)
 	size = s->base + s->len - s->ptr;
-    memmove(data, s->ptr, size);
+    if (size != 0) {
```
jaltman, line 50: a length of zero will copy nothing.

lib/krb5/store_emem.c
```diff
@@ -70,7 +72,9 @@ emem_store(krb5_storage *sp, const void *data, size_t size)
 	s->base = base;
 	s->ptr = (unsigned char*)base + off;
     }
-    memmove(s->ptr, data, size);
+    if (size != 0) {
```
jaltman, line 75: a length of zero will copy nothing.

jaltman (changes requested):
Calling memcpy() and memmove() or memset() with a known length of zero is safe.  The operations are no-ops.    The changes here are imposing the cost of a conditional in all cases in order to avoid the overhead of a function all in the unexpected case of length being zero.   

The checks to prevent passing a NULL pointer when the length is unknown are fine.

abartlet:
No worries, we won't push these further given your NACK.  My understanding is that these came up from real-world tools, but I tend to share your views and hope on the line between 'technically undefined' and 'what (I hope) a compiler may dare do'.

@josephsutton1 may wish to comment further.  I'll leave this open for the moment so we don't forget this decision while we work though our list of patches.

jaltman:
@abartlet  There are cases where NULL can be passed as input without length zero.  Those should still be fixed.  Please revise the PR and if you would make use of memset_s() where it is required.

Thanks

josephsutton1:
> Calling memcpy() and memmove() or memset() with a known length of zero is safe. The operations are no-ops.

Unfortunately, that's not necessarily the case, according to the C standard. The definitions of [memcpy()](https://en.cppreference.com/w/c/string/byte/memcpy), [memmove()](https://en.cppreference.com/w/c/string/byte/memmove), and [memset()](https://en.cppreference.com/w/c/string/byte/memset) state that the behaviour is undefined if any of the pointer arguments are NULL, and a compiler could make use of this to optimise away existing NULL checks in the source.

jaltman:
@josephsutton1 its fine if the compiler optimizes away NULL checks when the input length is known to be zero.   If the compiler is generating code that copies more than zero bytes when the input is zero, its a bug.

josephsutton1:
That would only be a bug if the pointers passed in were non-NULL. Since undefined behaviour is occurring, the compiler is technically allowed to do anything it likes.

jaltman:
I will approve a change that tests for NULL; I will not approve a change that tests for length equal to zero.

abartlet:
To be fair, an undefined behaviour of skipping the memcpy() when an argument in NULL Is probably safe-ish, and the ovbious thing to do, so not worth guarding against.  Hopefully the NULL cases have a count of 0, and it is all moot.  Ideally you would assert that to be the case.

But once you enter into undefined territory, what is to stop the over-eager compiler just asserting that those pointers can't possibly be NULL any longer, to more drastic ends?

Anyway, I'll add @josephsutton1 to my repo, and you @jaltman have access via the 'allow edits from maintainers'.

josephsutton1:
OK, I've pushed up some changes.

> But once you enter into undefined territory, what is to stop the over-eager compiler just asserting that those pointers can't possibly be NULL any longer, to more drastic ends?

Yes, it seems like GCC at least does this in practice: [https://www.imperialviolet.org/2016/06/26/nonnull.html](https://www.imperialviolet.org/2016/06/26/nonnull.html)

abartlet:
> 
> Yes, it seems like GCC at least does this in practice: https://www.imperialviolet.org/2016/06/26/nonnull.html

Yikes...

kaduk:
I will also add my weight here for the need to check for the zero-length case.  The standard is clear, and we have lots of evidence that compilers will learn to take advantage of it.

jaltman:
> https://www.imperialviolet.org/2016/06/26/nonnull.html

I am in favor of 
- Asserting that the pointers are or are not NULL
- Using helper functions that check that both the source and destination are not NULL before calling any of the mem-funds
- Providing an allocation function that returns a non-NULL value for size zero.

I am not in favor of adding conditional checks all over the source tree.  They reduce the ease of reading the code and are fragile.

lhoward:
I agree with @jaltman :)

jaltman:
I think there is an underlying problem with the der functions as it is impossible to distinguish between a freed object and an allocated object whose length is zero.   Calling der_copy_xxxx() on a freed object should cause the library to assert.    Whereas calling der_copy_xxxx() on an allocated object whose length is zero can be handled as a special case or be safely passed to memset() and memcpy().

The same is true for all of the data structures which include some form of 'length' and a pointer whether named 'data', 'contents', etc.

nicowilliams:
The DER codec functions don't allocate or release the top-level structure into which they are decoding or which they are releasing.

The DER codec functions have a serious flaw in their signatures: the encoder/decoder should take an optional tag meant to replace the to-be-encoded type's tag.  The hoops I had to jump through to deal with this for proper implementation of `IMPLICIT` tagging...

jaltman:
ef059b8831f4e12304246bb511b9b9e077b98c4e and 1c93a6ff2613f827a1414bb88a7388354e3d0702

lib/asn1/der_copy.c
```diff
@@ -103,7 +103,9 @@ der_copy_printable_string (const heim_printable_string *from,
     to->data   = malloc(to->length + 1);
     if(to->data == NULL)
 	return ENOMEM;
-    memcpy(to->data, from->data, to->length);
+    if (from->data != NULL) {
```
nicowilliams, line 106: This should be on the callers, not here.

lib/asn1/der_copy.c
```diff
@@ -122,7 +124,9 @@ der_copy_bmp_string (const heim_bmp_string *from, heim_bmp_string *to)
     to->data   = malloc(to->length * sizeof(to->data[0]));
     if(to->length != 0 && to->data == NULL)
 	return ENOMEM;
-    memcpy(to->data, from->data, to->length * sizeof(to->data[0]));
+    if (to->data != NULL && from->data != NULL) {
```
nicowilliams, line 127: It's on the caller to not pass in a NULL `from`.

lib/asn1/der_copy.c
```diff
@@ -152,7 +158,9 @@ der_copy_octet_string (const heim_octet_string *from, heim_octet_string *to)
     to->data   = malloc(to->length);
     if(to->length != 0 && to->data == NULL)
 	return ENOMEM;
-    memcpy(to->data, from->data, to->length);
+    if (to->data != NULL && from->data != NULL) {
```
nicowilliams, line 161: Ditto.

lib/asn1/der_copy.c
```diff
@@ -163,7 +171,9 @@ der_copy_heim_integer (const heim_integer *from, heim_integer *to)
...
```
nicowilliams, line 174: Ditto.

lib/asn1/der_copy.c
nicowilliams: Also, `to->data` has already been checked, thus this adds dead code for coverity to complain about.

lib/asn1/der_copy.c
nicowilliams: Either the callers should check or we should have an `assert()` here about both, `from` and `to` (`assert(from && to);`).  I would prefer the assertion.

lib/base/heimbase.c
```diff
@@ -86,11 +86,13 @@ struct heim_auto_release {
 void *
 heim_retain(void *ptr)
 {
-    struct heim_base *p = PTR2BASE(ptr);
+    struct heim_base *p = NULL;
```
nicowilliams, line 89: A compiler that blows up because of undefined pointer arithmetic (which, naturally, doesn't involve division by zero) where we never deref that pointer or act on the arithmetic is a pretty evil compiler...

lib/krb5/pac.c
```diff
@@ -360,8 +360,10 @@ krb5_pac_add_buffer(krb5_context context, krb5_pac p,
      * copy in new data part
      */
 
-    memcpy((unsigned char *)p->data.data + offset,
-	   data->data, data->length);
+    if (data->data != NULL) {
+	memcpy((unsigned char *)p->data.data + offset,
```
nicowilliams, line 364: `data->data` already has been checked for NULL.

lib/krb5/store_emem.c
```diff
@@ -47,7 +47,9 @@ emem_fetch(krb5_storage *sp, void *data, size_t size)
     emem_storage *s = (emem_storage*)sp->data;
     if((size_t)(s->base + s->len - s->ptr) < size)
 	size = s->base + s->len - s->ptr;
-    memmove(data, s->ptr, size);
+    if (data != NULL) {
```
nicowilliams, line 50: I'd rather have an assertion.

jaltman:
@josephsutton1 https://github.com/heimdal/heimdal/pull/944 is an alternative to the first commit in this PR.  

All of the other commits have already been addressed on the master branch.

jaltman:
all of the issues raised by this pull request have been addressed in other ways.

PR #873, opened 2021-12-07
kdc: Match Windows when advertising FAST

josephsutton1:
Windows doesn't unconditionally announce PADATA-FX-FAST or PADATA-ENCRYPTED-CHALLENGE, and it doesn't allow PADATA-ENC-TIMESTAMP to be used for preauthentication with FAST.

### DIFF
--- a/kdc/kerberos5.c
+++ b/kdc/kerberos5.c
@@ -2147,6 +2147,8 @@ _kdc_as_rep(astgs_request_t r)
 		continue;
 	    if (r->armor_crypto == NULL && (pat[n].flags & PA_REQ_FAST))
 		continue;
+	    if (r->armor_crypto != NULL && (pat[n].flags & PA_NOT_IN_FAST))
+		continue;
 
 	    kdc_log(context, config, 5,
 		    "Looking for %s pa-data -- %s", pat[n].name, r->cname);
--- a/kdc/kerberos5.c
+++ b/kdc/kerberos5.c
@@ -929,6 +929,7 @@ struct kdc_patypes {
 #define PA_ANNOUNCE	1
 #define PA_REQ_FAST	2 /* only use inside fast */
 #define PA_SYNTHETIC_OK	4
+#define PA_NOT_IN_FAST	8 /* only use outside fast */
     krb5_error_code (*validate)(astgs_request_t, const PA_DATA *pa);
 };
 
@@ -955,7 +956,7 @@ static const struct kdc_patypes pat[] = {
     { KRB5_PADATA_PA_PK_OCSP_RESPONSE , "OCSP", 0, NULL },
     { 
 	KRB5_PADATA_ENC_TIMESTAMP , "ENC-TS",
-	PA_ANNOUNCE,
+	PA_ANNOUNCE | PA_NOT_IN_FAST,
 	pa_enc_ts_validate
     },
     {
@@ -2203,6 +2204,14 @@ _kdc_as_rep(astgs_request_t r)
 	for (n = 0; n < sizeof(pat) / sizeof(pat[0]); n++) {
 	    if ((pat[n].flags & PA_ANNOUNCE) == 0)
 		continue;
+
+	    if (r->armor_crypto == NULL) {
+		if (pat[n].flags & PA_REQ_FAST)
+		    continue;
+	    } else if (pat[n].flags & PA_NOT_IN_FAST)
+		continue;
+	    }
+
 	    ret = krb5_padata_add(context, &error_method,
 				  pat[n].type, NULL, 0);
 	    if (ret)


lhoward:
@nicowilliams @jaltman what do you think?

josephsutton1:
Re-pushed with a cleaner approach.

lhoward:
Very nice!

lhoward:
We can accept the PR but, can you please add `PA_ANNOUNCE` back for `PADATA-FX-FAST`? RFC6113 says ”As with all pre-authentication types, the KDC SHOULD advertise PA-FX-FAST in a PREAUTH_REQUIRED error”. We would prefer to follow the RFC unless you've identified a specific Windows interoperability issue.

Cc @jaltman

josephsutton1:
OK, I've pushed without the PA_ANNOUNCE change for now.

kdc/kerberos5.c
```diff
@@ -2146,6 +2147,8 @@ _kdc_as_rep(astgs_request_t r)
 		continue;
 	    if (r->armor_crypto == NULL && (pat[n].flags & PA_REQ_FAST))
 		continue;
+	    if (r->armor_crypto != NULL && !(pat[n].flags & PA_REQ_FAST))
+		continue;
```
lhoward, line 2151: Won't this change break mechanisms like GSS that can work both with and without FAST?

lhoward: Did you mean to check for `PA_NOT_IN_FAST` instead?

kdc/kerberos5.c
```diff
@@ -931,6 +931,7 @@ struct kdc_patypes {
 #define PA_SYNTHETIC_OK	4
     krb5_error_code (*validate)(astgs_request_t, const PA_DATA *pa);
 };
+#define PA_NOT_IN_FAST	8 /* don't announce in fast */
```
lhoward, line 934: Also could you please move this to be after `PA_SYNTHETIC_OK`.

josephsutton1: Done.

kdc/kerberos5.c
josephsutton1: Right, I neglected the GSS padata type. Is it better with the changes I've now pushed?

lhoward: Is there a reason we don't want to allow PKINIT inside FAST? cc @nicowilliams

kdc/kerberos5.c
```diff
@@ -2203,6 +2208,15 @@ _kdc_as_rep(astgs_request_t r)
 	for (n = 0; n < sizeof(pat) / sizeof(pat[0]); n++) {
 	    if ((pat[n].flags & PA_ANNOUNCE) == 0)
 		continue;
+
+	    if (r->armor_crypto == NULL) {
+		if (pat[n].flags & PA_REQ_FAST)
+		    continue;
+	    } else {
+		if (pat[n].flags & PA_DONT_ANNOUNCE_FAST)
```
nicowilliams, line 2216: Make this `else if` pls.

kdc/kerberos5.c
```diff
@@ -929,18 +929,21 @@ struct kdc_patypes {
 #define PA_ANNOUNCE	1
 #define PA_REQ_FAST	2 /* only use inside fast */
 #define PA_SYNTHETIC_OK	4
+#define PA_NOT_IN_FAST	8 /* only use outside fast */
```
nicowilliams, line 932: Do we really need `PA_NOT_IN_FAST`?

I'm thinking I want `PA_REQ_FAST_PREF` so we can have a migration strategy for realms where FAST is not deployed.

kdc/kerberos5.c
nicowilliams: I don't see any reason to disallow PKINIT in FAST.

kdc/kerberos5.c
```diff
@@ -955,7 +958,7 @@ static const struct kdc_patypes pat[] = {
     { KRB5_PADATA_PA_PK_OCSP_RESPONSE , "OCSP", 0, NULL },
     { 
 	KRB5_PADATA_ENC_TIMESTAMP , "ENC-TS",
-	PA_ANNOUNCE,
+	PA_ANNOUNCE | PA_NOT_IN_FAST | PA_DONT_ANNOUNCE_FAST,
```
nicowilliams, line 961: This is the opposite of what we should want.  We should want `KRB5_PADATA_ENC_TIMESTAMP` to only be used inside a FAST armored tunnel.

But more importantly, it's pointless.  What matters as far as not sending encrypted timestamps not super-encrypted is the _client_.  If a `kinit` can be spoofed into sending an encrypted timestamp outside a FAST armored tunnel, then the password is compromised (because it can be attacked offline at that point).

nicowilliams: I'm inclined to accept (I'm even coding, as we speak) a patch to add two configuration parameters:

- `[kdc] enable_enc_timestamp = BOOL`
- `[kdc] require_armored_weak_preauth = BOOL`

and if `require_armored_weak_preauth = true` then allow enc-timestamp only inside armored FAST and only advertise it inside armored FAST.

nicowilliams: I'm also thinking of a HDB principal entry flag for "no enc timestamp".

nicowilliams: We should patch our kinit to always try `pk_anon_fast_armor` unless an armor ccache is given.

abartlet:
Please make sure those end up as booleans in the kdc configuration
block, so Samba can just set them.

Thanks!

kdc/kerberos5.c
josephsutton1: KRB5_PADATA_ENC_TIMESTAMP only uses the client's key, not the armor key, for validation, so don't we want to disallow it with FAST? Shouldn't it rather be KRB5_PADATA_ENCRYPTED_CHALLENGE that is only used inside a FAST tunnel?

nicowilliams: Oh, no, I'm confused.  `ENC-CHAL` replaces `ENC-TIMESTAMP` inside FAST.

lhoward:
I pushed a patch we can accept to the tip of #805. But this breaks check-fast on master, because (on a superficial analysis) it attempts to use ENC-TS inside FAST and when that fails removes it from the list of acceptable mechanisms. This presumably got fixed by the Apple code.

(Did you run make check with this PR?)

What does this mean? Practically, if we’ve shipped the FAST client code in master, then I guess we can’t accept this patch unless it’s behind a runtime flag.

lhoward:
Gah, looks like I haven’t pushed the change to #805 yet. I will tonight. But, yes, I think we need to make sure existing Heimdal clients aren’t broken. And that FAST AS code has been there for a while.

lhoward:
But yes, my patch is effectively what you that pushed, with a small stylistic change. Is tests/kdc/check-fast working on master for you with this change? If so, perhaps I missed something. If not, then see above.

josephsutton1:
I haven't tried running tests/kdc/check-fast yet, but I will do so and try to sort out any failures.

lhoward:
Great. The only thing is, if we have shipped a broken client then we probably can’t change the KDC. This probably means master might fail with FAST against Windows too, have you tried that?

lhoward:
The change to ignore `PA_NOT_IN_FAST` pre-authentication types when FAST is used is breaking the AS client on master, because there's no support for `PA-ENCRYPTED-CHALLENGE`. (It works with #805.) As far as I can tell, FAST support did ship in Heimdal 7.x.

After discussing with the team, we're going to close this PR and instead use #875 as a means of disabling `PA-ENC-TIMESTAMP`.

abartlet:
That makes sense, matching Windows while breaking ones own clients makes little sense, if `PA-ENC-TIMESTAMP` is no weaker in FAST than outside and there is no change in what ticket it becomes because of the FAST armour.  Samba may have to do differently but we will work that out ourselves, probably via an extension to #875

PR #874, opened 2021-12-07
kdc: Check for unsupported critical FAST options

josephsutton1:
They were not previously being checked for.


lhoward:
I don't know if we want the `reserved` patch, the RFC doesn't say we should reject can you resubmit with just the new error code? I don't think we need the critical flags in the ASN.1 either.

josephsutton1:
OK, I removed the check for the 'reserved' flag. Is there a way to detect unsupported options without adding them to the ASN.1? If I don't add them, decode_KrbFastReq() will only set the bits of fastreq.fast_options that are listed in the ASN.1, and then the check won't do anything.

lhoward:
Ah, didn't know that. Well, I'm happy to take the patch then unless @nicowilliams has a better idea? He's the ASN.1 compiler expert.

PR #875, opened 2021-12-09
Improve weak preauth config options

nicowilliams:


abartlet:
If this is enforced by default (as it appears to be) wouldn't this break many existing clients?  How many clients currently do anonymous PKINIT strengthened FAST by default?

lhoward:
> If this is enforced by default (as it appears to be) wouldn't this break many existing clients? How many clients currently do anonymous PKINIT strengthened FAST by default?

Both are disabled by default, no?

abartlet:
` c->enable_weak_preauth = FALSE;`

lhoward:
Right you are :)

abartlet:
Would a tri-state make sense here, being **disabled**, **disabled-in-fast** and **enabled**?

Samba could then set **disabled-in-fast** to avoid a situation where we try to combine the PAC from the armor into a claims PAC if there is no strong linkage.

lhoward:
> Would a tri-state make sense here, being **disabled**, **disabled-in-fast** and **enabled**?
> 
> Samba could then set **disabled-in-fast** to avoid a situation where we try to combine the PAC from the armor into a claims PAC if there is no strong linkage.

As I read the code, the KDC strengthens the reply key with the armour key even when `PA-ENC-TIMESTAMP` is used. The only difference is that for `PA-ENCRYPTED-CHALLENGE`, the client library (#805 branch) **requires** a strengthen key, as mandated by RFC 6113.

So the question is, what additional cryptographic value does including the armour key in the challenge key derivation provide? RFC 6113 says: ”Because the armor key is fresh and random, the challenge key is fresh and random”, which makes sense. But the whole thing is protected by `KrbFastArmoredReq` anyway, so I'm not sure if allowing `PA-ENC-TIMESTAMP` inside FAST is a vulnerability. But, I am not a cryptographer.

Anyway, tri-state as you suggest sounds like a good idea to me. @nicowilliams?

abartlet:
Thanks so much for looking that over, and providing that reassurance, I really should have done that research.  Thanks for considering a tri-state, that sounds like a good way for us to have a mode that passes Samba's current 'strictly windows behaviour' tests.

nicowilliams:
Indeed, it breaks our own tests.  I was being optimistic :) and trying to see if this got reviewed!  OK, the last part is a lie :), but ex-post it was my intent ;)

I'll enable weak preauth.

nicowilliams:
> Would a tri-state make sense here, being disabled, disabled-in-fast and enabled?

@abartlet @lhoward I think "no" if we can just always strengthen ENC-TIMESTAMP in FAST, then there's no harm from allowing it in FAST, so we might as well have just a boolean.

lhoward:
Only other question is whether `enable_weak_preauth` is perhaps a bit too broad a name for the configuration key. I know it's a balance but perhaps `enable_pa_timestamp` might be better? (or `disable_pa_timestamp`)

abartlet:
Samba would like the option to match AD byte-for-byte if possible, even
if there is no specific protocol need.  

We have a very strict set of tests now and it helps us chase down and
isolate issues if we can just match exactly.

lhoward:
Could you just not test this? ;) Anyway, I have no objection to the tri-state but I'm going to leave it to @nicowilliams.

nicowilliams:
> Samba would like the option to match AD byte-for-byte if possible, even
> if there is no specific protocol need.
> 
> We have a very strict set of tests now and it helps us chase down and
> isolate issues if we can just match exactly.

I'll make it two boolean parameters then.  I don't really like tri-states.

lhoward:
LGTM.

abartlet:
This breaks Samba where a kinit with an expired password now fails against our DC with: 
```
Client (WELLKNOWN/ANONYMOUS@SAMBA2008R2.EXAMPLE.COM) unknown
```
https://gitlab.com/samba-team/devel/samba/-/jobs/1881717562#L1828

I can confirm that a revert of 313caed18a2cb2e88daa99c8e96a6e8a3e04bccb fixes `make test TESTS=samba4.blackbox.kinit`

nicowilliams:
> This breaks Samba where a kinit with an expired password now fails against our DC with:
> 
> ```
> Client (WELLKNOWN/ANONYMOUS@SAMBA2008R2.EXAMPLE.COM) unknown
> ```
> 
> https://gitlab.com/samba-team/devel/samba/-/jobs/1881717562#L1828
> 
> I can confirm that a revert of [313caed](https://github.com/heimdal/heimdal/commit/313caed18a2cb2e88daa99c8e96a6e8a3e04bccb) fixes `make test TESTS=samba4.blackbox.kinit`

Thanks for the report.  I'll fix it.

nicowilliams:
Wow, this took me down a rabbit hole.

First of all, when `kpasswdd` rejects a new password then we get `get_new_tickets()` returning `ENOTTY` because we don't actually have a `KRB5` error code for that.  Ugh.

Second, we have a `KRB5_FAST_OPTIMISTIC` flag for `FAST` state in `init_creds_pw.c`, but we don't ever set it.  In fact, it's exactly what I want, and then I can go back to not having a fallback in `kinit` (though the remaining changes to it would remain).  So I need a setter for this option.

I'll continue tomorrow.

abartlet:
Thanks for digging into this.  This isn't blocking us, I've just put a revert into our patch set for now.

nicowilliams:
I've reverted that commit and am reworking it.

PR #877, opened 2021-12-14
asn1: Fix binary search off-by-one read

josephsutton1:
Previously, if left==right==A1_HEADER_LEN(tos), this would read past the end of the template array. Now we treat [left, right) as a half-open interval and no longer try to read from 'right'.


nicowilliams:
Thanks!

PR #878, opened 2021-12-14
kdc: update PAC hooks for Samba

lhoward:
Samba includes the user's long-term credentials (encrypted in the AS reply key) to allow legacy authentication protocols such as NTLM to work even if the pre-authentication mechanism replaced the reply key (as PKINIT does).

Samba also needs to know whether the client explicitly requested a PAC be included (or excluded), in order to defer PAC exclusion until a service ticket is issued (thereby avoiding a name binding attack if the user is renamed between TGT and service ticket issuance).

References:

https://bugzilla.samba.org/show_bug.cgi?id=11441
https://bugzilla.samba.org/show_bug.cgi?id=14561

Closes: #864

Original authors:
 - Joseph Sutton <josephsutton@catalyst.net.nz>
 - Andrew Bartlett <abartlet@samba.org>
 - Stefan Metzmacher <metze@samba.org>

### DIFF
--- a/kdc/kdc_locl.h
+++ b/kdc/kdc_locl.h
@@ -99,6 +99,7 @@ struct astgs_request_desc {
     /* only valid for tgs-req */
     unsigned int rk_is_subkey : 1;
     unsigned int fast_asserted : 1;
+    unsigned int replaced_reply_key : 1;
 
     krb5_crypto armor_crypto;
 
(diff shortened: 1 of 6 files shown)


abartlet:
Looking at this now, will run the Samba tests and let you know, but
happy with the concept.

The boolean will be useful for auditing also.

lhoward:
Unless I've missed something – which is entirely possible – it's exactly your patch with the requested changes :)

abartlet:
This passes `make test TESTS=pkinit` in Samba.  Full CI going on here: https://gitlab.com/samba-team/devel/samba/-/pipelines/428830347

abartlet (approved):

PR #879, opened 2021-12-14
kdc: Add support for explicit armoring from MS-KILE

josephsutton1:
This is required for compound identity.

### DIFF
--- a/kdc/fast.c
+++ b/kdc/fast.c
@@ -442,6 +442,8 @@ fast_unwrap_request(astgs_request_t r,
     krb5_flags ap_req_options;
     Key *armor_key = NULL;
     krb5_keyblock armorkey;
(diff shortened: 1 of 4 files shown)


lhoward:
LGTM, but given this is not described in RFC6113 and is very poorly described in [MS-KILE] 3.3.5.7.4, could you update the commit message with a bit of an explanation of what's going on?

e.g. when FAST is used with a TGS-REQ, the armour key is usually implicitly derived from the TGT rather than being explicitly included as it is for AS-REQs. Windows allows an armour key to be explicitly provided for TGS-REQs as well, because... (more research as to why may be required)

lhoward:
Ah, I guess it is to allow the device claims to be propagated from an armour TGT.

josephsutton1:
OK, I added some more information to the commit message.

PR #884, opened 2021-12-14
kdc: Don't keep trying keys for encrypted-challenge

josephsutton1:
Don't keep trying keys for encrypted-challenge if one decrypts but fails to verify. This also enables a KRB5KRB_AP_ERR_SKEW error to be returned to the client.


lhoward:
Is this the same behavior as ENC-TS? Haven’t checked yet but if they’re aligned that makes sense.

josephsutton1:
Yes, it seems like for the ENC-TS case we only try again with the next key if we failed to decrypt the EncryptedData, but not if we failed to decode the timestamp or the skew check failed.

PR #886, opened 2021-12-14
hdb: Do not return HDB_ERR_WRONG_REALM if h->entry.force_canonicalise…

josephsutton1:
… is set.

h->entry.force_canonicalise is set by Samba when it has control over the entry totally and wants to be very clear as to what is returned.

Samba may return HDB_ERR_WRONG_REALM in the future if it also gains alias support, eg for cross-realm UPN lookups.

### DIFF
--- a/lib/hdb/common.c
+++ b/lib/hdb/common.c
@@ -1478,7 +1478,22 @@ hdb_fetch_kvno(krb5_context context,
     ret = fetch_it(context, db, principal, flags, t, etype, kvno, h);
     if (ret == HDB_ERR_NOENTRY)
 	krb5_set_error_message(context, ret, "no such entry found in hdb");
+
+    /*
+     * This supports aliases in HDB.
+     *
+     * The check for force_canonicalize is to support Samba.
+     *
+     * Samba will return a different realm (the canonical realm) for a
+     * lookup if the client presents a NetBIOS name.  In this case so
+     * krb5_realm_compare() would return false even for full entries
+     * in our own realm.
+     *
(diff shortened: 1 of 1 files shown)


lhoward:
Happy to merge but having trouble following the comment, could you perhaps split into a couple of sentences? :)

```c
     * This supports aliases in HDB.
      *
      * The check for force_canonicalize is to support Samba which will
      * return a different realm in the case of a lookup with a NetBIOS
      * name and will return HDB_ERR_WRONG_REALM internally when
      * aliasing using Active Directory rules is implemented
```

abartlet:
> Happy to merge but having trouble following the comment, could you perhaps split into a couple of sentences? :)

Sure, will do. :-)

lhoward:
Commit applied separately. When porting your patches over to Heimdal, it would be good if comments and commit messages described them in as abstract (i.e. non-Samba-specific) a way as possible. Here's the commit message I used:

>    hdb: do not return HDB_ERR_WRONG_REALM if force_canon set  
>    In hdb_fetch_kvno(), do not return HDB_ERR_WRONG_REALM if the backend set the force_canonicalize flag

and the comment:

```c
    /*
     * This check is to support aliases in HDB; the force_canonicalize
     * check is to allow HDB backends to support realm name canon
     * independently of principal aliases (used by Samba).
     */
    if (ret == 0 && !(flags & HDB_F_ADMIN_DATA) &&
        !h->entry.flags.force_canonicalize &&
        !krb5_realm_compare(context, principal, h->entry.principal))
            ret = HDB_ERR_WRONG_REALM;
```

PR #887, opened 2021-12-14
heimdal:kdc: Only check for default salt for des-cbc-crc enctype

josephsutton1:
This function is used to prefer keys that have a default salt over keys that do not, to preserve compatibility with older clients that do not check the returned salt. With Active Directory salt generation, this can lead to RC4 keys (which do not use a salt) being preferred over AES keys (which use a non-default salt for machine accounts). To avoid this behaviour, we can apply this check only for the des-cbc-crc enctype.

### DIFF
--- a/kdc/kerberos5.c
+++ b/kdc/kerberos5.c
@@ -119,6 +119,23 @@ is_default_salt_p(const krb5_salt *default_salt, const Key *key)
     return TRUE;
 }
 
+/*
+ * Detect if `key' is the using the the precomputed `default_salt'
+ * (for des-cbc-crc) or any salt otherwise.
+ *
+ * This is for avoiding Kerberos v4 (yes really) keys in AS-REQ as
+ * that salt is strange, and a buggy client will try to use the
+ * principal as the salt and not the returned value.
+ */
+
+static krb5_boolean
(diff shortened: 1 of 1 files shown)


lhoward:
I don't know enough about the code and behaviour of older clients to know whether to accept this or not. At least by common definitions of "older" clients, probably all pre-RFC3961 enctypes should be checked.

Cc @nicowilliams @jaltman who will know more

abartlet:
This is the wrong patch.  I wrote a different one built around the concept of a _good_ salt, as otherwise (while not in Samba as we always return the salt) the wrong ` PA-ETYPE-INFO*` values.

lhoward:
OK, please close if it is addressed elsewhere.

abartlet:
I've found my better patch and I'll upload it here soon, just want to ensure it passes tests, not sure why the wrong one was still on our tree.

lhoward:
I can't quite get my head around this right now, hopefully @nicowilliams or @jaltman can help.

jaltman:
This change was refactored and merged as 9a0372d99

jaltman:
Cherry-picked to heimdal-7-1-branch as fb6345459

PR #888, opened 2021-12-14
Pass extra info to hdb_auth_status()

abartlet:
From a commit by @metze-samba and myself in Samba's lorikeet-heimdal repo.  

We need more information in our auditing hooks for Samba to fill in it's internal audit logs with.   A common helper function was added to make this maintainable.

@metze-samba I dropped to TODO on this as I think is well baked now.

### DIFF
--- a/kdc/kerberos5.c
+++ b/kdc/kerberos5.c
@@ -753,10 +753,12 @@ pa_enc_chal_validate(astgs_request_t r,
 	goto out;
     }
 
-    ret = KRB5KDC_ERR_PREAUTH_FAILED;
-
-    if (invalidPassword)
+    if (invalidPassword) {
 	auth_status->auth_status = HDB_AUTH_WRONG_PASSWORD;
+	ret = KRB5KDC_ERR_PREAUTH_FAILED;
+    } else {
+	ret = KRB5KDC_ERR_ETYPE_NOSUPP;
+    }
  out:
     free_EncryptedData(&enc_data);
 
--- a/kdc/kerberos5.c
+++ b/kdc/kerberos5.c
@@ -507,7 +507,7 @@ pa_pkinit_validate(astgs_request_t r,
 	ret = KRB5KRB_AP_ERR_BAD_INTEGRITY;
 	_kdc_r_log(r, 4, "Failed to decode PKINIT PA-DATA -- %s",
 		   r->cname);
-	auth_status->auth_status = HDB_AUTH_PKINIT_FAILURE;
+	auth_status->auth_status = HDB_AUTHSTATUS_PKINIT_FAILURE;
 	goto out;
     }
 
@@ -515,7 +515,7 @@ pa_pkinit_validate(astgs_request_t r,
     if (ret) {
 	_kdc_set_e_text(r, "PKINIT certificate not allowed to "
 			"impersonate principal");
-	auth_status->auth_status = HDB_AUTH_PKINIT_FAILURE;
+	auth_status->auth_status = HDB_AUTHSTATUS_PKINIT_FAILURE;
 	goto out;
     }
     auth_status->auth_details = client_cert;
@@ -536,7 +536,7 @@ pa_pkinit_validate(astgs_request_t r,
     ret = _kdc_add_initial_verified_cas(r->context, r->config,
 					pkp, &r->et);
 
-    auth_status->auth_status = HDB_AUTH_PKINIT_SUCCESS;
+    auth_status->auth_status = HDB_AUTHSTATUS_PKINIT_SUCCESS;
  out:
     if (pkp)
 	_kdc_pk_free_client_param(r->context, pkp);
@@ -565,7 +565,7 @@ pa_gss_validate(astgs_request_t r,
 	if (ret) {
 	    _kdc_set_e_text(r, "GSS-API client not allowed to "
 			    "impersonate principal");
-	    auth_status->auth_status = HDB_AUTH_GSS_FAILURE;
+	    auth_status->auth_status = HDB_AUTHSTATUS_GSS_FAILURE;
 	    goto out;
 	}
 	auth_status->auth_details = client_name;
@@ -590,7 +590,7 @@ pa_gss_validate(astgs_request_t r,
 	goto out;
     }
 
-    auth_status->auth_status = HDB_AUTH_GSS_SUCCESS;
+    auth_status->auth_status = HDB_AUTHSTATUS_GSS_SUCCESS;
  out:
     if (gcp)
 	_kdc_gss_free_client_param(r, gcp);
@@ -624,7 +624,7 @@ pa_enc_chal_validate(astgs_request_t r,
        ret = KRB5KDC_ERR_CLIENT_REVOKED;
        kdc_log(r->context, r->config, 0,
                "Client (%s) is locked out", r->cname);
-       auth_status->auth_status = HDB_AUTH_CLIENT_LOCKED_OUT;
+       auth_status->auth_status = HDB_AUTHSTATUS_CLIENT_LOCKED_OUT;
        return ret;
     }
 
@@ -749,12 +749,12 @@ pa_enc_chal_validate(astgs_request_t r,
 	/*
 	 * Success
 	 */
-	auth_status->auth_status = HDB_AUTH_CORRECT_PASSWORD;
+	auth_status->auth_status = HDB_AUTHSTATUS_CORRECT_PASSWORD;
 	goto out;
     }
 
     if (invalidPassword) {
-	auth_status->auth_status = HDB_AUTH_WRONG_PASSWORD;
+	auth_status->auth_status = HDB_AUTHSTATUS_WRONG_PASSWORD;
 	ret = KRB5KDC_ERR_PREAUTH_FAILED;
     } else {
 	ret = KRB5KDC_ERR_ETYPE_NOSUPP;
@@ -795,7 +795,7 @@ pa_enc_ts_validate(astgs_request_t r,
        ret = KRB5KDC_ERR_CLIENT_REVOKED;
        kdc_log(r->context, r->config, 0,
                "Client (%s) is locked out", r->cname);
-       auth_status->auth_status = HDB_AUTH_CLIENT_LOCKED_OUT;
+       auth_status->auth_status = HDB_AUTHSTATUS_CLIENT_LOCKED_OUT;
        return ret;
     }
 
@@ -866,7 +866,7 @@ pa_enc_ts_validate(astgs_request_t r,
 	krb5_free_error_message(r->context, msg);
 
 	free(auth_status->free_ptr);
-	auth_status->auth_status = HDB_AUTH_WRONG_PASSWORD;
+	auth_status->auth_status = HDB_AUTHSTATUS_WRONG_PASSWORD;
 	auth_status->auth_details = str ? str : "unknown enctype";
 	auth_status->free_ptr = str;
 
@@ -934,7 +934,7 @@ pa_enc_ts_validate(astgs_request_t r,
 	       r->cname, str ? str : "unknown enctype");
     _kdc_audit_addkv((kdc_request_t)r, 0, "pa-etype", "%d",
 		     (int)pa_key->key.keytype);
-    auth_status->auth_status = HDB_AUTH_CORRECT_PASSWORD;
+    auth_status->auth_status = HDB_AUTHSTATUS_CORRECT_PASSWORD;
     auth_status->auth_details = str ? str : "unknown enctype";
     auth_status->free_ptr = str;
 
@@ -2101,7 +2101,7 @@ _kdc_as_rep(astgs_request_t r)
     }
     default:
     {
-	struct kdc_pa_auth_status auth_status = {HDB_AUTH_CLIENT_UNKNOWN, NULL, NULL};
+	struct kdc_pa_auth_status auth_status = {HDB_AUTHSTATUS_CLIENT_UNKNOWN, NULL, NULL};
 	msg = krb5_get_error_message(r->context, ret);
 	kdc_log(r->context, config, 4, "UNKNOWN -- %s: %s", r->cname, msg);
 	krb5_free_error_message(r->context, msg);
@@ -2184,7 +2184,7 @@ _kdc_as_rep(astgs_request_t r)
 		    krb5_boolean default_salt;
 
 		    if (auth_status.auth_status == -1)
-			auth_status.auth_status = HDB_AUTH_GENERIC_FAILURE;
+			auth_status.auth_status = HDB_AUTHSTATUS_GENERIC_FAILURE;
 		    _kdc_audit_auth_status(r,
 					   &auth_status,
 					   pat[n].name);
@@ -2212,7 +2212,7 @@ _kdc_as_rep(astgs_request_t r)
 		r->replaced_reply_key = (pat[n].flags & PA_REPLACE_REPLY_KEY) != 0;
 
 		if (auth_status.auth_status == -1)
-			auth_status.auth_status = HDB_AUTH_GENERIC_SUCCESS;
+			auth_status.auth_status = HDB_AUTHSTATUS_GENERIC_SUCCESS;
 
 		_kdc_audit_auth_status(r,
 				       &auth_status,
@@ -2308,7 +2308,7 @@ _kdc_as_rep(astgs_request_t r)
 
     {
 	struct kdc_pa_auth_status auth_status
-		= {HDB_AUTHZ_SUCCESS,
+		= {HDB_AUTHSTATUS_AUTHORIZATION_SUCCESS,
 		   NULL,
 		   NULL};
 
--- a/kdc/kerberos5.c
+++ b/kdc/kerberos5.c
@@ -38,6 +38,39 @@
 #undef __attribute__
 #define __attribute__(X)
 
+struct kdc_pa_auth_status {
+    int auth_status;
+    const char *auth_details;
+    void *free_ptr;
+};
+
+static krb5_error_code
+_kdc_audit_auth_status(astgs_request_t r,
+		       struct kdc_pa_auth_status *status,
+		       const char *pa_type)
+{
+    struct HDB *hdb;
+    krb5_error_code ret = 0;
+
+    if (r->clientdb)
+	hdb = r->clientdb;
+    else
+	hdb = r->config->db[0];
+
+    if (hdb && hdb->hdb_auth_status)
+	ret = hdb->hdb_auth_status(r->context,
+				   hdb,
+				   r->client,
+				   &r->tv_start,
+				   r->addr,
+				   r->cname,
+				   status->auth_status,
+				   status->auth_details,
+				   pa_type);
+
+    return ret;
+}
+
 void
 _kdc_fix_time(time_t **t)
 {
@@ -461,7 +494,9 @@ _kdc_log_timestamp(astgs_request_t r, const char *type,
 #ifdef PKINIT
 
 static krb5_error_code
-pa_pkinit_validate(astgs_request_t r, const PA_DATA *pa)
+pa_pkinit_validate(astgs_request_t r,
+		   const PA_DATA *pa,
+		   struct kdc_pa_auth_status *auth_status)
 {
     pk_client_params *pkp = NULL;
     char *client_cert = NULL;
@@ -472,15 +507,19 @@ pa_pkinit_validate(astgs_request_t r, const PA_DATA *pa)
 	ret = KRB5KRB_AP_ERR_BAD_INTEGRITY;
 	_kdc_r_log(r, 4, "Failed to decode PKINIT PA-DATA -- %s",
 		   r->cname);
+	auth_status->auth_status = HDB_AUTH_PKINIT_FAILURE;
 	goto out;
     }
-    
+
     ret = _kdc_pk_check_client(r, pkp, &client_cert);
     if (ret) {
 	_kdc_set_e_text(r, "PKINIT certificate not allowed to "
 			"impersonate principal");
+	auth_status->auth_status = HDB_AUTH_PKINIT_FAILURE;
 	goto out;
     }
+    auth_status->auth_details = client_cert;
+    auth_status->free_ptr = client_cert;
 
     r->pa_endtime = _kdc_pk_endtime(pkp);
     if (!r->client->entry.flags.synthetic)
@@ -488,7 +527,6 @@ pa_pkinit_validate(astgs_request_t r, const PA_DATA *pa)
 
     _kdc_r_log(r, 4, "PKINIT pre-authentication succeeded -- %s using %s",
 	       r->cname, client_cert);
-    free(client_cert);
 
     ret = _kdc_pk_mk_pa_reply(r, pkp);
     if (ret) {
@@ -497,6 +535,8 @@ pa_pkinit_validate(astgs_request_t r, const PA_DATA *pa)
     }
     ret = _kdc_add_initial_verified_cas(r->context, r->config,
 					pkp, &r->et);
+
+    auth_status->auth_status = HDB_AUTH_PKINIT_SUCCESS;
  out:
     if (pkp)
 	_kdc_pk_free_client_param(r->context, pkp);
@@ -507,7 +547,9 @@ pa_pkinit_validate(astgs_request_t r, const PA_DATA *pa)
 #endif /* PKINIT */
 
 static krb5_error_code
-pa_gss_validate(astgs_request_t r, const PA_DATA *pa)
+pa_gss_validate(astgs_request_t r,
+		const PA_DATA *pa,
+		struct kdc_pa_auth_status *auth_status)
 {
     gss_client_params *gcp = NULL;
     char *client_name = NULL;
@@ -523,14 +565,16 @@ pa_gss_validate(astgs_request_t r, const PA_DATA *pa)
 	if (ret) {
 	    _kdc_set_e_text(r, "GSS-API client not allowed to "
 			    "impersonate principal");
+	    auth_status->auth_status = HDB_AUTH_GSS_FAILURE;
 	    goto out;
 	}
+	auth_status->auth_details = client_name;
+	auth_status->free_ptr = client_name;
 
 	r->pa_endtime = _kdc_gss_endtime(r, gcp);
 
 	_kdc_r_log(r, 4, "GSS pre-authentication succeeded -- %s using %s",
 		   r->cname, client_name);
-	free(client_name);
 
 	ret = _kdc_gss_mk_composite_name_ad(r, gcp);
 	if (ret) {
@@ -546,6 +590,7 @@ pa_gss_validate(astgs_request_t r, const PA_DATA *pa)
 	goto out;
     }
 
+    auth_status->auth_status = HDB_AUTH_GSS_SUCCESS;
  out:
     if (gcp)
 	_kdc_gss_free_client_param(r, gcp);
@@ -554,7 +599,9 @@ pa_gss_validate(astgs_request_t r, const PA_DATA *pa)
 }
 
 static krb5_error_code
-pa_enc_chal_validate(astgs_request_t r, const PA_DATA *pa)
+pa_enc_chal_validate(astgs_request_t r,
+		     const PA_DATA *pa,
+		     struct kdc_pa_auth_status *auth_status)
 {
     krb5_data pepper1, pepper2;
     int invalidPassword = 0;
@@ -577,6 +624,7 @@ pa_enc_chal_validate(astgs_request_t r, const PA_DATA *pa)
        ret = KRB5KDC_ERR_CLIENT_REVOKED;
        kdc_log(r->context, r->config, 0,
                "Client (%s) is locked out", r->cname);
+       auth_status->auth_status = HDB_AUTH_CLIENT_LOCKED_OUT;
        return ret;
     }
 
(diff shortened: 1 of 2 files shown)


kdc/kerberos5.c
```diff
@@ -38,6 +38,44 @@
 #undef __attribute__
 #define __attribute__(X)
 
+struct kdc_pa_auth_state {
```
lhoward, line 41: Indentation

kdc/kerberos5.c
```diff
@@ -799,22 +851,30 @@ pa_enc_ts_validate(astgs_request_t r, const PA_DATA *pa)
 		   "(enctype %s) error %s",
 		   r->cname, str ? str : "unknown enctype", msg);
 	krb5_free_error_message(r->context, msg);
-	free(str);
+
+	if (auth_state->free_ptr != NULL) {
```
lhoward, line 855: It’s always safe to free(NULL)

lhoward (changes requested):

kdc/kerberos5.c
```diff
@@ -799,22 +851,30 @@ pa_enc_ts_validate(astgs_request_t r, const PA_DATA *pa)
...
-
 	ret = KRB5KDC_ERR_PREAUTH_FAILED;
 	goto out;
     }
     free_EncryptedData(&enc_data);
+    if (auth_state->free_ptr != NULL) {
+	    free(auth_state->free_ptr);
+    }
+    *auth_state = (struct kdc_pa_auth_state) {
+	.auth_status = -1,
```
lhoward, line 876: Is this portable to all C compilers used to build Heimdal?

abartlet: Any clues on what the more exotic examples would be?  (This passed Samba CI on our platforms, but those are fairly modern)

lhoward: Just checking if C99 is our baseline, if so then designated initialisers are OK.

kdc/kerberos5.c
```diff
@@ -38,6 +38,44 @@
...
+    struct HDB *hdb = NULL;
+    int ret = 0;
+
+    if (r->clientdb != NULL) {
+	hdb = r->clientdb;
+    } else if (config->db[0]) {
+	hdb = config->db[0];
+    }
+
+    if (hdb != NULL && hdb->hdb_auth_status != NULL) {
```
lhoward, line 64: This is fine but, generally idiomatic to Heimdal is avoiding explicitly specifying NULL

kdc/kerberos5.c
```diff
@@ -38,6 +38,44 @@
...
+    krb5_context context = r->context;
+    krb5_kdc_configuration *config = r->config;
+    struct HDB *hdb = NULL;
+    int ret = 0;
+
+    if (r->clientdb != NULL) {
+	hdb = r->clientdb;
+    } else if (config->db[0]) {
+	hdb = config->db[0];
+    }
```
lhoward, line 62: Re: below comment, at least be consistent – either explicitly specify NULL, or don't :)

lhoward: Also, Heimdal style is generally to avoid unnecessary braces

lhoward: You could rewrite this as

```c
if (r->clientdb)
    hdb = r->clientdb;
else
    hub = config->db[0]
```

unless it was `config->db` you meant to check before dereferencing.

kdc/kdc_locl.h
```diff
@@ -81,6 +81,7 @@ struct astgs_request_desc {
     /* PA methods can affect both the reply key and the session key (pkinit) */
     krb5_enctype sessionetype;
     krb5_keyblock reply_key;
+    unsigned int reply_kvno;
```
lhoward, line 84: Use `krb5_kvno` or if it needs to be unsigned, `krb5uint32`.

kdc/kerberos5.c
```diff
@@ -461,7 +499,8 @@ _kdc_log_timestamp(astgs_request_t r, const char *type,
 #ifdef PKINIT
 
 static krb5_error_code
-pa_pkinit_validate(astgs_request_t r, const PA_DATA *pa)
+pa_pkinit_validate(astgs_request_t r, const PA_DATA *pa,
+		   struct kdc_pa_auth_state *auth_state)
```
lhoward, line 503: In this case please reindent so all parameters are on their own line (except for the first one)

kdc/kerberos5.c
```diff
@@ -472,31 +511,39 @@ pa_pkinit_validate(astgs_request_t r, const PA_DATA *pa)
...
+
     ret = _kdc_pk_check_client(r, pkp, &client_cert);
     if (ret) {
 	_kdc_set_e_text(r, "PKINIT certificate not allowed to "
 			"impersonate principal");
+	auth_state->auth_status = HDB_AUTH_PKINIT_FAILURE;
 	goto out;
     }
+    auth_state->auth_details = client_cert;
+    auth_state->free_ptr = client_cert;
```
lhoward, line 526: `free_ptr` feels a little unidiomatic Heimdal. Perhaps an integer indicating whether it should be freed would be better.

kdc/kerberos5.c
```diff
@@ -2111,6 +2189,10 @@ _kdc_as_rep(astgs_request_t r)
 	    i = 0;
 	    pa = _kdc_find_padata(req, &i, pat[n].type);
 	    if (pa) {
+		struct kdc_pa_auth_state auth_state = {
+			.auth_status = -1,
```
lhoward, line 2193: Indentation

kdc/kerberos5.c
```diff
@@ -2119,12 +2201,23 @@ _kdc_as_rep(astgs_request_t r)
...
 		    krb5_boolean default_salt;
 
+		    if (auth_state.auth_status == -1) {
+			auth_state.auth_status = HDB_AUTH_GENERIC_FAILURE;
+		    }
+		    _kdc_audit_auth_status(r,
+					   auth_state.auth_status,
+					   auth_state.auth_details,
+					   pat[n].name);
+		    if (auth_state.free_ptr != NULL) {
```
lhoward, line 2217: `free(NULL)` is OK, remove if

kdc/kerberos5.c
```diff
@@ -2119,12 +2201,23 @@ _kdc_as_rep(astgs_request_t r)
...
                 _kdc_audit_addkv((kdc_request_t)r, KDC_AUDIT_VIS, "pa", "%s",
                                  pat[n].name);
-		ret = pat[n].validate(r, pa);
+		ret = pat[n].validate(r, pa, &auth_state);
 		if (ret != 0) {
 		    krb5_error_code  ret2;
 		    Key *ckey = NULL;
 		    krb5_boolean default_salt;
 
+		    if (auth_state.auth_status == -1) {
```
lhoward, line 2210: Remove extraneous brace

kdc/kerberos5.c
lhoward: Unless you're going to replace it with a boolean that is.

kdc/kerberos5.c
abartlet: Thanks.  I'll fix that.  Samba style is that the optional braces are a disaster waiting to happen so we always include them, so I do apologise for the style creep and thank you for your patience with us. https://gitlab.com/samba-team/samba/-/blob/master/README.Coding.md

kdc/kerberos5.c
```diff
@@ -2143,7 +2236,26 @@ _kdc_as_rep(astgs_request_t r)
...
 		found_pa = 1;
+
+		/*
+		 * Set r->replaced_reply_key for PAC generation (tells
+		 * Samba to include the NT hash in the PAC for NTLM)
+		 * and in case the audit code would like a boolean
+		 * operator to know if the authentiation was PKINIT
+		 * (and similar) or ENC-TS/ENC-CHAL (where the long
+		 * term key is used).
+		 */
```
lhoward, line 2247: Prefer to keep the frontend AD-agnostic to the extent possible. There is already a comment in `generate_pac()`. I think you can remove this comment.

kdc/kerberos5.c
```diff
@@ -2143,7 +2236,26 @@ _kdc_as_rep(astgs_request_t r)
...
 		r->replaced_reply_key = (pat[n].flags & PA_REPLACE_REPLY_KEY) != 0;
+		if (auth_state.auth_status == -1) {
+			auth_state.auth_status = HDB_AUTH_GENERIC_SUCCESS;
+		}
+		_kdc_audit_auth_status(r,
+				       auth_state.auth_status,
+				       auth_state.auth_details,
+				       pat[n].name);
+		if (auth_state.free_ptr != NULL) {
```
lhoward, line 2256: See above

kdc/kerberos5.c
abartlet: OK.

lhoward:
No problem in principle with this PR (once the comments have been addressed), but would be curious to hear what @jaltman and @nicowilliams think, it seems we have two auditing frameworks – this one and `heim_audit_addkv()` and perhaps we need to take a step back and consolidate them?

kdc/kerberos5.c
nicowilliams: It's not, sadly.  @jaltman still needs to support older Microsoft compilers that don't support designated initializers.  I know, it's awful.  I desperately want to be able to use designated initializers!

kdc/kerberos5.c
```diff
@@ -507,7 +554,8 @@ pa_pkinit_validate(astgs_request_t r, const PA_DATA *pa)
 #endif /* PKINIT */
 
 static krb5_error_code
-pa_gss_validate(astgs_request_t r, const PA_DATA *pa)
+pa_gss_validate(astgs_request_t r, const PA_DATA *pa,
+		struct kdc_pa_auth_state *auth_state)
```
lhoward, line 558: Not filling in `auth_state` here?

kdc/kerberos5.c
```diff
@@ -38,6 +38,44 @@
 #undef __attribute__
 #define __attribute__(X)
 
+struct kdc_pa_auth_state {
+	int auth_status;
+	const char *auth_details;
+	void *free_ptr;
+};
```
lhoward, line 45: It's really auth *status*, not state.

kdc/kerberos5.c
```diff
@@ -38,6 +38,44 @@
...
+
+static krb5_error_code
+_kdc_audit_auth_status(astgs_request_t r,
```
lhoward, line 48: Also, could this take a struct rather than individual arguments; you have the struct.

kdc/kerberos5.c
abartlet: Looking at @josephsutton1's new patch, now I see why @metze-samba did it this way.  The distinct pointer is needed so that a const pointer is not passed to free().  This is allowed in Heimdal (perhaps with warnings) but Samba's stricter compile settings won't allow that.  So by keeping two copies of the pointer, it works when set to a real constant string (`"foo"`) and a `malloc()` based string.

So sorry, we probably do need that :-(

kdc/kerberos5.c
```diff
@@ -2052,6 +2121,14 @@ _kdc_as_rep(astgs_request_t r)
...
 	krb5_free_error_message(r->context, msg);
 	ret = KRB5KDC_ERR_C_PRINCIPAL_UNKNOWN;
+	_kdc_audit_auth_status(r, HDB_AUTH_CLIENT_UNKNOWN, NULL, NULL);
+	goto out;
+    }
+    if (r->client->entry.flags.locked_out) {
+	ret = KRB5KDC_ERR_CLIENT_REVOKED;
+	kdc_log(r->context, r->config, 0,
+		"Client (%s) is locked out", r->cname);
+	_kdc_audit_auth_status(r, HDB_AUTH_CLIENT_LOCKED_OUT, NULL, NULL);
```
abartlet, line 2131: This hunk looks wrong, I think the logic is from before #788 was merged as it would, if I read correctly, apply the locked_out logic to a GSS or PKINIT login. 

Either way, we should not be changing code flow in this PR, that should be in a distinct patch if needed later.

abartlet: It may be that the `_kdc_audit_auth_status()` bit is needed elsewhere, eg at the sites patched by 33fccb8bbec20b8a01263b629571404662b3a9c3

abartlet:
Per @lhoward's request, patch #2 shoudl start with `kdc: ` not `heimdal: `

`heimdal: Pass extra information to hdb_auth_status() to log success and failures`

jaltman:
Our preference is for the name of the directory or module followed by colon at the start of the subject line.  Thanks

abartlet:
Much the same in Samba, we just need to fix these up as they get ported over.  Thanks for your patience with us as we upstream these patches.  Hopefully this will mean we can much more easily keep up with current Heimdal and so both better participate upstream and provide timely feedback from our now quite extensive testsuite.

lhoward:
FYI I think there are still a few comments to be addressed on this one.

josephsutton1:
I think they are all addressed now, except for passing a struct to `_kdc_audit_auth_status()` rather than individual arguments.

kdc/kerberos5.c
```diff
@@ -38,6 +38,42 @@
...
+    void *free_ptr;
+};
+
+static krb5_error_code
+_kdc_audit_auth_status(astgs_request_t r,
+		       int auth_status,
+		       const char *auth_details,
+		       const char *pa_type)
+{
+    krb5_context context = r->context;
```
lhoward, line 53: You can remove this temporary context variable

kdc/kerberos5.c
```diff
@@ -38,6 +38,42 @@
...
+    krb5_kdc_configuration *config = r->config;
+    struct HDB *hdb = NULL;
```
lhoward, line 55: Remove NULL initializer, it’s set later and this may mask bugs if the subsequent code ever changes.

kdc/kerberos5.c
```diff
@@ -857,14 +918,21 @@ pa_enc_ts_validate(astgs_request_t r, const PA_DATA *pa)
     if (ret)
 	return ret;
 
+    if (pa_key->mkvno != NULL)
+	r->reply_kvno = *pa_key->mkvno;
```
nicowilliams, line 922: The mkvno is not the kvno.  It's the kvno of the master key.

We're mixing two different things here.

nicowilliams:
> kdc: use the correct kvno number for PKINIT in the AS-REP 

Makes no sense to me.  Why would we want to signal the master kvno instead of the client's kvno in the KDC reply's enc-part?

kdc/kerberos5.c
nicowilliams: It isn't Heimdal style.  When in Rome...  We don't mind the extra braces, but don't expect us to write them.

kdc/kerberos5.c
nicowilliams: I'm guilty of the `freeme` style...  I think this is OK.

nicowilliams:
> No problem in principle with this PR (once the comments have been addressed), but would be curious to hear what @jaltman and @nicowilliams think, it seems we have two auditing frameworks – this one and `heim_audit_addkv()` and perhaps we need to take a step back and consolidate them?

Ideally we could unify these two, however, `lib/hdb` can't have knowledge of the `r` structure we're using in the KDC -- well, it can have knowledge of the common part, since that's from `lib/base` IIRC.  Can you check if the `HEIM_SVC_REQUEST_DESC_COMMON_ELEMENTS` are sufficient for your purposes?

abartlet:
> > kdc: use the correct kvno number for PKINIT in the AS-REP
> 
> Makes no sense to me. Why would we want to signal the master kvno instead of the client's kvno in the KDC reply's enc-part?

Perhaps we should just drop that bit.  It was a patch by @metze-samba that was triggering conflicts if we didn't select before this one, but isn't important for the fundamental purpose here.  We can leave that private or pull to a different PR.

lhoward:
> > > kdc: use the correct kvno number for PKINIT in the AS-REP
> > 
> > 
> > Makes no sense to me. Why would we want to signal the master kvno instead of the client's kvno in the KDC reply's enc-part?
> 
> Perhaps we should just drop that bit. It was a patch by @metze-samba that was triggering conflicts if we didn't select before this one, but isn't important for the fundamental purpose here. We can leave that private or pull to a different PR.

Perhaps @metze-samba can explain the reasoning behind this? Are you sure the mkvno wasn't always zero or a well known constant? mkvno and kvno aren't interchangeable and the mkvno is never visible in the protocol.

abartlet:
Would swapping
```
  &r->tv_start,
  r->addr,
  r->cname,
```
for `hdb_r` (being a struct defined with HEIM_SVC_REQUEST_DESC_COMMON_ELEMENTS) be enough of a start on this?

The hdb hook is for account lockout and audit logging, and I agree there is some crossover, but it needs more than just the common elements as it needs the `client` structure (HDB entry) and the HDB database also, plus some other details, some of which are hidden into the `kv` and some not.

lhoward:
I'm OK with these changes (except for the mkvno bit). But, I'll defer final approval to @nicowilliams

abartlet:
OK, I'll leave as-is and just rip out the mkvno bit

lhoward:
> OK, I'll leave as-is and just rip out the mkvno bit

OK, but you might want to understand why you need it in your branch, for your own benefit :)

kdc/kerberos5.c
```diff
@@ -38,6 +38,41 @@
...
+
+static krb5_error_code
+_kdc_audit_auth_status(astgs_request_t r,
+		       int auth_status,
+		       const char *auth_details,
+		       const char *pa_type)
+{
+    krb5_kdc_configuration *config = r->config;
+    struct HDB *hdb;
+    int ret = 0;
```
lhoward, line 55: Sorry to keep finding things. `ret` should be `krb5_error_code`

`r->config` can be accessed directly without a temporary variable

lib/hdb/hdb.h
```diff
@@ -80,9 +80,18 @@ enum hdb_lockop{ HDB_RLOCK, HDB_WLOCK };
 #define HDB_CAP_F_SHARED_DIRECTORY      8
 
 /* auth status values */
-#define HDB_AUTH_SUCCESS		0
+#define HDB_AUTHZ_SUCCESS		0
```
lhoward, line 83: Is it `AUTH` or `AUTHZ`?

Also, `#define HDB_AUTH_UNKNOWN -1`?

abartlet: AUTHZ is because it was for authorization, eg it all worked, not just that it was authenticated with a password (for the password check hooks, which need to reset the bad pwd count).

lhoward: OK, then shouldn't the others be prefixed with `HDB_AUTHZ_` too? Looks a bit inconsistent. And perhaps the callback should be `hdb_authz_status()`?

abartlet: No, the rest are about authentication, not authorization.

abartlet: Well except the GSS bit, but that was just cargo-culted from PKINIT as none of us have really taken the time to pick out how that fits into this.

lhoward: GSS is also an authentication mechanism, in this context.

lhoward: I get what you're saying but `AUTHZ` just one case looks like a typo to me, but I don't mind too much. What do you think @nicowilliams?

abartlet: I can add comments to `hdb.h` if that helps?

lhoward: deferring to @nicowilliams

abartlet: eg 
```
diff --git a/lib/hdb/hdb.h b/lib/hdb/hdb.h
index 938b207034..4252e119b8 100644
--- a/lib/hdb/hdb.h
+++ b/lib/hdb/hdb.h
@@ -80,17 +80,17 @@ enum hdb_lockop{ HDB_RLOCK, HDB_WLOCK };
 #define HDB_CAP_F_SHARED_DIRECTORY      8
 
 /* auth status values */
-#define HDB_AUTHZ_SUCCESS              0
+#define HDB_AUTHZ_SUCCESS              0 /* Overall authorization success */
 #define HDB_AUTH_WRONG_PASSWORD                1
 #define HDB_AUTH_INVALID_SIGNATURE     2
-#define HDB_AUTH_CORRECT_PASSWORD      3
+#define HDB_AUTH_CORRECT_PASSWORD      3 /* Authentication success with a password */
 #define HDB_AUTH_CLIENT_UNKNOWN        4
 #define HDB_AUTH_CLIENT_LOCKED_OUT     5
 #define HDB_AUTH_GENERIC_SUCCESS       6
 #define HDB_AUTH_GENERIC_FAILURE       7
-#define HDB_AUTH_PKINIT_SUCCESS        8
+#define HDB_AUTH_PKINIT_SUCCESS        8 /* PKINIT authentication success */
 #define HDB_AUTH_PKINIT_FAILURE        9
-#define HDB_AUTH_GSS_SUCCESS           10
+#define HDB_AUTH_GSS_SUCCESS           10  /* GSS authentication success */
 #define HDB_AUTH_GSS_FAILURE           11
 
 /* key usage for master key */
```

nicowilliams:
> Would swapping ...

Basically.  I think this might be worth doing because whatever context you need to audit in HDB, others might want elsewhere too.  This bit of commonality was added by @elric1, and I find it quite convenient.  I especially like the bag of key/values, which can be used to add structured information.

nicowilliams:
I'm OK with `AUTH` in this context.  The boundary between authentication and authorization is a bit fuzzy in this context.

Anyways, LGTM w/o the `mkvno` commit.  If you can use `HEIM_SVC_REQUEST_DESC_COMMON_ELEMENTS` without casting it in stone, and if you want to, then go ahead and use it, else never mind and say you're ready.

abartlet:
I think having the HDB ABI change every time HEIM_SVC_REQUEST_DESC_COMMON_ELEMENTS changes would not be intuitive.  Harmless while samba requires to build with the source of Heimdal, but if we ever manage to use a system version, it would be trouble.

lib/hdb/hdb.h
lhoward: Could have HDB_AUTHZ_UNKNOWN -1 please?

lhoward: Also, again, really this doesn't make sense to me. It's conflating authorisation and authentication, they're in my mind orthogonal not subsets. I'd prefer more verbiage if it's clear what it means. e.g. `HDB_AUTHZ_AUTH_FAILED~WRONG_PASSWORD`.

abartlet: I've added comments.

jaltman: My take on the HDB_AUTH vs HDB_AUTHZ confusion is that these are constants that are supposed to describe auth_status values.  They are not intended to be general HDB constants.  Therefore, I believe the prefix should be tied to the name of the field. 

Perhaps HDB_AUTHSTATUS or something similar.  If it is so important to distinguish between AUTHN and AUTHZ then perhaps the field name needs to be changed as well.

jaltman: > Could have HDB_AUTHZ_UNKNOWN -1 please?

I would prefer UNKNOWN to be zero.

nicowilliams:
We have a way to do head-to-head ABI checking with the `*_get_instance()` functions...

lib/hdb/hdb.h
lhoward: That sounds good, for the other constants too?

lib/hdb/hdb.h
```diff
@@ -80,10 +80,76 @@ enum hdb_lockop{ HDB_RLOCK, HDB_WLOCK };
...
+ */
+#define HDB_AUTH_PKINIT_FAILURE	9
+
+/*
+ * Successful authentication with GSS authentication
+ */
+#define HDB_AUTH_GSS_SUCCESS		10
+
+/*
+ * Failed authentication with GSS authentication
```
lhoward, line 149: It's GSS-API pre-authentication (not authentication)

lib/hdb/hdb.h
```diff
@@ -80,10 +80,76 @@ enum hdb_lockop{ HDB_RLOCK, HDB_WLOCK };
...
+ */
+#define HDB_AUTH_GENERIC_FAILURE	7
+
+/*
+ * Successful authentication with PKINIT (smart card login etc)
+ */
+#define HDB_AUTH_PKINIT_SUCCESS	8
+
+/*
+ * Failed authentication with PKINIT (smart card login etc)
```
lhoward, line 139: s/authentication/pre-authentication/

lib/hdb/hdb.h
abartlet: Fixed.

lib/hdb/hdb.h
abartlet: Yes, they are consistent.

kdc/kerberos5.c
abartlet: Done.

kdc/kerberos5.c
abartlet: Dropped.

kdc/kerberos5.c
abartlet: fixed.

kdc/kerberos5.c
abartlet: Will push a fix shortly.

metze-samba:
> > > > kdc: use the correct kvno number for PKINIT in the AS-REP
> > > 
> > > 
> > > Makes no sense to me. Why would we want to signal the master kvno instead of the client's kvno in the KDC reply's enc-part?
> > 
> > 
> > Perhaps we should just drop that bit. It was a patch by @metze-samba that was triggering conflicts if we didn't select before this one, but isn't important for the fundamental purpose here. We can leave that private or pull to a different PR.
> 
> Perhaps @metze-samba can explain the reasoning behind this? Are you sure the mkvno wasn't always zero or a well known constant? mkvno and kvno aren't interchangeable and the mkvno is never visible in the protocol.

For PKINIT (and also FAST) responses, the AS-REP enc-part should not have a kvno field,
as the used encryption is not the longterm key that belongs to the kvno value.

I'm not sure about the mkvmo anymore, it might be related to this, where I want to return
more than one key for an account, so that password changes don't have temporary
failures regarding still existing tickets and possible account lockouts:
https://gitlab.com/samba-team/samba/-/merge_requests/664/diffs

I guess for now we can remove the if (pa_key->mkvno != NULL) logic and only use
r->reply_kvno = r->client->entry.kvno; instead, but that can be it's own pull request,
@abartlet do you want to change that and submit it as new pull request?

lhoward:
Perhaps the correct fix is, `if (r->replaced_reply_key)` then set `kvno` to zero?

`mkvno` has no business in the Kerberos 5 protocol, it's HDB-internal (I guess maybe it's exposed in the replication/administration protocol but that's not relevant to the KDC).

lhoward:
Also, do we need a new `reply_kvno` field? Can't we just make it `r->replaced_reply_key ? 0 : r->client->entry.kvno`?

metze-samba:
> Also, do we need a new `reply_kvno` field? Can't we just make it `r->replaced_reply_key ? 0 : r->client->entry.kvno`?

Why is a new 'bool replaced_reply_key' better than a new 'unsigned int reply_kvno;'?
Note r->client->entry.kvno is 'unsigned int' in struct HDB_entry...

As there's only a single place where r->client->entry.kvno should be used, I think
an explicit r->reply_kvno = r->client->entry.kvno; would be good there and it could
be (and may already is initialized to 0) default, so it just needs to be set in one place.

BTW: regarding the password history I just found hdb_kvno2keys() and choice_HDB_extension_data_hist_keys...

abartlet:
On Thu, 2021-12-16 at 16:50 -0800, metze-samba wrote:
> Why is a new 'bool replaced_reply_key' better than a new 'unsigned
> int reply_kvno;'?

Because we got r-> replace_reply_key anyway in #864 so we can re-use
that infrastructure.

Anyway, this should be on a new PR.

lhoward:
Would agree except `r->replaced_reply_key` is already there, so let's use it please :)

abartlet:
Lets move this discussion to #899 where I've done just that.

abartlet:
And are we finally ready for a merge on this one?  I think I've addressed everything now.

lhoward:
Please see #912. We're going to replace `hdb_auth_status` with a `hdb_audit` function with the following signature:

```c
/* auth event type enumeration */
#define HDB_AUTH_EVENT_INVALID                  -1
#define HDB_AUTH_EVENT_AUTHORIZATION_SUCCESS    0   /* all checks passed */
#define HDB_AUTH_EVENT_CLIENT_UNKNOWN           1   /* client unknown */
#define HDB_AUTH_EVENT_CLIENT_LOCKED_OUT        2   /* client locked out */
#define HDB_AUTH_EVENT_CLIENT_TIME_SKEW         3   /* client time skew */
#define HDB_AUTH_EVENT_LTK_PREAUTH_FAILED       4   /* long term key preauth failed */
#define HDB_AUTH_EVENT_LTK_PREAUTH_SUCCEEDED    5   /* long term key preauth succeeded */
#define HDB_AUTH_EVENT_PKINIT_SUCCEEDED         6   /* PKINIT preauth succeeded */
#define HDB_AUTH_EVENT_PKINIT_FAILED            7   /* PKINIT preauth succeeded */
#define HDB_AUTH_EVENT_GSS_PA_SUCCEEDED         8   /* GSS preauth succeeded */
#define HDB_AUTH_EVENT_GSS_PA_FAILED            9   /* GSS preauth failed */
#define HDB_AUTH_EVENT_UNK_PREAUTH_FAILED       10  /* unknown preauth failed */
#define HDB_AUTH_EVENT_UNK_PREAUTH_SUCCEEDED    11  /* unknown preauth succeeded */

/* auth event keys, query request with heim_audit_getkv() */
#define HDB_REQUEST_KV_AUTH_EVENT_TYPE          "#auth_event_type"      /* heim_number_t */
#define HDB_REQUEST_KV_AUTH_EVENT_DETAILS       "#auth_event_details"   /* heim_string_t */
#define HDB_REQUEST_KV_PA_NAME                  "pa"                    /* heim_string_t */

    krb5_error_code (*hdb_audit)(krb5_context context,
                                struct HDB *hdb,
                                hdb_entry_ex *client,
                                hdb_request_t r);
```

where `hdb_request_t` contains `HEIM_SVC_REQUEST_DESC_COMMON_ELEMENTS` such as the client address and name.

You should be able to pull everything you need out of this (you can cast it to `kdc_request_t` if you need more things).

As a result of unifying the auditing API, the auth status (now called event) and details are surfaced by calling `heim_audit_getkv(HDB_REQUEST_KV...)`. Let me know if you need some sample code.

lhoward:
My next question for the Samba folks is, what does peppering the KDC code with auth event (né status) auditing achieve? 

For example, you can figure out whether the client was locked out by checking for `HDB_REQUEST_KV_ERROR_CODE` == `KRB5KDC_ERR_CLIENT_REVOKED`.

You can determine which pre-authentication mechanism was tried by checking `HDB_REQUEST_KV_PA_NAME` and looking at the error code. Probably the only thing I can't see how you could easily disambiguate is long-term key pre-authentication failing due to a wrong password.

I'm not opposed to leaving this functionality in but I wonder if it could be simplified. Looking at your `hdb_samba4_auth_status()`, it seems that principally you want to know when pre-authentication succeeded or failed in order to manage lockout counts?

lhoward:
For example, I think want to collapse to a generic pre-auth succeeded/failed and let you discriminate based on the pre-auth mech that was tried.

lhoward:
OTOH, I guess, if you want to be agnostic to new PA mechs that use long-term keys, maybe it's better we do it on the front-end...

PR #889, opened 2021-12-14
kdc: Send ETYPE-INFO2 instead of PW-SALT for validated timestamp

josephsutton1:
This matches Windows behaviour.

### DIFF
--- a/kdc/kerberos5.c
+++ b/kdc/kerberos5.c
@@ -177,15 +177,21 @@ get_pa_etype_info2(krb5_context context,
     return 0;
 }
 
-static void
+static krb5_error_code
 set_salt_padata(METHOD_DATA *md, Salt *salt)
 {
     if (salt) {
-       realloc_method_data(md);
+       krb5_error_code ret = realloc_method_data(md);
+       if (ret)
+           return ret;
+       ret = der_copy_octet_string(&salt->salt,
+                                   &md->val[md->len - 1].padata_value);
+       if (ret)
+           return ret;
        md->val[md->len - 1].padata_type = salt->type;
-       der_copy_octet_string(&salt->salt,
-                             &md->val[md->len - 1].padata_value);
     }
+
+    return 0;
 }
 
 const PA_DATA*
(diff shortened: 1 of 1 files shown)


kdc/kerberos5.c
```diff
@@ -60,6 +60,123 @@ realloc_method_data(METHOD_DATA *md)
     return 0;
 }
 
+extern int _krb5_AES_SHA1_string_to_default_iterator;
+extern int _krb5_AES_SHA2_string_to_default_iterator;
+
+static krb5_error_code
+make_s2kparams(int value, size_t len, krb5_data **ps2kparams)
```
lhoward, line 67: A forward declaration would have resulted in less code churn, but I'm OK with this.

kdc/kerberos5.c
```diff
@@ -178,18 +178,12 @@ get_pa_etype_info2(krb5_context context,
 }
 
 static krb5_error_code
-set_salt_padata(METHOD_DATA *md, Salt *salt)
+set_salt_padata(krb5_context context,
```
lhoward, line 181: “This matches Windows behaviour” is a fine commit message for Samba, but not for Heimdal. Could you please explain what this fixes with reference to the relevant non-Microsoft specifications (e.g. RFC 4120).

kdc/kerberos5.c
lhoward: Actually, let's go with a forward decl to avoid churn.

lhoward:
FWIW from RFC 4120

```
   When the AS server is to include pre-authentication data in a
   KRB-ERROR or in an AS-REP, it MUST use PA-ETYPE-INFO2, not PA-ETYPE-
   INFO, if the etype field of the client's AS-REQ lists at least one
   "newer" encryption type.  Otherwise (when the etype field of the
   client's AS-REQ does not list any "newer" encryption types), it MUST
   send both PA-ETYPE-INFO2 and PA-ETYPE-INFO (both with an entry for
   each enctype).  A "newer" enctype is any enctype first officially
   specified concurrently with or subsequent to the issue of this RFC.
   The enctypes DES, 3DES, or RC4 and any defined in [RFC1510] are not
   "newer" enctypes.
```

lhoward:
Ah, I see, shouldn't we use `get_pa_etype_info_both()` for an AS-REP then?

josephsutton1:
That's probably the case, but I'll need to do some more testing against Windows to check what it does for different etypes.

kdc/kerberos5.c
nicowilliams: Please, yes.

lhoward:
No problem, please let us know and refactor to minimise the size of the diff.

PR #890, opened 2021-12-14
kdc: Don't conceal error code when using FAST

josephsutton1:
This matches Windows behaviour, which also places the error code in the outer error.

### DIFF
--- a/kdc/fast.c
+++ b/kdc/fast.c
@@ -384,7 +384,6 @@ _kdc_fast_mk_error(astgs_request_t r,
 	    return ret;
 	}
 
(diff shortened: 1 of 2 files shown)


lhoward:
Is this patch just to require Windows, or is it because Windows actually breaks when the error codes don't match? Because RFC 6113 says to ignore the outer error code, and it would be good not to reveal the error code.

josephsutton1:
So far we've been using this patch to ease checking the error code for KDC testing, but I'm not sure whether Windows requires an outer error code or not.

lhoward:
Perhaps best keep this as a private patch then, making testing easier isn't a good enough reason to increase information disclosure. Closing, please re-open if it turns out Windows is indeed not RFC 6113-compliant.

PR #891, opened 2021-12-15
kdc: Check PAC of FAST armor tickets

josephsutton1:
Look up the client name of the armor TGT in the database, and check that the PAC is valid.

### DIFF
--- a/kdc/fast.c
+++ b/kdc/fast.c
@@ -435,7 +435,9 @@ fast_unwrap_request(astgs_request_t r,
 		    krb5_auth_context tgs_ac)
 {
     krb5_principal armor_server = NULL;
+    krb5_principal armor_client_principal = NULL;
     hdb_entry_ex *armor_user = NULL;
+    hdb_entry_ex *armor_client = NULL;
     PA_FX_FAST_REQUEST fxreq;
     krb5_auth_context ac = NULL;
     krb5_ticket *ticket = NULL;
@@ -657,6 +659,100 @@ fast_unwrap_request(astgs_request_t r,
 	}
     }
 
+    if (fxreq.u.armored_data.armor) {
+	char *armor_server_principal_name = NULL;
+	int flags;
+	krb5_boolean ad_kdc_issued = FALSE;
+	krb5_pac mspac = NULL;
+
+	ret = krb5_unparse_name(r->context, armor_server,
+				&armor_server_principal_name);
+	if (ret)
+	    goto out;
+
+	ret = _kdc_verify_flags(r->context, r->config,
+				&ticket->ticket,
+				armor_server_principal_name);
+	free(armor_server_principal_name);
+	if (ret) {
+	    _kdc_audit_addreason((kdc_request_t)r,
+				 "Armor TGT expired or invalid");
+	    goto out;
+	}
+
+	if (tgs_ticket)
+	    flags = HDB_F_FOR_TGS_REQ;
+	else
+	    flags = HDB_F_FOR_AS_REQ;
+	if (fastreq.req_body.kdc_options.canonicalize)
+	    flags |= HDB_F_CANON;
+
+	ret = _krb5_principalname2krb5_principal(r->context,
+						 &armor_client_principal,
+						 ticket->ticket.cname,
+						 ticket->ticket.crealm);
+	if (ret)
+	    goto out;
+
+	ret = _kdc_db_fetch(r->context, r->config, armor_client_principal,
+			    HDB_F_GET_CLIENT | flags,
+			    NULL, NULL, &armor_client);
+	if (ret) {
+	    ret = KRB5KDC_ERR_C_PRINCIPAL_UNKNOWN;
+	    goto out;
+	}
+
+	ret = kdc_check_flags(r,
+			      tgs_ac == NULL,
+			      armor_client,
+			      NULL);
+	if (ret) {
+	    goto out;
+	}
+
+	ret = _kdc_check_pac(r->context, r->config, armor_client_principal, NULL,
+			     armor_client, armor_user,
+			     armor_user, armor_user,
+			     &armor_key->key, &armor_key->key,
+			     &ticket->ticket, &ad_kdc_issued, &mspac);
+	if (ret) {
+	    const char *msg = krb5_get_error_message(r->context, ret);
+	    char *cname = NULL;
+	    char *sname = NULL;
+
+	    krb5_unparse_name(r->context, armor_client_principal, &cname);
+
+	    if (fastreq.req_body.sname) {
+		krb5_principal server_princ = NULL;
+		if (!_krb5_principalname2krb5_principal(r->context,
+							&server_princ,
+							*fastreq.req_body.sname,
+							fastreq.req_body.realm)) {
+		    krb5_unparse_name(r->context, server_princ, &sname);
+		    krb5_free_principal(r->context, server_princ);
+		}
+	    }
+
+	    kdc_log(r->context, r->config, 4,
+		    "Verify PAC of armor ticket failed for %s (%s) from %s with %s",
+		    sname ? sname : "<unknown>",
+		    cname ? cname : "<unknown>",
+		    r->from, msg);
+	    free(sname);
+	    free(cname);
+	    krb5_free_error_message(r->context, msg);
+	    goto out;
+	}
+	if ((r->config->require_pac && !mspac) || (mspac && !ad_kdc_issued)) {
+	    ret = KRB5KDC_ERR_BADOPTION;
+	    kdc_log(r->context, r->config, 0,
+		    "Armor ticket not signed with PAC; FAST failed (%s).",
+		    mspac || !r->config->require_pac ? "Ticket unsigned" : "No PAC");
+	    goto out;
+	}
+	krb5_pac_free(r->context, mspac);
+    }
+
     /*
      * check for unsupported mandatory options
      */
@@ -695,10 +791,12 @@ fast_unwrap_request(astgs_request_t r,
     if (ticket && ticket != tgs_ticket)
 	krb5_free_ticket(r->context, ticket);
 
-    if (armor_server)
-	krb5_free_principal(r->context, armor_server);
+    krb5_free_principal(r->context, armor_server);
+    krb5_free_principal(r->context, armor_client_principal);
     if(armor_user)
 	_kdc_free_ent(r->context, armor_user);
+    if (armor_client)
+	_kdc_free_ent(r->context, armor_client);
 
     return ret;
 }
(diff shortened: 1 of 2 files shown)


kdc/fast.c
```diff
@@ -560,6 +568,83 @@ fast_unwrap_request(astgs_request_t r,
...
+
+	ret = krb5_unparse_name(r->context, armor_server,
+				&armor_server_principal_name);
+	if (ret)
+	    goto out;
+
+	ret = _kdc_verify_flags(r->context, r->config,
+				&ticket->ticket,
+				armor_server_principal_name);
+	free(armor_server_principal_name);
```
lhoward, line 580: `krb5_free_unparsed_name()`

kdc/fast.c
```diff
@@ -560,6 +568,83 @@ fast_unwrap_request(astgs_request_t r,
...
+	if (tgs_ticket != NULL)
+	    flags = HDB_F_FOR_TGS_REQ;
+	else
+	    flags = HDB_F_FOR_AS_REQ;
+	if (r->req.req_body.kdc_options.canonicalize)
+	    flags |= HDB_F_CANON;
+
+	/* Fetch the name from the TGT. */
+	ret = _krb5_principalname2krb5_principal(r->context,
+						 &fast_principal,
```
lhoward, line 596: `fast_principal` is a bit ambiguous. Maybe `armor_client_principal` would be more consistent.

kdc/fast.c
```diff
@@ -560,6 +568,83 @@ fast_unwrap_request(astgs_request_t r,
...
+						 ticket->ticket.cname,
+						 ticket->ticket.crealm);
+	if (ret)
+	    goto out;
+
+	/* Look up the name given in the TGT in the database. */
```
lhoward, line 602: Some of these comments that are explained by the code itself can probably be removed.

kdc/fast.c
```diff
@@ -560,6 +568,83 @@ fast_unwrap_request(astgs_request_t r,
...
+	/*
+	 * If the account is present in the database, check the account
+	 * flags.
+	 */
+	ret = kdc_check_flags(r,
+			      tgs_ac == NULL,
+			      fast_client,
+			      NULL);
+	if (ret) {
+	    _kdc_free_ent(r->context, fast_client);
```
lhoward, line 620: Can this be done in out block?

kdc/fast.c
```diff
@@ -560,6 +568,83 @@ fast_unwrap_request(astgs_request_t r,
...
+	    goto out;
+	}
+
+	/* Verify the PAC of the TGT. */
+	ret = _kdc_check_PAC(r->context, r->config, fast_principal, NULL,
```
lhoward, line 625: Will this work if there is no PAC?

kdc/fast.c
```diff
@@ -560,6 +568,83 @@ fast_unwrap_request(astgs_request_t r,
...
+			     fast_client, armor_user,
+			     armor_user, armor_user,
+			     &armor_key->key, &r->ticket_key->key,
+			     &ticket->ticket, &ad_kdc_issued, &mspac);
+	_kdc_free_ent(r->context, fast_client);
+	if (ret) {
+	    const char *msg = krb5_get_error_message(r->context, ret);
+
+	    kdc_log(r->context, r->config, 4,
+		    "Verify PAC failed for %s (%s) from %s with %s",
```
lhoward, line 635: Maybe note it's from armor ticket

kdc/fast.c
```diff
@@ -560,6 +568,83 @@ fast_unwrap_request(astgs_request_t r,
...
+		    r->sname, r->cname, r->from, msg);
+	    krb5_free_error_message(r->context, msg);
+	    goto out;
+	}
+	if (mspac == NULL || !ad_kdc_issued) {
+	    ret = KRB5KDC_ERR_BADOPTION;
+	    kdc_log(r->context, r->config, 0,
+		    "Armor ticket not signed with PAC; FAST failed (%s).",
+		    mspac ? "Ticket unsigned" : "No PAC");
```
lhoward, line 644: This may create a backwards compatibility issue for us because older Heimdal KDCs won't issue PACs unconditionally. I think you will need to hide this behind a feature flag.

kdc/krb5tgs.c
```diff
@@ -1814,8 +1814,8 @@ tgs_build_reply(astgs_request_t priv,
     flags &= ~HDB_F_SYNTHETIC_OK;
     priv->client = client;
 
-    ret = check_PAC(context, config, cp, NULL, client, server, krbtgt, krbtgt,
-		    &priv->ticket_key->key, &priv->ticket_key->key, tgt, &kdc_issued, &mspac);
+    ret = _kdc_check_PAC(context, config, cp, NULL, client, server, krbtgt, krbtgt,
```
lhoward, line 1817: you can lowercase PAC here

kdc/fast.c
abartlet: See #892 for the appropriate flag.

josephsutton1:
OK, I addressed the comments. This now relies on #892 for the `require_pac` flag.

josephsutton1:
Note that this patchset will fail to build until #892 is merged.

abartlet:
@josephsutton1 note comment by @lhoward https://github.com/heimdal/heimdal/commit/83a80cd53b70b8f6e5b75dc95d24155947611a7b#commitcomment-62004954

abartlet:
Looks like we need to allow the client not to exist (because it is anonymous) if we are willing to allow the PAC not to exist.

abartlet:
The `_kdc_db_fetch()` would have been `db_fetch_client()` from `krb5tgs.c` in an earlier version of this, but we tightened this up too much in our concerns about ensuring we really have the client and are not confused by realm aliases.

abartlet:
So also to allow a FAST amour from a trusted realm, we need to go back to `db_fetch_client()`, as well as the anonyous PKINIT thing.

lhoward:
OK, I'll back this out for now.

lhoward:
PS. Please put "Closes: #xxx" in the commit message.

lhoward:
Reverted. Can you also check all the other `check_pac` changes won't break cross-realm clients? Sorry, I should have thought of this.

abartlet:
I've been thinking about this, and we will drop attempting to merge this.  My work to fix it showed this is fairly complex (what is the right way to get our own realm, reliably?) and it makes no major improvement to security or functionality.

We will be back, but once we put the PAC into a device claim etc, thinking carefully about the cross-realm part.

lhoward:
What does Windows do?

abartlet:
Windows requires it, which is how we got here.  @josephsutton1 wrote
some awesome tests in Samba to explore all this, and these patches were
to make that pass.

lhoward:
Ok, well we should be able to verify local realm client names and all realm TGS signatures?

abartlet:
Yes. 

https://github.com/abartlet/heimdal/commit/1c3189dfe224e229517dc3895b5d21895b582ccb
is part of
https://github.com/abartlet/heimdal/commits/fix-fast

I just realised I was spending time fixing a patch that was beyond the
remit of 'get Heimdal upgraded and FAST into Samba' and I should focus
on the other patches.

Andrew,

-- 
Andrew Bartlett (he/him)        https://samba.org/~abartlet/
Samba Team Member (since 2001)  https://samba.org
Samba Developer, Catalyst IT    https://catalyst.net.nz/services/samba

lhoward:
If not checking the armour PAC is a vulnerability then, we should get a patch in to support it that does work with cross-realm armour clients.

lhoward:
(I guess it's only an issue with device claims.)

abartlet:
That's my assumption. It matters for the other cases (S4U2Self, User2User) because of the name confusion, but I don't see how it would matter here, as the name of the armor user is unused until device claims, which we hope to get to soonish (just not in this effort).

lhoward:
I reckon this patch is OK, because `_kdc_check_pac()` (based on the comments) can handle cross-realm clients. We just need the patch for synthetic clients in the local realm. Gonna re-merge.

abartlet:
See https://github.com/abartlet/heimdal/commits/fix-fast for how far I got trying to fix this.  Some subtle little things in here still I fear...

lhoward:
Ah, OK. Well, it doesn't affect us for now as PAC checking is by default disabled, so do you want to patch against master?

lhoward:
BTW, you *need* to check the PAC for constrained delegation to avoid a class of attacks, right? So the call to `_kdc_check_pac()` at the S4U2Proxy call site should add an additional argument `force_require_pac` or similar?

lhoward:
As in, even in the non-Samba case, because the PAC replaces KRB5SignedPath, which is used to protect the evidence ticket from being tampered with by the delegating service.

abartlet:
I thought the issue was with the `_kdc_db_fetch()` bit.  What if `armor_client` is in a different realm?

josephsutton1:
The constrained delegation case should be OK, because `mspac` is explicitly checked to be non-NULL after the call to `_kdc_check_pac()`.

lhoward:
> The constrained delegation case should be OK, because `mspac` is explicitly checked to be non-NULL after the call to `_kdc_check_pac()`.

Ah, excellent catch. I realise replacing `KRB5SignedPath` with the PAC runs contrary to our stated objective of not requiring a flag day so, maybe Heimdal 8 can require a PAC in all cases anyway, but we'll figure that out separately. Leave as is-is for now.

lhoward:
> I thought the issue was with the `_kdc_db_fetch()` bit. What if `armor_client` is in a different realm?

Right, so this should be no different to a TGT being in a different realm in the TGS path. So we need to allow `_kdc_db_fetch()` to fail?

lhoward:
Like `_kdc_db_fetch()` in krb5tgs.c, can't we just squash `HDB_ERR_NOT_FOUND_HERE` to 0?

lhoward:
More to the point, can't we just make `db_fetch_client()` public?

abartlet:
that's what I was doing, making `db_fetch_client()` public as `_kdc_db_fetch_client_for_tgs()` and therefore allowing HDB_ERR_NOENTRY if this was not our realm.  But finding our realm was non-trivial and I just realised this was a little more work than I had time for right now.

abartlet:
BTW, `HDB_ERR_NOT_FOUND_HERE` is for the RODC, where the full secrets are not on the entry, `HDB_ERR_NOENTRY` is for the name was not found in the DB.

lhoward:
I think our realm can just be `fastreq.req_body.realm`?

lhoward:
Please review #901.

lhoward:
Noted re: `HDB_ERR_NOT_FOUND_HERE`. Moving further comments to #901.

PR #895, opened 2021-12-15
Optimistic anon fast

nicowilliams:


lib/krb5/init_creds_pw.c
```diff
@@ -2954,6 +2953,19 @@ krb5_init_creds_set_fast_anon_pkinit(krb5_context context,
     return 0;
 }
 
+KRB5_LIB_FUNCTION krb5_error_code KRB5_LIB_CALL
+krb5_init_creds_set_fast_anon_pkinit_optimistic(krb5_context context,
```
lhoward, line 2957: Do we really need a new API with a very long function name? :) How about a flag?

Or even marking `krb5_init_creds_set_fast_ccache()` as deprecated and a new `krb5_init_creds_set_fast()` that takes a ccache and `KRB5_FAST_XXX` flags (which we could make public)? (Although in that case probably `KRB5_FAST_OPTIMISTIC` shouldn't also require `KRB5_FAST_REQUIRED` to be set, that doesn't make sense.)

```c
KRB5_LIB_FUNCTION krb5_error_code KRB5_LIB_CALL
krb5_init_creds_set_fast(krb5_context context,
    krb5_init_creds_context ctx,
    krb5_ccache fast_ccache,
    krb5_flags in_flags,
    krb5_flags *out_flags);
```

If an armour cache is provided, `KRB5_FAST_KDC_VERIFIED` would be implied (as we will never save to a ccache if the KDC was not verified; there is no public API for this).

However, `KRB5_FAST_KDC_VERIFIED` would be unset on `out_flags` if anonymous PKINIT was used with no KDC verification.

BTW, there are a bunch of flags (inherited from Apple) which aren't used. Perhaps we should set `KRB5_FAST_KDC_REPLY_KEY_REPLACED`.

PR #899, opened 2021-12-17
kdc: Do not set a kvno in the reply for PKINIT

abartlet:
Based on discussion at https://github.com/heimdal/heimdal/pull/888#issuecomment-996308872
in turn based on a patch by Stefan Metzmacher <metze@samba.org>

### DIFF
--- a/kdc/kerberos5.c
+++ b/kdc/kerberos5.c
@@ -1949,6 +1949,7 @@ _kdc_as_rep(astgs_request_t r)
     const char *msg;
     hdb_entry_ex *krbtgt = NULL;
     Key *krbtgt_key;
+    int client_reply_kvno;
 
     memset(&rep, 0, sizeof(rep));
 
@@ -2567,14 +2568,20 @@ _kdc_as_rep(astgs_request_t r)
     }
 
     /*
-     *
+     * We need to avoid sending a client KVNO if PKINIT or GSS was
+     * used, as they are not related to a long-term key.
      */
+
+    if (r->replaced_reply_key)
+	client_reply_kvno = 0;
+    else
+	client_reply_kvno = r->client->entry.kvno;
 
     ret = _kdc_encode_reply(r->context, config,
 			    r, req->req_body.nonce,
 			    &rep, &r->et, &r->ek, setype,
 			    r->server->entry.kvno, &skey->key,
-			    r->client->entry.kvno,
+			    client_reply_kvno,
 			    0, &r->e_text, r->reply);
     if (ret)
 	goto out;


kdc/kerberos5.c
```diff
@@ -1949,6 +1949,7 @@ _kdc_as_rep(astgs_request_t r)
     const char *msg;
     hdb_entry_ex *krbtgt = NULL;
     Key *krbtgt_key;
+    int client_reply_kvno;
```
lhoward, line 1952: Not to be too fussy but, should be `unsigned int` to match definition in `hdb_asn1.h`.

lhoward: Or `krb5_kvno`.

kdc/kerberos5.c
```diff
@@ -2567,14 +2568,20 @@ _kdc_as_rep(astgs_request_t r)
...
+	client_reply_kvno = 0;
+    else
+	client_reply_kvno = r->client->entry.kvno;
+
     ret = _kdc_encode_reply(r->context, config,
 			    r, req->req_body.nonce,
 			    &rep, &r->et, &r->ek, setype,
 			    r->server->entry.kvno, &skey->key,
-			    r->client->entry.kvno,
+			    client_reply_kvno,
```
lhoward, line 2584: I'd be happy with `r->replaced_reply_key ? 0 : r->client->entry.kvno` here but an extra variable is fine. However please update comment to reflect the code, i.e. checking for `r->replaced_reply_key` rather than PKINIT or GSS. Otherwise, when a new FAST factor is added that replaces the reply key, the comment will be out of date.

```c
/*
 * Don't send client entry kvno if the pre-authentication mechanism
 * replaced the reply key.
 */
```

lhoward:
Comments notwithstanding LGTM.

kdc/kerberos5.c
metze-samba: @lhoward why is KRB5_PADATA_ENCRYPTED_CHALLENGE not marked with PA_REPLACE_REPLY_KEY
in 2087e07c1ef53b11163e9efdff71becdbc0212f7.

I did that in my original patch as the reply key is not the clients longterm key.

lhoward: But, it is derived from the client’s long-term key, which is what we are concerned with when deciding whether to include PAC_CREDENTIAL_INFO in the returned PAC. I will replace with PA_USES_LONG_TERM_KEY.

lhoward: Ah, but then we also have the issue of whether to set `kvno` to zero. So we need `PA_REPLACE_REPLY_KEY` as well. Gotcha.

lhoward: Fixed in e89848719a7455b607b84180da34236ffd100f9b

Issue #871, opened 2021-12-03
Samba "lorikeet-heimdal" patches to be merged


abartlet:
This issue is just a heads-up to list somewhere the outstanding patches on the Samba side that we are trying to upstream, and possibly discuss which patches are of most interest so these can be proposed as a priority.

https://gitlab.com/samba-team/devel/lorikeet-heimdal/-/commits/lorikeet-heimdal-202112010221-patched/

Once #805 is merged we can propose a lot more of these, as most rely on or improve the FAST code (plus that is the context in which they were tested on the Samba side). 

The technique in lorikeet-hiemdal is that we regularly rebase that tree, giving the branch a new name each time, so into the future just watch out for new similar branch names in the repo.  We then import that into Samba and attempt to get our tests to pass.  The process will repeat until we get as many patches as we can into Heimdal and all of Samba's tests passing.

abartlet:
Many thanks for the assistance merging many of these already.  With #805 merged we are more confident that the patches being proposed are those tested in Samba so they will be coming more quickly (plus there are a few to refine the FAST support).

lhoward:
You mean they’re more? ;)

lhoward:
Can you s/^heimdal:// on the commit messages so they match ours? Makes preparing NEWS easier.

abartlet:
> Can you s/^heimdal:// on the commit messages so they match ours? Makes preparing NEWS easier.

Sure, I'll do that. I'll also check that the Samba commit hashes actually mean something (they are generated by a script that made more sense when the patches had already landed in Samba master, which isn't the case as often in this phase).

The current set of outstanding patches is here: https://gitlab.com/samba-team/devel/lorikeet-heimdal/-/commits/lorikeet-heimdal-202112140728/

lhoward:
With your 0b9c113c, can you make it conditional on `force_canonicalize` in the server principal?

lhoward:
With your 053e6709, requiring the PAC to always be present might create some interoperability problems for non-Samba realms. I need to check if mixed KDC deployments are supported.

lhoward:
Your de90e29b is replaced by #875

lhoward:
Your  a9722322  is replaced by #875

lhoward:
Your 4ef7f213 incorrectly identifies RFC8009 enctypes as "older" enctypes. This is incorrect.

lhoward:
Your 828edc36 will trigger the following in libkrb5:

```c
[...]

PR #900, opened 2021-12-17
tgs-rep: always return canonical realm even if canonicalize falg is not set, same as Windows.

abartlet:
This is an import of an older Samba commit, updated over time in our lorikeet-heimdal tree we would like to get upstream if possible.

The AD-behaviour regression was introduced by upstream commit:
378f34b4be9865ed3949918fba8d2dd877b395c0

Signed-off-by: Isaac Boukris <iboukris@gmail.com>

[abartlet@samba.org Similar to Samba commit a9e6119ca0c2a78ef314c3162122539ee834aa04
 but made conditional on server->entry.flags.force_canonicalize to
 allow upstream submission]

### DIFF
--- a/kdc/krb5tgs.c
+++ b/kdc/krb5tgs.c
@@ -611,7 +611,11 @@ tgs_make_reply(astgs_request_t r,
     if(ret)
 	goto out;
 
-    ret = copy_Realm(&server_principal->realm, &rep.ticket.realm);
+    if (server->entry.flags.force_canonicalize)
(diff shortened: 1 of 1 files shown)


abartlet:
Updated as suggested by @lhoward here: https://github.com/heimdal/heimdal/issues/871#issuecomment-993319327

lhoward:
Reverting this commit as it is superfluous. The server principal name is canonicalised prior to calling `tgs_make_reply()` if `force_canonicalize` is set on the server principal entry.

PR #901, opened 2021-12-17
kdc: allow cross-realm FAST armor TGT

lhoward:
08e0305b introduced a patch to validate armor ticket PACs, but required that the armor client principal was in the local realm (as it did not allow _kdc_db_fetch() to fail).

Allwo cross-realm FAST armor clients by using the same logic to look up the client principal as the TGS itself does, i.e. use db_fetch_client() which ignores HDB_ERR_NOT_FOUND_HERE.

### DIFF
--- a/kdc/fast.c
+++ b/kdc/fast.c
@@ -434,21 +434,18 @@ fast_unwrap_request(astgs_request_t r,
 		    krb5_ticket *tgs_ticket,
 		    krb5_auth_context tgs_ac)
 {
-    krb5_principal armor_server = NULL;
-    krb5_principal armor_client_principal = NULL;
-    hdb_entry_ex *armor_user = NULL;
-    hdb_entry_ex *armor_client = NULL;
-    PA_FX_FAST_REQUEST fxreq;
+    krb5_principal armor_server_principal = NULL;
+    char *armor_server_principal_name = NULL;
+    PA_FX_FAST_REQUEST fxreq = {0};
     krb5_auth_context ac = NULL;
     krb5_ticket *ticket = NULL;
     krb5_flags ap_req_options;
-    Key *armor_key = NULL;
     krb5_keyblock armorkey;
     krb5_keyblock explicit_armorkey;
     krb5_boolean explicit_armor;
     krb5_error_code ret;
     krb5_ap_req ap_req;
-    KrbFastReq fastreq;
+    KrbFastReq fastreq = {0};
     const PA_DATA *pa;
     krb5_data data;
     size_t len;
@@ -519,7 +516,7 @@ fast_unwrap_request(astgs_request_t r,
 
 	/* Save that principal that was in the request */
 	ret = _krb5_principalname2krb5_principal(r->context,
-						 &armor_server,
+						 &armor_server_principal,
 						 ap_req.ticket.sname,
 						 ap_req.ticket.realm);
 	if (ret) {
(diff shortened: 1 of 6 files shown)


lhoward:
OK, so is the realm from the request body the right realm? Is there an interaction with referrals we need to consider?

abartlet:
What locks `fastreq.req_body.realm` to being our realm, given this is a security check?

lhoward:
You're right, it is complicated...

lhoward:
If are actually going to issue a non-referral ticket with a PAC (the only case where validating the armour PAC is interesting from a security perspective, I think?) then it will match our realm.

But at the time of validating the armour ticket we probably don't know whether we're issuing a referral or not.

So maybe we are going to need a `armor_pac_validated` bit in the request that can triggers an error if we *don't* issue a referral?

lhoward:
As I understand it, we only need to validate the armour PAC when device claims are being inserted in a service ticket, so we can probably ignore the validation result when we are issuing a referral.

However we might be getting to the point where we need to check with dochelp.

lhoward:
Obviously that assumes that the device claims aren't inserted at the AS-REQ point but, I think this might actually be behind the design decision to allow [explicit armour](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-kile/ee7826f5-d985-4a6b-a167-0df0525daea0), which is an extension to RFC 6113.

lhoward:
Also, you want the device claims from the device you're requesting the service ticket on, not the device you initially acquired the TGT (normally these will be the same but, TGTs can be forwarded, e.g. over RDP).

lhoward:
OK, I need to think this through more, but I *think* it's OK to only validate the armour PAC when there's explicit armour on a TGS-REQ, and we're not issuing a referral.

lhoward:
Or, obviously, the alternative is to check the PAC just at the very end when we know we're issuing a service ticket...

abartlet:
@josephsutton1 can you look over this for @lhoward ?

lhoward:
Hmm, something about this doesn’t feel quite right yet, but I’m not sure what :)

lhoward:
My current theory is that we only need to require a validated armour PAC if `add_ticket_sig` is set in `tgs_build_reply()`.

{code}
    /*
     * Service tickets can have device claims added, which requires
     * the PAC in the armor ticket be validated.
     */
    if (add_ticket_sig && priv->armor_crypto &&
        priv->config->require_pac && !priv->validated_armor_pac) {
        ret = KRB5KDC_ERR_BADOPTION;
        kdc_log(priv->context, priv->config, 0,
                "Armor ticket missing valid PAC when issuing service ticket");
        goto out;
    }
{code}

But, maybe we need to move this check out of `fast_unwrap_request()` and only do it when we need to.

lhoward:
Factored out into a new function, `_kdc_fast_check_armor_pac()`, which only gets called when we are about to issue a service ticket. @josephsutton1 could you please review this?

josephsutton1:
Sorry for not reviewing this earlier. Testing on Windows shows that device claims are only added for a TGS-REQ with an explicit armor ticket, so only checking the PAC in that case seems to make sense. The only issue I found is that `r->client_princ` and `r->server_princ` are not set for a TGS-REQ, which is the case when the PAC is being checked, so krb5_unparse_name() crashes when trying to unparse them. I made a pull request for that [here](https://github.com/heimdal/heimdal/pull/903).

lhoward:
> Sorry for not reviewing this earlier. Testing on Windows shows that device claims are only added for a TGS-REQ with an explicit armor ticket, so only checking the PAC in that case seems to make sense. The only issue I found is that `r->client_princ` and `r->server_princ` are not set for a TGS-REQ, which is the case when the PAC is being checked, so krb5_unparse_name() crashes when trying to unparse them. I made a pull request for that [here](https://github.com/heimdal/heimdal/pull/903).

Thanks for the fix, has been applied.

Yes, it only makes sense to add device claims with explicit armour, otherwise forwarded TGTs would represent the device the user authenticated on, rather than the device the user is accessing the service from.

PR #902, opened 2021-12-19
Add canonical client principal to PAC

lhoward:
KDC and libkrb5 support for the UPN_DNS_INFO_EX PAC info buffer, which contains the canonical client name from the TGT.


kdc/kerberos5.c
```diff
@@ -1860,6 +1861,36 @@ generate_pac(astgs_request_t r, const Key *skey, const Key *tkey)
 	return ret;
     }
 
+    will_canonicalize = r->req.req_body.kdc_options.canonicalize ||
+	r->client->entry.flags.force_canonicalize;
+
+    /*
+     * For non-synthetic principals which match the realm of the issued
```
nicowilliams, line 1868: What about synthetic clients though?  Why isn't their name canonical?

lhoward: Per out-of-band conversation, because they're ephemeral. But, really, because I'm concerned this path might break TktBridgeAP. So yes, this check can probably be removed.

lhoward:
cc @josephsutton1

lib/krb5/krb5.conf.5
```diff
@@ -518,6 +518,10 @@ If this flag is true, then all application protocol authentication
 requests will be flagged to indicate that the application supports
 channel bindings when operating over a secure channel.
 The default value is false.
+.It Li gss_canonicalize_source_name = Va boolean
```
nicowilliams, line 521: Why default to false?  Why not true?  What might break?  Why have this be configurable?  Why is this in GSS and not in the `rd_req` side in `lib/krb5`?  Wouldn't we want this to be used for non-GSS apps?  (Well, we don't want non-GSS krb5 apps, but that's another story.)

If this should move to `lib/krb5`, maybe rename this param to `use_client_name_from_pac`.

nicowilliams: Also, with the `Principal` decoration I'm adding, maybe we could stuff the canonical and alias names in there and provide an accessor (name attributes).

nicowilliams:
LGTM except for the config matter (see commentary).

lib/krb5/krb5.conf.5
lhoward: > Why default to false? Why not true? What might break? Why have this be configurable? Why is this in GSS and not in the `rd_req` side in `lib/krb5`? Wouldn't we want this to be used for non-GSS apps? (Well, we don't want non-GSS krb5 apps, but that's another story.)
> 
> If this should move to `lib/krb5`, maybe rename this param to `use_client_name_from_pac`.

Can move to `lib/krb5`, good idea. Defaulting to TRUE will change behaviour where aliases are used without the `canonicalize` flag which would break @jaltman's use case.

lhoward: Also, note, because of the rd_req API, this will involve replacing `ticket->client` with the canonical name from the PAC; the ticket client will be lost forever.

PR #904, opened 2021-12-23
kdc: Return PA-SUPPORTED-ENCTYPES to indicate TGS etype support

josephsutton1:
In an AS-REQ exchange with the TGS, we now return PA-SUPPORTED-ENCTYPES containing supported enctypes of the TGS if the canonicalize flag is set. We also set the FAST-supported bit to let Windows client know that FAST is available.

Setting the FAST-supported bit in this padata type informs Windows that FAST support is available, and Windows won't try FAST if it is not set.

### DIFF
--- a/kdc/kerberos5.c
+++ b/kdc/kerberos5.c
@@ -2032,6 +2032,59 @@ add_enc_pa_rep(astgs_request_t r)
 			   KRB5_PADATA_FX_FAST, NULL, 0);
 }
 
+static krb5_error_code
+add_supported_enctypes(astgs_request_t r)
+{
+    krb5_error_code ret;
+    uint32_t *supported_enctypes = NULL;
+    uint32_t bit = 1;
+    Key *key = NULL;
+
+    static const krb5_enctype enctypes[] = {
+	ETYPE_DES_CBC_CRC,
+	ETYPE_DES_CBC_MD5,
+	ETYPE_ARCFOUR_HMAC_MD5,
+	ETYPE_AES128_CTS_HMAC_SHA1_96,
+	ETYPE_AES256_CTS_HMAC_SHA1_96,
+	ETYPE_NULL
+    };
+    const krb5_enctype *p = enctypes;
+
+    if (!r->ek.encrypted_pa_data) {
+	ALLOC(r->ek.encrypted_pa_data);
+	if (!r->ek.encrypted_pa_data)
+	    return ENOMEM;
+    }
+
+    supported_enctypes = malloc(sizeof *supported_enctypes);
+    if (!supported_enctypes)
+	return ENOMEM;
+
+    /* Indicate support for FAST */
+    *supported_enctypes = 0x00010000;
+
+    for (; *p != (krb5_enctype)ETYPE_NULL; ++p, bit <<= 1) {
+	if (krb5_enctype_valid(r->context, *p) != 0 &&
+	    !_kdc_is_weak_exception(r->server->entry.principal, *p))
+	    continue;
+
+	/*
+	 * If the encryption type is enabled and we have a key for it, indicate
+	 * that we support it by setting the respective bit.
+	 */
+	if (hdb_enctype2key(r->context,
+			    &r->server->entry,
+			    NULL,
+			    *p,
+			    &key) == 0)
+	    *supported_enctypes |= bit;
+    }
+
+    return krb5_padata_add(r->context, r->ek.encrypted_pa_data,
+			   KRB5_PADATA_SUPPORTED_ETYPES,
+			   supported_enctypes, sizeof *supported_enctypes);
+}
+
 /*
  * Add an authorization data element indicating that a synthetic
  * principal was used, so that the TGS does not accidentally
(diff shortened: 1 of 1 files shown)


kdc/kerberos5.c
```diff
@@ -2032,6 +2032,59 @@ add_enc_pa_rep(astgs_request_t r)
...
+	if (!r->ek.encrypted_pa_data)
+	    return ENOMEM;
+    }
+
+    supported_enctypes = malloc(sizeof *supported_enctypes);
+    if (!supported_enctypes)
+	return ENOMEM;
+
+    /* Indicate support for FAST */
+    *supported_enctypes = 0x00010000;
```
lhoward, line 2064: Symbolic name please?

kdc/kerberos5.c
```diff
@@ -2032,6 +2032,59 @@ add_enc_pa_rep(astgs_request_t r)
...
+	 */
+	if (hdb_enctype2key(r->context,
+			    &r->server->entry,
+			    NULL,
+			    *p,
+			    &key) == 0)
+	    *supported_enctypes |= bit;
+    }
+
+    return krb5_padata_add(r->context, r->ek.encrypted_pa_data,
```
lhoward, line 2083: This code needs to be made portable (i.e. work on big-endian systems).

kdc/kerberos5.c
```diff
@@ -2032,6 +2032,59 @@ add_enc_pa_rep(astgs_request_t r)
 			   KRB5_PADATA_FX_FAST, NULL, 0);
 }
 
+static krb5_error_code
+add_supported_enctypes(astgs_request_t r)
```
lhoward, line 2036: Please put a comment referencing [MS-KILE] 2.2.7 so we know that this is a MS extension to Kerberos.

kdc/kerberos5.c
```diff
@@ -2032,6 +2032,59 @@ add_enc_pa_rep(astgs_request_t r)
...
+	ETYPE_DES_CBC_CRC,
+	ETYPE_DES_CBC_MD5,
+	ETYPE_ARCFOUR_HMAC_MD5,
+	ETYPE_AES128_CTS_HMAC_SHA1_96,
+	ETYPE_AES256_CTS_HMAC_SHA1_96,
+	ETYPE_NULL
+    };
+    const krb5_enctype *p = enctypes;
+
+    if (!r->ek.encrypted_pa_data) {
```
lhoward, line 2053: Given this is also done by `add_enc_pa_rep()`, perhaps we need a helper version of `krb5_padata_add()` that checks and adds this. e.g. `add_enc_pa_data()` or `add_encrypted_pa_data()`.

kdc/kerberos5.c
```diff
@@ -2032,6 +2032,59 @@ add_enc_pa_rep(astgs_request_t r)
...
+	if (hdb_enctype2key(r->context,
+			    &r->server->entry,
+			    NULL,
+			    *p,
+			    &key) == 0)
+	    *supported_enctypes |= bit;
+    }
+
+    return krb5_padata_add(r->context, r->ek.encrypted_pa_data,
+			   KRB5_PADATA_SUPPORTED_ETYPES,
```
lhoward, line 2084: Should also add this to `kdc_patypes`, it's not clear why it needs to be there but, `KRB5_PADATA_REQ_ENC_PA_REP` is also there.

kdc/kerberos5.c
```diff
@@ -2745,6 +2798,22 @@ _kdc_as_rep(astgs_request_t r)
 	}
     }
 
+    /*
+     * Add SUPPORTED_ENCTYPES if the canonicalize bit is set. Until
+     * support is added for KerbSupportedEncryptionTypes and we can
+     * indicate what encryption types a given principal supports, we
+     * do this only for an AS-REQ exchange with the TGS.
+     */
+    if (f.canonicalize && is_tgs) {
```
lhoward, line 2807: What's the rationale behind testing for `f.canonicalize`? To not send this to non-Windows clients? Also, why for TGS requests only? 

[MS-KILE] 3.3.5.6 just says:

>The KDC SHOULD<50> return in the encrypted part of the AS-REP message a PA-DATA structure with padata-type set to PA-SUPPORTED-ENCTYPES [165] (section 2.2.8), to indicate what encryption types (section 2.2.7) are supported by the KDC, and whether Claims or FAST are supported.<51>

kdc/kerberos5.c
lhoward: e.g. `KRB5_PA_SUPPORTED_ENCTYPES_FAST` or something

lhoward: or `KRB5_PA_SUPPORTED_ENCTYPES_FAST_AVAILABLE`

kdc/kerberos5.c
lhoward: Ideally we would put this into the backend given it's very Windows-specific... but, I guess, we don't want to expose `astgs_request_desc` to the plugin API...

lhoward:
Given this is very Windows-specific, how about we add something like:

```c
typedef krb5_error_code
(KRB5_CALLCONV *krb5plugin_windc_finalize_reply)(
        void *,
        krb5_context context,
        krb5_kdc_configuration *config,
        hdb_entry_ex *client,
        hdb_entry_ex *server,
        KDC_REP *rep,
        EncTicketPart *et,
        EncKDCRepPart *ek);
```

to the WinDC interface?

lhoward:
This could get called immediately prior to encoding the ticket.

lhoward:
See #905, I reckon this might be a better approach that doesn't pollute the KDC so much. We're OK with the PAC because it lets us replace `KRB5SignedPath`, but I'm not sure we want to import Windows-specific PA types that we don't need.

@nicowilliams what think you?

lhoward:
I'm going to close this because it can now be implemented purely in Samba by implementing a `finalize_reply()` method in the windc plugin.

kdc/kerberos5.c
josephsutton1: From testing, I found that Windows only seems to add this padata type if the `canonicalize` bit is set, so this was really just to match Windows. This only being for TGS requests was to avoid having to fetch the msDS-SupportedEncryptionTypes attribute, until proper integration for it was added.

josephsutton1:
Thanks for the feedback, and for adding the plugin hook! I'll look into implementing this on the Samba side.

lhoward:
Yeah, I figure this is the way to go because then you can pretty much do anything you want in both the AS and TGS prior to sending the reply, without having to file a PR :)

We can add more hooks if you want.

Eventually, we'd love to move all the Windows protocols (S4U and PAC) into their own functions that take a single `astgs_request_t` parameter, with a view to making those plugins (even if they are ’built-in’ plugins compiled into `libkdc`).

PR #905, opened 2021-12-23
kdc: add finalize_reply API to windc plugin

lhoward:
Allow a windc plugin to finalize the KDC reply (including the encrypted ticket and reply parts) before encoding for transmission.


lhoward:
Closing, replaced by #906.

PR #906, opened 2021-12-23
Make part of astgs_request_t public

lhoward:
Encapsulate as much state as possible in `astgs_request_t`, with a view to eventually migrating the KDC to a table-driven architecture. In the interim, use a single `astgs_request_t` argument for pre- and post-hook plugins, e.g.:

```c
static krb5_error_code KRB5_CALLCONV
client_access(void *ctx, astgs_request_t r)
{
    krb5_warnx(r->context, "client_access");
    return 0;
}

static krb5_error_code KRB5_CALLCONV
finalize_reply(void *ctx, astgs_request_t r)
{
    krb5_warnx(r->context, "finalize_reply");
    return 0;
}
```

The `astgs_request_t` structure is separated into a public and private component, depending on whether we are building the KDC or not.

### DIFF
--- a/kdc/altsecid_gss_preauth_authorizer.c
+++ b/kdc/altsecid_gss_preauth_authorizer.c
@@ -372,10 +372,7 @@ out:
 
 static KRB5_LIB_CALL krb5_error_code
 authorize(void *ctx,
-          krb5_context context,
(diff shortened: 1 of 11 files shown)


nicowilliams:
Beautiful!

PR #909, opened 2021-12-29
Name Attributes

nicowilliams:
`gss_get_name_attribute()`, `gss_inquire_name()`, `gss_display_name_ext()`, `gss_export_name_composite()`, and import of `GSS_C_NT_COMPOSITE_EXPORT` name tokens is included.

Missing functionality: `gss_set_name_attribute()` and `gss_delete_name_attribute()`.  Also, no plugin system is included at this time (a plugin would allow for additional attributes / prefixes, mappings/lookups, and/or local policy).

Attributes:

 - `urn:mspac:` (same as MIT) (`XXX` missing are the URNs for the specific buffers)
 - `urn:ietf:kerberos:nameattr-...`:
    - attributes of all Kerberos names:
      - `realm` (the realm of the name)
      - `ncomp` (the number of components of the name)
      - `ncomp#<digit>` (the corresponding component of the name)
    - attributes of names from established security contexts:
    - `canonical-name` (the canonical name of the initiator, if there is one, as an exported name token and/or as an RFC1964 syntax principal for the display form)
    - `peer-realm` (the name of the peer's realm, when the name came from an established security context)
    - `transit-path` (currently only for the initiator name on the acceptor side)
    - `ticket-authz-data` (only for the initiator name on the acceptor side)
    - `authenticator-authz-data` (currently only for the initiator name on the acceptor side)
    - `authz-data#<N>` (...; either from the ticket or the authenticator)
    - `local-login-user`
 
The `authenticated` output parameter of `gss_get_name_attribute()` is set for the `realm`, `ncomp`, and `ncomp#<digit>` attributes IFF the name is a peer from an established security context.

The `authenticated` output parameter of `gss_get_name_attribute()` is set for the PAC and `AD-KDC-ISSUED` IFF their keyed checksums were valid.

The `authenticated` output parameter of `gss_get_name_attribute()` is set for `peer-realm`, `canonical-name`, and `transit-path`.

Otherwise `authenticated` is set to false.

The `complete` output parameter of `gss_get_name_attribute()` for all these attributes is set to true.

All of these attributes are single-valued.

This is ready for review.

Wishlist:

 - PAC login buffer extraction, parsing, and exposing of individual fields as attributes, especially SIDs
 - more transited path support
 - a way to indicate supported critical attributes via `gss_set_name_attribute()`
 - a way to request PAC and such
 - a way to set authz-data to send via the `Authenticator`

### DIFF
--- a/lib/gssapi/krb5/import_name.c
+++ b/lib/gssapi/krb5/import_name.c
@@ -169,9 +169,11 @@ import_export_name (OM_uint32 *minor_status,
 		    const gss_buffer_t input_name_buffer,
 		    gss_name_t *output_name)
 {
+    CompositePrincipal *composite;
     unsigned char *p;
-    uint32_t length;
+    size_t length, sz;
     OM_uint32 ret;
+    int is_composite;
     char *name;
 
     if (input_name_buffer->length < 10 + GSS_KRB5_MECHANISM->length)
@@ -181,13 +183,17 @@ import_export_name (OM_uint32 *minor_status,
 
     p = input_name_buffer->value;
 
-    if (memcmp(&p[0], "\x04\x01\x00", 3) != 0 ||
+    if (p[0] != 0x04 ||
+        (p[1] != 0x01 && p[1] != 0x02) ||
+        p[2] != 0x00 ||
 	p[3] != GSS_KRB5_MECHANISM->length + 2 ||
 	p[4] != 0x06 ||
 	p[5] != GSS_KRB5_MECHANISM->length ||
 	memcmp(&p[6], GSS_KRB5_MECHANISM->elements,
 	       GSS_KRB5_MECHANISM->length) != 0)
 	return GSS_S_BAD_NAME;
+
+    is_composite = p[1] == 0x02;
 
     p += 6 + GSS_KRB5_MECHANISM->length;
 
@@ -196,6 +202,28 @@ import_export_name (OM_uint32 *minor_status,
 
     if (length > input_name_buffer->length - 10 - GSS_KRB5_MECHANISM->length)
 	return GSS_S_BAD_NAME;
+
+    if (is_composite) {
+        if ((composite = calloc(1, sizeof(*composite))) == NULL) {
+            *minor_status = ENOMEM;
+            return GSS_S_FAILURE;
+        }
+
+        ret = decode_CompositePrincipal(p, length, composite, &sz);
+	if (ret) {
+            *minor_status = ret;
+            return GSS_S_FAILURE;
+	}
+        if (sz != length) {
+            free_CompositePrincipal(composite);
+            free(composite);
+            *minor_status = EINVAL;
+            return GSS_S_FAILURE;
+        }
+
+        *output_name = (void *)composite;
+        return GSS_S_COMPLETE;
+    }
 
     name = malloc(length + 1);
     if (name == NULL) {
@@ -207,7 +235,6 @@ import_export_name (OM_uint32 *minor_status,
 
     ret = parse_krb5_name(minor_status, context, name, output_name);
     free(name);
-
     return ret;
 }
 
@@ -239,7 +266,8 @@ OM_uint32 GSSAPI_CALLCONV _gsskrb5_import_name
 				 context,
 				 input_name_buffer,
 				 output_name);
-    else if (gss_oid_equal(input_name_type, GSS_C_NT_EXPORT_NAME)) {
+    else if (gss_oid_equal(input_name_type, GSS_C_NT_EXPORT_NAME) ||
+             gss_oid_equal(input_name_type, GSS_C_NT_COMPOSITE_EXPORT)) {
 	return import_export_name(minor_status,
 				  context,
 				  input_name_buffer,
--- a/lib/gssapi/krb5/name_attrs.c
+++ b/lib/gssapi/krb5/name_attrs.c
@@ -1,0 +1,692 @@
+/*
+ * Copyright (c) 2021 Kungliga Tekniska Högskolan
+ * (Royal Institute of Technology, Stockholm, Sweden).
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * 3. Neither the name of the Institute nor the names of its contributors
+ *    may be used to endorse or promote products derived from this software
+ *    without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE INSTITUTE AND CONTRIBUTORS ``AS IS'' AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED.  IN NO EVENT SHALL THE INSTITUTE OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+ * OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ */
+
+#include "gsskrb5_locl.h"
+
+/*
+ * (Not-yet-)Standard name attributes for Kerberos MNs,
+ * GSS_KRB5_NAME_ATTRIBUTE_BASE_URN + "...".
+ *
+ * I.e., "urn:ietf:kerberos:nameattr-...".  (XXX Register this URN namespace
+ * with IANA.)
+ *
+ * Note that we do use URN fragments.
+ *
+ * Specific attributes below the base URN:
+ *
+ *  - name access attributes:
+ *     - "realm"                    -> realm of name
+ *     - "name-ncomp"               -> count of name components
+ *     - "name-ncomp#<digit>"       -> name component N (0 <= N <= 9)
+ *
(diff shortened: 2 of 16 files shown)


lib/gssapi/krb5/import_name.c
```diff
@@ -211,6 +211,49 @@ import_export_name (OM_uint32 *minor_status,
...
+        memcmp(&p[2], "\x00\x09", 2) != 0 ||
+	memcmp(&p[4], GSS_KRB5_MECHANISM->elements,
+	       GSS_KRB5_MECHANISM->length) != 0)
+	return GSS_S_BAD_NAME;
+
+    length -= 4 + GSS_KRB5_MECHANISM->length;
+    p += 4 + GSS_KRB5_MECHANISM->length;
+
+    if ((name = calloc(1, sizeof(*name))) == NULL)
+        return ENOMEM;
```
lhoward, line 244: `*minor_status = ENOMEM; return GSS_S_FAILURE`

lib/gssapi/krb5/import_name.c
```diff
@@ -211,6 +211,49 @@ import_export_name (OM_uint32 *minor_status,
...
+
+    ret = decode_CompositePrincipal(p, length, name, &sz);
+    if (ret == 0 && sz != length) {
+        free_CompositePrincipal(name);
+        free(name);
+        return EINVAL;
```
lhoward, line 250: return `GSS_S_FAILURE`, set `*minor_status`

lib/gssapi/krb5/import_name.c
```diff
@@ -211,6 +211,49 @@ import_export_name (OM_uint32 *minor_status,
...
+    }
+
+    *output_name = (void *)name;
+    return ret;
```
lhoward, line 254: return `GSS_S_COMPLETE`, or `GSS_S_FAILURE`; set `*minor_status`

lib/gssapi/krb5/name_attrs.c
```diff
@@ -0,0 +1,691 @@
...
+        *authenticated = 0;
+    if (complete)
+        *complete = 0;
+    if (more)
+        *more = 0;
+    if (value) {
+        value->length = 0;
+        value->value = NULL;
+    }
+    if (display_value) {
```
lhoward, line 181: Can use `_mg_buffer_zero()` here

lib/gssapi/krb5/name_attrs.c
```diff
@@ -0,0 +1,691 @@
...
+    }
+
+    *minor_status = 0;
+    if (authenticated)
+        *authenticated = 0;
+    if (complete)
+        *complete = 0;
+    if (more)
+        *more = 0;
+    if (value) {
```
lhoward, line 177: Can use `_mg_buffer_zero()` here

lib/gssapi/krb5/name_attrs.c
```diff
@@ -0,0 +1,691 @@
...
+            kret = krb5_unparse_name(context, p,
+                                     (char **)&display_value->value);
+            if (kret == 0)
+                display_value->length = strlen(display_value->value);
+        }
+
+        krb5_free_principal(context, p);
+        if (kret) {
+            if (value) {
+                free(value->value);
```
lhoward, line 328: `gss_release_buffer()` with tmpMinor?

lib/gssapi/krb5/name_attrs.c
```diff
@@ -0,0 +1,691 @@
...
+            ASN1_MALLOC_ENCODE(TransitedEncoding, value->value, value->length,
+                               nameattrs->transited, &sz, kret);
+        *minor_status = kret;
+        return kret == 0 ? GSS_S_COMPLETE : GSS_S_FAILURE;
+    }
+
+    return GSS_S_UNAVAILABLE;
+
+enomem:
+    if (value) {
```
lhoward, line 510: `gss_release_buffer()`

lib/gssapi/krb5/name_attrs.c
```diff
@@ -0,0 +1,691 @@
...
+    ADD_URN("name-ncomp#8");
+    ADD_URN("name-ncomp#9");
+    ADD_URN("peer-realm");
+    ADD_URN("ticket-authz-data#pac");
+    ADD_URN("urn:mspac:");
+    ADD_URN("authenticator-authz-data"); /* XXX Add fragments? */
+    ADD_URN("ticket-authz-data"); /* XXX Add fragments? */
+    ADD_URN("authz-data");
+    ADD_URN("transit-path");
+    ADD_URN("canonical-name");
```
lhoward, line 594: could be part of table-driven approach?

lib/gssapi/krb5/name_attrs.c
```diff
@@ -0,0 +1,691 @@
...
+    *minor_status = 0;
+    if (display_name == NULL)
+        return GSS_S_CALL_INACCESSIBLE_WRITE;
+    display_name->length = 0;
+    display_name->value = NULL;
+
+    if (gss_oid_equal(display_as_name_type, GSS_C_NT_USER_NAME)) {
+        if (p->name.name_string.len != 1)
+            return GSS_S_UNAVAILABLE;
+        return _gsskrb5_localname(minor_status, name, GSS_KRB5_MECHANISM,
```
lhoward, line 623: `GSS_C_NT_USER_NAME` might be a local name in the standard, but in practice it's a synonym for `GSS_KRB5_NT_PRINCIPAL_NAME`. This wouldn't round-trip, not sure if I would advise this.

lib/gssapi/mech/gss_import_name.c
```diff
@@ -299,3 +297,71 @@ gss_import_name(OM_uint32 *minor_status,
...
+                                     &new_canonical_name);
+    if (major_status != GSS_S_COMPLETE) {
+        _gss_mg_error(m, *minor_status);
+        return major_status;
+    }
+
+    /*
+     * Now we make a new name and mark it as an MN.
+     */
+    name = _gss_create_name(new_canonical_name, m);
```
lhoward, line 357: Remind me the rationale behind factoring this out?

lib/krb5/rd_req.c
```diff
@@ -387,6 +382,27 @@ krb5_verify_ap_req2(krb5_context context,
...
+     * The ticket authenticates the client, and conveys naming attributes that
+     * we want to expose in GSS using RFC6680 APIs.
+     *
+     * So we same the ticket enc-part in the client's krb5_principal object
+     * (note though that the session key will be absent in that copy of the
+     * ticket enc-part).
+     */
+    ret = _krb5_ticket2krb5_principal(context, &t->client, &t->ticket,
+                                      ac->authenticator->authorization_data);
+    if (ret) goto out;
```
lhoward, line 395: newline

lhoward (commented):
Will have a closer look later, a few things for now :)

lib/gssapi/krb5/name_attrs.c
```diff
@@ -0,0 +1,691 @@
...
+        if (complete)
+            *complete = 1;
+
+        if (value) {
+            ASN1_MALLOC_ENCODE(AuthorizationData, value->value, value->length,
+                               nameattrs->authenticator_ad, &sz, kret);
+            *minor_status = kret;
+        }
+        return kret == 0 ? GSS_S_COMPLETE : GSS_S_FAILURE;
+    } else if (ATTR_EQ(&attr, GSS_KRB5_NAME_ATTRIBUTE_BASE_URN
```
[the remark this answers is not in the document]
nicowilliams: I'd have to have functions for all of them.  It would make this one function smaller, but that's it.

lib/gssapi/krb5/name_attrs.c
nicowilliams: Only if I also do that for the get-name-attribute side.  I don't think it would be a simplification, but maybe that's just me.  I could do it though, if you feel strongly about it.

lib/gssapi/krb5/name_attrs.c
nicowilliams: Hmm, maybe, yeah.  Round-tripping is not a hard requirement.

lib/gssapi/mech/gss_import_name.c
nicowilliams: The token format I chose to use is fairly different from the dumb one in RFC 2743.

lhoward:
Looking at [mech_eap](http://github.com/janetuk/mech_eap/), which supports naming extensions (including composite names) and is known to work with both MIT and Heimdal, composite name tokens include the token type and mechanism OID prefix, and nothing specific is expected of the mechglue. `gss_export_name_composite()` calls straight through to the mechanism, and `gss_import_name(GSS_C_NT_COMPOSITE_EXPORT)` does validate the prefix to find the correct mechanism, but otherwise passes the entire token to the mechanism.

nicowilliams:
Er, there was no support for importing composite name tokens before this because `lib/gssapi/mech/gss_import_name.c`:`gss_import_name()` was only checking for `GSS_C_NT_EXPORT_NAME`, which means I can in fact change the format to my liking.  That said, maybe you have `mech_eap`-using apps that were importing exported composite name tokens using `GSS_C_NT_EXPORT_NAME`?

Also, I don't understand this bit:

```
110         if (len < t + 4)
111                 return (GSS_S_BAD_NAME);
112         p += t;
113         len -= t;
114
115         t = (p[0] << 24) | (p[1] << 16) | (p[2] << 8) | p[3];
116         p += 4;
117         len -= 4;
118
119         if (!composite && len != t)
120                 return (GSS_S_BAD_NAME);
```

What's going on with line `119`?

nicowilliams:
I've fixed all the things except the table driven attribute thing.  Working on that next.

lib/gssapi/krb5/name_attrs.c
```diff
@@ -0,0 +1,691 @@
...
+        /*
+         * Output the peer's realm.  The value and display value are the
+         * same in this case.
+         */
+        if (authenticated)
+            *authenticated = 1;
+        if (complete)
+            *complete = 1;
+        if (value) {
+            if ((value->value = strdup(nameattrs->peer_realm[0])) == NULL)
```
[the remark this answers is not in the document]
nicowilliams: I'll skip this one for now.

lib/gssapi/krb5/import_name.c
nicowilliams: Fixed.

lib/gssapi/krb5/name_attrs.c
nicowilliams: I'll keep this for now.  We're free to change this any time we want.

lib/krb5/rd_req.c
nicowilliams: Actually, this is kinda the style in this file.  I'll keep this on one line.

nicowilliams:
The `mech_eap` thing...  There is something of an ABI change, which is that before if you called `gss_import_name()` with `GSS_C_NT_EXPORT_NAME` as the name type but the name token was a composite name token, then we accepted that and called the mechanism's import function with `GSS_C_NT_EXPORT_NAME`, but now we call it with `GSS_C_NT_COMPOSITE_EXPORT` instead.

lhoward:
> The `mech_eap` thing... There is something of an ABI change, which is that before if you called `gss_import_name()` with `GSS_C_NT_EXPORT_NAME` as the name type but the name token was a composite name token, then we accepted that and called the mechanism's import function with `GSS_C_NT_EXPORT_NAME`, but now we call it with `GSS_C_NT_COMPOSITE_EXPORT` instead.
> 
> I may further fix that to match the previous behavior...

mech_eap will import composite names with either name type. What it won't do is import a non-composite name with `GSS_C_NT_COMPOSITE_EXPORT`.

PR #910, opened 2021-12-30
add attribute dictionary to kdc_request_t

lhoward:
Add a heim_dict_t to the KDC request structure for use by pre-authentication mechanisms and plugins.


nicowilliams:
LGTM, except that we're not declaring calling conventions for functions exported by libkdc, but we should.

lhoward:
Will fix before merge.

PR #912, opened 2021-12-31
Harmonize auditing APIs

lhoward:


nicowilliams:
Well, the kv thing is designed to yield human-readable, machine-parseable log messages.  It's very particular, and the use is very particular.

The HDB audit thing for Samba is really only after state to feed back to the HDB for locking out users.  It could be made simple, and it could be made to use the kv framework.  All it needs to know is whether an AS-REQ succeeded or not, and if not, why.  The reason for failure could be varied, so the details matter, but it's mostly about "password incorrect".  Is there anything else it needs?

Also, the problem with the kv stuff is that we vis everything immediately upon adding a kv.  The vis stuff can be unvis'ed, though idk how round trippy it is, but if it doesn't round-trip then one thing we could do is delay the vis work until heim_audit_trail(), then we can definitely use the raw kv at that point.

In fact, we could even allow values of all heimbase/CF types, including non-JSON, and then we could encode for logging only the ones that can be formatted as text.

lhoward:
> The kv thing is designed to yield human-readable, machine-parseable log messages. It's very particular, and the use is very particular. Can we generalize it? I think so.

Agreed (see current patch).

> The HDB audit thing for Samba is really only after state to feed back to the HDB for locking out users. It could be made simple, and it could be made to use the kv framework. All it needs to know is whether an AS-REQ succeeded or not, and if not, why. The reason for failure could be varied, so the details matter, but it's mostly about "password incorrect". Is there anything else it needs? But in principle it (or something else) could need a lot of information.

Well, this is a good question, see latest comments in #888. Looking at the Samba code, I think it really only cares about long-term key pre-authentication method decrypt failures (i.e. password failures) and successes, to manage lockouts. I think, with their cooperation (cc @abartlet @josephsutton1) we could eliminate most of the authentication event/status enumeration, particularly now the final error code is available to the audit method as `HDB_REQUEST_KV_ERROR_CODE`.

> Also, the problem with the kv stuff is that we `vis` everything immediately upon adding a `kv`. The vis stuff can be `unvis`'ed, though idk how round trippy it is, but if it doesn't round-trip then one thing we could do is delay the `vis` work until `heim_audit_trail()`, then we can definitely use the raw `kv` at that point.

Doing this accurately would require retaining the `vis` flags, right? Which is possible (could be a hidden key) but it's a bit ugly. I suspect this work can be delayed: the audit points I replaced to take numbers and booleans never set the `vis` flags, and indeed `KDC_AUDIT_VIS` is only set once in `kerberos5.c` and never in `krb5tgs.c`.

> In fact, we could even allow values of all heimbase/CF types, including non-JSON, and then we could encode for logging only the ones that can be formatted as text.

This is pretty much what I ended up doing: numbers, booleans, strings, the NULL type can be encoded for logging. Of course, there might be some circumstances where you want to hide a printable type from logging, and that is currently done by prefixing the key with a '#' symbol.

> As well, the `kv`s are not indexed -- it's just an array. Maybe it should be represented as a dict with array values. This would make "querying" the `kv` store possible (though a linear search of the array is probably ok given how few keys there are, we do have `heim_dict_t`). This would then allow us to have an API consisting of key name & value type and format conventions that Samba (and anything else) could use. Also, this would allow us to have plugins for auditing and for logging in somewhat different formats (e.g., we could have a JSON log).

Right, the patch changes `kv` from an array tot a dictionary.

lib/base/log.c
```diff
@@ -808,14 +816,142 @@ heim_audit_addkv_timediff(heim_svc_req_desc r, const char *k,
...
+
+    if (kvb->pos < sizeof(kvb->buf) - 1)
+	kvb->buf[kvb->pos++] = ' ';
+    for (; *k && kvb->pos < sizeof(kvb->buf) - 1; kvb->pos++)
+	kvb->buf[kvb->pos] = *k++;
+    if (kvb->pos < sizeof(kvb->buf) - 1)
+	kvb->buf[kvb->pos++] = '=';
+    for (; *v && kvb->pos < sizeof(kvb->buf) - 1; kvb->pos++)
+	kvb->buf[kvb->pos] = *v++;
+}
```
lhoward, line 947: @nicowilliams please check for any off-by-one errors etc

lib/base/log.c
```diff
@@ -838,26 +974,15 @@ heim_audit_trail(heim_svc_req_desc r, heim_error_code ret, const char *retname)
...
+    heim_dict_iterate_f(r->kv, &kvb, audit_trail_iterator);
+    kvb.buf[kvb.pos] = '\0';
 
+    /* FIXME: do we actually want to log the reason prefix? */
     heim_log(r->hcontext, r->logf, 3, "%s %s %s %s %s%s%s%s",
              r->reqtype, retval, r->from,
              r->cname ? r->cname : "<unknown>",
              r->sname ? r->sname : "<unknown>",
-             kvbuf, r->reason ? " " : "",
+             kvb.buf, r->reason ? " reason=" : "",
```
lhoward, line 986: as far as I could tell, the `reason` would have had a `reason=` prefix by virtue of `heim_audit_vaddreason()` calling `fmtkv()`. This doesn't really strike me as aesthetically desirable but I didn't want to change the log format.

nicowilliams:
Does this depend on #910?

lhoward:
> Does this depend on #910?

No.

nicowilliams:
> > Also, the problem with the kv stuff is that we `vis` everything immediately upon adding a `kv`. The vis stuff can be `unvis`'ed, though idk how round trippy it is, but if it doesn't round-trip then one thing we could do is delay the `vis` work until `heim_audit_trail()`, then we can definitely use the raw `kv` at that point.
> 
> Doing this accurately would require retaining the `vis` flags, right? Which is possible (could be a hidden key) but it's a bit ugly. I suspect this work can be delayed: the audit points I replaced to take numbers and booleans never set the `vis` flags, and indeed `KDC_AUDIT_VIS` is only set once in `kerberos5.c` and never in `krb5tgs.c`.

Lets not have a kv for the VIS flags.  You just have to know for each kv what VIS flags to use.

I see you're not delaying the VIS work.

> > In fact, we could even allow values of all heimbase/CF types, including non-JSON, and then we could encode for logging only the ones that can be formatted as text.
> 
> This is pretty much what I ended up doing: numbers, booleans, strings, the NULL type can be encoded for logging. Of course, there might be some circumstances where you want to hide a printable type from logging, and that is currently done by prefixing the key with a '#' symbol.

+1

> > As well, the `kv`s are not indexed -- it's just an array. Maybe it should be represented as a dict with array values. This would make "querying" the `kv` store possible (though a linear search of the array is probably ok given how few keys there are, we do have `heim_dict_t`). This would then allow us to have an API consisting of key name & value type and format conventions that Samba (and anything else) could use. Also, this would allow us to have plugins for auditing and for logging in somewhat different formats (e.g., we could have a JSON log).
> 
> Right, the patch changes `kv` from an array tot a dictionary.

+1

nicowilliams:
I think we can get rid of `KDC_AUDIT_EATWHITE` by not formatting spaces for pre-auth and etype lists, and KDC options flags.

IDK what `LAST` in `HEIM_SVC_AUDIT_VISLAST` means, but `HEIM_SVC_AUDIT_VISLAST` == `VIS_CSTYLE | VIS_OCTAL | VIS_NL` and `HEIM_SVC_AUDIT_VIS` == `VIS_CSTYLE | VIS_OCTAL | VIS_NL VIS_WHITE` == `VIS_CSTYLE | VIS_OCTAL | VIS_NL  | VIS_SP | VIS_TAB`.

We don't use `HEIM_SVC_AUDIT_VISLAST`, so we might as well remove it.

That leaves use with a few cases where we use `0` but the value has nothing to visualize, so... I think we can just get rid of all the `AUDIT` flags altogether and then just visualize in the final step.

nicowilliams:
OK, yes, I think we can move the VIS work to `heim_audit_trail()`/`audit_trail_iterator()`.

lib/hdb/hdb.h
```diff
@@ -79,78 +81,50 @@ enum hdb_lockop{ HDB_RLOCK, HDB_WLOCK };
...
+ * intended for logging that are not defined below; the constants below are
+ * there to ease migration from the older auth_status HDB API.
  */
-#define HDB_AUTHSTATUS_GENERIC_SUCCESS	        6
 
-/*
- * Failed authentication with a pre-authentication mechanism
- */
-#define HDB_AUTHSTATUS_GENERIC_FAILURE	        7
+#define HDB_REQUEST_KV_AUTH_EVENT		"#auth_event"		/* heim_number_t */
```
nicowilliams, line 111: `#`?

lhoward: Hides it from printing in logs

nicowilliams:
Anyways, LGTM, though I think we should move the vis stuff.

PR #914, opened 2022-01-01
gss: pass GSS_C_NO_OID name type to mechanism

lhoward:


lhoward:
Going to close this for now, because the tests (and thus possibly applications), depend on `GSS_C_NO_OID` importing as `GSS_C_NT_USER_NAME`.

PR #916, opened 2022-01-02
kdc: add audit plugin API to windc API

lhoward:
Allow the windc plugin to also implement an allback callback. As part of this change, both the HDB and windc audit function signatures are changed to return void.


lhoward:
Closing; will instead merge to new PR which contains windc to kdc plugin rename.

PR #917, opened 2022-01-02
kdc: add referral_policy callback to windc plugin

lhoward:
Add a referral policy hook to the TGS as a more elegant way of resolving PR detection for Samba).


lhoward:
Closing; will instead merge to new PR which contains windc to kdc plugin rename.

PR #919, opened 2022-01-03
complete support for DB6

jaltman:
The prior inclusion of DB6 checks within the tree was incomplete.   The db6/db.h header was tested for and used in some places but the test for db_create() within libdb-6 if present did not exist.   

The roken ndbm-wrap functionality was also unaware of DB6.


lhoward:
LGTM

jaltman:
7fb740adb..3632158aa  HEAD -> heimdal-7-1-branch

PR #920, opened 2022-01-05
kdc: add rewrite_request plugin API

lhoward:
Add a plugin API that allows a plugin to rewrite the request body.

Edit: not sure if this is actually useful.


lhoward:
Closing for now.

PR #922, opened 2022-01-06
Use ASN.1 type decoration for profit in HDB and for PACs

nicowilliams:
PREVIOUSLY:

> This adds support for asn1_compile --decorate=... variation that causes decoration of an ASN.1 SET/SEQUENCE type with a field of a non-ASN.1 type.
> 
> This means we can now have an ASN.1 type to represent a request that can then have a "hidden" field -- hidden in that it is neither encoded nor decoded.  This field will be copied and freed when the decoration is of an ASN.1 type or of a external, C type that comes with copy constructor and destructor functions.  Decoration with a `void *` field which is neither copied nor freed is also supported.

That was reviewed by @lhoward.

NOW:

> Follow-up commits add a `krb5_pac` to `PrincipalNameAttrs` and also replaces `hdb_entry_ex` by decorating `HDB_entry` with a C type that points to the backend-specific context.

### DIFF
--- a/kadmin/load.c
+++ b/kadmin/load.c
@@ -519,7 +519,7 @@ doit(const char *filename, int mergep)
 	if (parse_keys(&ent.entry, e.key)) {
 	    fprintf (stderr, "%s:%d:error parsing keys (%s)\n",
 		     filename, lineno, e.key);
-	    hdb_free_entry (context, &ent);
+	    hdb_free_entry (context, db, &ent);
             ret = 1;
 	    continue;
 	}
@@ -527,35 +527,35 @@ doit(const char *filename, int mergep)
 	if (parse_event(&ent.entry.created_by, e.created) == -1) {
 	    fprintf (stderr, "%s:%d:error parsing created event (%s)\n",
 		     filename, lineno, e.created);
-	    hdb_free_entry (context, &ent);
+	    hdb_free_entry (context, db, &ent);
             ret = 1;
 	    continue;
 	}
 	if (parse_event_alloc (&ent.entry.modified_by, e.modified) == -1) {
 	    fprintf (stderr, "%s:%d:error parsing event (%s)\n",
 		     filename, lineno, e.modified);
-	    hdb_free_entry (context, &ent);
+	    hdb_free_entry (context, db, &ent);
             ret = 1;
 	    continue;
 	}
 	if (parse_time_string_alloc (&ent.entry.valid_start, e.valid_start) == -1) {
 	    fprintf (stderr, "%s:%d:error parsing time (%s)\n",
 		     filename, lineno, e.valid_start);
-	    hdb_free_entry (context, &ent);
+	    hdb_free_entry (context, db, &ent);
             ret = 1;
 	    continue;
 	}
 	if (parse_time_string_alloc (&ent.entry.valid_end,   e.valid_end) == -1) {
 	    fprintf (stderr, "%s:%d:error parsing time (%s)\n",
 		     filename, lineno, e.valid_end);
-	    hdb_free_entry (context, &ent);
+	    hdb_free_entry (context, db, &ent);
             ret = 1;
 	    continue;
 	}
 	if (parse_time_string_alloc (&ent.entry.pw_end,      e.pw_end) == -1) {
 	    fprintf (stderr, "%s:%d:error parsing time (%s)\n",
 		     filename, lineno, e.pw_end);
-	    hdb_free_entry (context, &ent);
+	    hdb_free_entry (context, db, &ent);
             ret = 1;
 	    continue;
 	}
@@ -563,7 +563,7 @@ doit(const char *filename, int mergep)
 	if (parse_integer_alloc (&ent.entry.max_life,  e.max_life) == -1) {
 	    fprintf (stderr, "%s:%d:error parsing lifetime (%s)\n",
 		     filename, lineno, e.max_life);
-	    hdb_free_entry (context, &ent);
+	    hdb_free_entry (context, db, &ent);
             ret = 1;
 	    continue;
 
@@ -571,7 +571,7 @@ doit(const char *filename, int mergep)
 	if (parse_integer_alloc (&ent.entry.max_renew, e.max_renew) == -1) {
 	    fprintf (stderr, "%s:%d:error parsing lifetime (%s)\n",
 		     filename, lineno, e.max_renew);
-	    hdb_free_entry (context, &ent);
+	    hdb_free_entry (context, db, &ent);
             ret = 1;
 	    continue;
 	}
@@ -579,7 +579,7 @@ doit(const char *filename, int mergep)
 	if (parse_hdbflags2int (&ent.entry.flags, e.flags) != 1) {
 	    fprintf (stderr, "%s:%d:error parsing flags (%s)\n",
 		     filename, lineno, e.flags);
-	    hdb_free_entry (context, &ent);
+	    hdb_free_entry (context, db, &ent);
             ret = 1;
 	    continue;
 	}
@@ -587,7 +587,7 @@ doit(const char *filename, int mergep)
 	if(parse_generation(e.generation, &ent.entry.generation) == -1) {
 	    fprintf (stderr, "%s:%d:error parsing generation (%s)\n",
 		     filename, lineno, e.generation);
-	    hdb_free_entry (context, &ent);
+	    hdb_free_entry (context, db, &ent);
             ret = 1;
 	    continue;
 	}
@@ -595,13 +595,13 @@ doit(const char *filename, int mergep)
 	if (parse_extensions(&e.extensions, &ent.entry.extensions) == -1) {
 	    fprintf (stderr, "%s:%d:error parsing extension (%s)\n",
 		     filename, lineno, e.extensions);
-	    hdb_free_entry (context, &ent);
+	    hdb_free_entry (context, db, &ent);
             ret = 1;
 	    continue;
 	}
 
 	ret2 = db->hdb_store(context, db, HDB_F_REPLACE, &ent);
-	hdb_free_entry (context, &ent);
+	hdb_free_entry (context, db, &ent);
 	if (ret2) {
 	    krb5_warn(context, ret2, "db_store");
 	    break;
--- a/kadmin/load.c
+++ b/kadmin/load.c
@@ -418,7 +418,7 @@ doit(const char *filename, int mergep)
     int lineno;
     int flags = O_RDWR;
     struct entry e;
-    hdb_entry_ex ent;
+    hdb_entry ent;
     HDB *db = _kadm5_s_get_db(kadm_handle);
 
     f = fopen(filename, "r");
@@ -506,7 +506,7 @@ doit(const char *filename, int mergep)
 	skip_next(p);
 
 	memset(&ent, 0, sizeof(ent));
-	ret2 = krb5_parse_name(context, e.principal, &ent.entry.principal);
+	ret2 = krb5_parse_name(context, e.principal, &ent.principal);
 	if (ret2) {
 	    const char *msg = krb5_get_error_message(context, ret);
 	    fprintf(stderr, "%s:%d:%s (%s)\n",
@@ -516,7 +516,7 @@ doit(const char *filename, int mergep)
 	    continue;
 	}
 
-	if (parse_keys(&ent.entry, e.key)) {
+	if (parse_keys(&ent, e.key)) {
 	    fprintf (stderr, "%s:%d:error parsing keys (%s)\n",
 		     filename, lineno, e.key);
 	    hdb_free_entry (context, db, &ent);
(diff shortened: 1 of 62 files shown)


lib/asn1/Makefile.am
```diff
@@ -23,6 +23,7 @@ endif
 libasn1_la_LIBADD = \
 	libasn1base.la \
 	@LIB_com_err@ \
+	$(LIB_heimbase) \
```
nicowilliams, line 26: Oops, this needs to be separate now.

nicowilliams:
@lhoward this makes the PAC decoration of name attributes a mutable shared object, does it not?  does that present problems for threading?

lhoward:
> @lhoward this makes the PAC decoration of name attributes a mutable shared object, does it not? does that present problems for threading?

Practically not because only the KDC mutates the PAC, but in theory yes. That's why I suggested implementing `heim_copy()`.

lhoward:
Windows build is failing again.

nicowilliams:
> > @lhoward this makes the PAC decoration of name attributes a mutable shared object, does it not? does that present problems for threading?
> 
> Practically not because only the KDC mutates the PAC, but in theory yes. That's why I suggested implementing `heim_copy()`.

This is why I'd like us to move `heimbase` to something more like the immutable data structure approach of `libjq`'s `jv` APIs.  That would probably be a lot of work.

For now this is probably OK because, indeed, nothing mutates the PAC or needs to besides the KDC, and the KDC doesn't and won't use this API, not yet, and not in a threaded way anyways, so we're ok.

lhoward:
> > > @lhoward this makes the PAC decoration of name attributes a mutable shared object, does it not? does that present problems for threading?
> > 
> > 
> > Practically not because only the KDC mutates the PAC, but in theory yes. That's why I suggested implementing `heim_copy()`.
> 
> This is why I'd like us to move `heimbase` to something more like the immutable data structure approach of `libjq`'s `jv` APIs. That would probably be a lot of work.
> 
> For now this is probably OK because, indeed, nothing mutates the PAC or needs to besides the KDC, and the KDC doesn't and won't use this API, not yet, and not in a threaded way anyways, so we're ok.

The KDC calls `krb5_verify_ap_req2()` which does not verify the PAC. Instead, it does its own PAC verification (which makes sense, it has to be able to verify the KDC signatures). So there aren't any issues here.

I'm still getting my head around the `jv` API where reference parameters are consumed by their callee, something doesn't feel quite right about this from an API perspective, but I can see how it helps eliminate copies with uniquely referenced objects.

nicowilliams:
> I'm still getting my head around the `jv` API where reference parameters are consumed by their callee, something doesn't feel quite right about this from an API perspective, but I can see how it helps eliminate copies with uniquely referenced objects.

The rule in the `jv` API is that all `jv` arguments have reference consumed except for `jv_copy()`'s and `jv_get_kind()`'s `jv` argument.  This makes sense except for a few functions:

- `jv_string_length_bytes()`,
- `jv_array_length()`,
- and `jv_invalid_has_msg()`

for those three (and maybe one or two more, I don't remember) it was a mistake to have them consume a reference because invariably one wants to use the `jv` right after and so one always calls, for example, `jv_invalid_has_msg(jv_copy(v))`.

But the upshot is that as long as the refcount incref/decref is atomic, then you get a naturally-thread-safe API because automatically no shared state is ever mutable.  It's also very easy to reason about things.

On the down side, memory leaks are much harder to debug because tools like `valgrind` know nothing about `jv_copy()` and generally have no way to be taught about it, so you see a stack trace where a leaked object was allocated, but you've no idea where the `jv_copy()` that you're actually leaking was taken and then you have to spend some time reasoning about the code because the leak is likely not obvious staring at the original allocation point.  I feel this is worth the safety.

nicowilliams:
@abartlet how disruptive will this change to Heimdal be?

lhoward:
> I haven't kept fully up to date with all of the Heimdal changes so far, but I don't think this change will be too disruptive to Samba. I think we'll just need to keep a pointer to the HDB database somewhere (maybe in the `hdb_entry->context` member) for some spots that need to call `hdb_free_entry()`.

You could do that, or just pass the HDB parameter down from the caller as all HDB backend functions take one.

josephsutton1:
True, although we also call `hdb_free_entry()` in the `pac_verify()` plugin function, which doesn't seem to have an HDB parameter.

jaltman:
> True, although we also call `hdb_free_entry()` in the `pac_verify()` plugin function, which doesn't seem to have an HDB parameter.

What is being freed from pac_verify()?

lhoward:
> True, although we also call `hdb_free_entry()` in the `pac_verify()` plugin function, which doesn't seem to have an HDB parameter.

Ah, good point. Well, for consistency, all the `kdc-plugin.h` APIs that take a `hdb_entry` should also take a `HDB` argument.

josephsutton1:
> > True, although we also call `hdb_free_entry()` in the `pac_verify()` plugin function, which doesn't seem to have an HDB parameter.
> 
> What is being freed from pac_verify()?

If an RWDC receives a service ticket that was signed by an RODC, then we fetch the RODC entry from the database to verify the signature, and the entry gets freed afterwards.

jaltman:
The signature of hdb_free_entry() doesn't take a HDB.   It takes a krb5_context and the hdb_entry_ex.
`void hdb_free_entry(krb5_context context, hdb_entry_ex *ent)`

lhoward:
> > > True, although we also call `hdb_free_entry()` in the `pac_verify()` plugin function, which doesn't seem to have an HDB parameter.
> > 
> > 
> > What is being freed from pac_verify()?
> 
> If an RWDC receives a service ticket that was signed by an RODC, then we fetch the RODC entry from the database to verify the signature, and the entry gets freed afterwards.

Ah, OK, in this case we'll just amend the `kdc-plugin` prototypes that don't take a `astgs_request_t` to include a `krb5_kdc_configuration *` argument, and you can get to your backend via `kdc_config->db[0]` as you do in other parts of your code.

josephsutton1:
> The signature of hdb_free_entry() doesn't take a HDB. It takes a krb5_context and the hdb_entry_ex. `void hdb_free_entry(krb5_context context, hdb_entry_ex *ent)`

That's the case in master, but this PR changes the signature to:
`void hdb_free_entry(krb5_context context, HDB *db, hdb_entry *ent)`

> Ah, OK, in this case we'll just amend the `kdc-plugin` prototypes that don't take a `astgs_request_t` to include a `krb5_kdc_configuration *` argument, and you can get to your backend via `kdc_config->db[0]` as you do in other parts of your code.

Thanks!

lhoward:
> > The signature of hdb_free_entry() doesn't take a HDB. It takes a krb5_context and the hdb_entry_ex. `void hdb_free_entry(krb5_context context, hdb_entry_ex *ent)`
> 
> That's the case in master, but this PR changes the signature to: `void hdb_free_entry(krb5_context context, HDB *db, hdb_entry *ent)`

Correct :)
 
> > Ah, OK, in this case we'll just amend the `kdc-plugin` prototypes that don't take a `astgs_request_t` to include a `krb5_kdc_configuration *` argument, and you can get to your backend via `kdc_config->db[0]` as you do in other parts of your code.
> 
> Thanks!

Done!

lhoward (approved):

lhoward:
> > I'm still getting my head around the `jv` API where reference parameters are consumed by their callee, something doesn't feel quite right about this from an API perspective, but I can see how it helps eliminate copies with uniquely referenced objects.
> 
> The rule in the `jv` API is that all `jv` arguments have reference consumed except for `jv_copy()`'s and `jv_get_kind()`'s `jv` argument. This makes sense except for a few functions:
> 
> * `jv_string_length_bytes()`,
> * `jv_array_length()`,
> * and `jv_invalid_has_msg()`
> 
> for those three (and maybe one or two more, I don't remember) it was a mistake to have them consume a reference because invariably one wants to use the `jv` right after and so one always calls, for example, `jv_invalid_has_msg(jv_copy(v))`.
> 
> But the upshot is that as long as the refcount incref/decref is atomic, then you get a naturally-thread-safe API because automatically no shared state is ever mutable. It's also very easy to reason about things.

That is very clever, agreed.

But as a contrary point, should we consider our use cases? It seems to me that the consumers of `heimbase` are either single-threaded already (like the KDC), or require the caller to guarantee thread-safety (e.g. `libkrb5` or `libgssapi`).

nicowilliams:
> Ah, OK, in this case we'll just amend the `kdc-plugin` prototypes that don't take a `astgs_request_t` to include a `krb5_kdc_configuration *` argument, and you can get to your backend via `kdc_config->db[0]` as you do in other parts of your code.

Does that impose undue stability on `krb5_kdc_configuration`?

lhoward:
> > Ah, OK, in this case we'll just amend the `kdc-plugin` prototypes that don't take a `astgs_request_t` to include a `krb5_kdc_configuration *` argument, and you can get to your backend via `kdc_config->db[0]` as you do in other parts of your code.
> 
> Does that impose undue stability on `krb5_kdc_configuration`?

It's part of the public API in`kdc.h`. It's also part of `astgs_request_t`. So, no.

nicowilliams:
> That is very clever, agreed.

Stephen Dolan is clever, indeed.

> But as a contrary point, should we consider our use cases? It seems to me that the consumers of `heimbase` are either single-threaded already (like the KDC), or require the caller to guarantee thread-safety (e.g. `libkrb5`).

Right, this is quite true.  My hope has been that if we were to move to `libjq` for these things then we could get a decent JSON parser/printer, as well as the ability to code some things in `jq`, like SPF for `capaths`.  Maybe I should just quit whining and make `lib/base`'s JSON support not suck.

>     * mostly, I guess there are cases where we do lock: appears to be GSSAPI and mcache.

Right.  Even in GSS we should mostly not need locks except for wrap/unwrap and get-MIC/verify-MIC, and then only because some sequence number window updates can't easily be implemented w/o locks.

lhoward:
> > > Ah, OK, in this case we'll just amend the `kdc-plugin` prototypes that don't take a `astgs_request_t` to include a `krb5_kdc_configuration *` argument, and you can get to your backend via `kdc_config->db[0]` as you do in other parts of your code.
> > 
> > 
> > Does that impose undue stability on `krb5_kdc_configuration`?
> 
> It's part of the public API in`kdc.h`. It's also part of `astgs_request_t`. So, no.

I mean, if you think it should be private, then it needs to be moved from `kdc.h` to `kdc_locl.h`, or perhaps it has a public and private component as does `astgs_request_t`. But that should be a separate PR obviously.

lhoward:
> > That is very clever, agreed.
> 
> Stephen Dolan is clever, indeed.

I agree. It does come at the cost of what is a non-intuitive API (but, given `libheimbase` consumers are all internal, not really an issue).

I know it's probably in the `jq` link you sent but, what happens for the equivalent of `heim_dict_get_value()`? If it is a base type like a string or a number (which is immutable by design, even in `libheimbase` today), then it can increment the reference count and return. But, what if the value is a collection type? It will need to do a (deep) copy, rihgt?
 
> Right, this is quite true. My hope has been that if we were to move to `libjq` for these things then we could get a decent JSON parser/printer, as well as the ability to code some things in `jq`, like SPF for `capaths`. Maybe I should just quit whining and make `lib/base`'s JSON support not suck.

Maybe. I guess, that's the point of this design phase :)

> > * mostly, I guess there are cases where we do lock: appears to be GSSAPI and mcache.
> > ```
> 
> Right. Even in GSS we should mostly not need locks except for wrap/unwrap and get-MIC/verify-MIC, and then only because some sequence number window updates can't easily be implemented w/o locks.

Yes, I wonder if we could remove all the locks (assuming caller serialises context establishment and deletion; does the spec say anything about that? could be an issue for applications like DCE RPC where requests can come in on different threads), and just use atomics for sequence management.

nicowilliams:
> I mean, if you think it should be private, then it needs to be moved from `kdc.h` to `kdc_locl.h`, or perhaps it has a public and private component as does `astgs_request_t`. But that should be a separate PR obviously.

In the past I've thought it was private, so I've made changes in the middle of it.  Can we haz a comment in it so no one forgets?

lhoward:
> > I mean, if you think it should be private, then it needs to be moved from `kdc.h` to `kdc_locl.h`, or perhaps it has a public and private component as does `astgs_request_t`. But that should be a separate PR obviously.
> 
> In the past I've thought it was private, so I've made changes in the middle of it. Can we haz a comment in it so no one forgets?

We could split into public/private parts, the question is what parts needs to be public. I don't really have any idea.  Clearly

{code}
    struct HDB **db;
    int num_db;
{code}

apropos of the discussion in this thread, but maybe `app` and `logf` too.

Also there are a gazillion `krb5_boolean` and `int` boolean value, should we make those `unsigned int:1`.

nicowilliams:
True, we could put the things we're willing to commit to at the top.

lhoward:
> True, we could put the things we're willing to commit to at the top.

See #928

jaltman:
In general I believe structures such as astgs_request_t should be opaque private and we should provide thread safe accessors to access the internals.   Then the data structure can change over time.    

As for freeing the hdb_entry I think the entry should contain a pointer to the HDB that allocated it and then the entry can free itself.

lhoward:
> In general I believe structures such as astgs_request_t should be opaque private and we should provide thread safe accessors to access the internals. Then the data structure can change over time.

Note: the KDC is presently single-threaded.

lhoward:
BTW, we already have the `_kdc_audit_getkv()` accessor for audit plugins.

PR #926, opened 2022-01-10
Set TGS error sname to request sname

lhoward:


lhoward:
Possibly we should honor force_canonicalize on the server entry.

nicowilliams:
LGTM

PR #927, opened 2022-01-11
Fix warnings

nicowilliams:


kadmin/del.c
```diff
@@ -64,9 +64,8 @@ do_del_ns_entry(krb5_principal nsp, void *data)
     char *unsp = NULL;
 
     if (krb5_principal_get_num_comp(context, nsp) != 2) {
-        (void) krb5_unparse_name(context, nsp, &unsp);
-        krb5_warn(context, ret = EINVAL, "Not a valid namespace name %s",
-                   unsp ? unsp : "<Out of memory>");
+        ret = krb5_unparse_name(context, nsp, &unsp);
```
lhoward, line 67: This change looks bogus, the old code was better. `unsp` isn't being used and the error is the result of `krb5_unparse_name()`, it should be `EINVAL`.

lhoward: Or, at least, is the `krb5_unparse_name()` call required?

lib/gssapi/spnego/negoex_util.c
```diff
@@ -86,10 +86,13 @@ _gss_negoex_begin(OM_uint32 *minor, gssspnego_ctx ctx)
...
-    struct negoex_auth_mech *mech, *next;
+    struct negoex_auth_mech *mech;
 
-    HEIM_TAILQ_FOREACH_SAFE(mech, &ctx->negoex_mechs, links, next) {
-	_gss_negoex_release_auth_mech(context, mech);
+    while (!HEIM_TAILQ_EMPTY(&ctx->negoex_mechs)) {
+        mech = HEIM_TAILQ_FIRST(&ctx->negoex_mechs);
+        HEIM_TAILQ_REMOVE(&ctx->negoex_mechs, mech, links);
+        _gss_negoex_release_auth_mech(context, mech);
+        mech = NULL;
```
lhoward, line 95: Is there any point setting `mech` to NULL? Is there a way to implement a safe version of the macro rather than having additional code at the call site?

nicowilliams: Yes, I am going to spend some time thinking about this particular pattern.  There has to be a better way.

lhoward:
Appears some builds are failing with unknown option `-Wno-unused-but-set-variable`.

lhoward:
And ` -Wenum-conversion`.

nicowilliams:
This has mostly been merged one by one by now.

PR #928, opened 2022-01-13
Make most of KDC configuration private

lhoward:
Only expose a small part of `krb5_kdc_configuration` outside KDC.


lhoward:
Closing this, let's use accessors instead and make everything private.

PR #929, opened 2022-01-13
kdc: move auth event definitions into KDC header

lhoward:
Move KDC auth event macro definitions out of `hdb.h` and into a new KDC header, `kdc-audit.h`.


jaltman:
Looks good

lhoward:
Thanks @jaltman. This is with a view towards possibly removing the audit API entirely from `hdb.h` (Samba will have a KDC plugin anyway, so it doesn't affect them; the only reason to leave it there is perhaps if we want to use it someday for `kadmin` too).

PR #933, opened 2022-01-14
KDC data structure accessors for use by plugins

lhoward:


lhoward:
Closing, trying alternative approach.

PR #934, opened 2022-01-14
kdc: make auditing API public

lhoward:
Samba plugins will need to use auditing API without including krb5-private.h, so make the auditing APIs public.

Essentially this patch is `s/_kdc_audit/kdc_audit/` except for `_kdc_audit_trail`.


jaltman (approved):

lhoward:
Rebased

PR #937, opened 2022-01-16
gssapi/mech: mech_locl.h roken.h must be included earlier

jaltman:
If included roken.h should be immediately following config.h.
Doing so ensures that all platform specific headers are
included in the proper order and avoids unnecessary includes
of headers managed by roken.h.


jaltman:
All builds completed successfully through the test stage.   Failures in "distclean".

PR #938, opened 2022-01-16
remove krb5 enctype casts

jaltman:
Avoid unnecessary casting to krb5_enctype by use of CPP macros instead of anonymous enums for alternative enctype names.


jaltman:
distclean failures

PR #939, opened 2022-01-16
lib/hcrypto: ENGINE_by_dso do not leak handle

jaltman:
Must dlclose(handle) before returning even if 'engine' is
returned to caller.


jaltman:
the osx failure is unrelated

PR #943, opened 2022-01-19
kdc: add accessor functions for KDC request structure

lhoward:


lhoward:
Note to self: renaming `ret` to `error_code` is bad, or should at least be in a separate commit. Accessor names should possibly be customisable to not embed the field name in the symbol name.

kdc/Makefile.am
```diff
@@ -244,7 +244,7 @@ include_HEADERS = kdc.h $(srcdir)/kdc-protos.h
 noinst_HEADERS = $(srcdir)/kdc-private.h
 
 krb5dir = $(includedir)/krb5
-krb5_HEADERS = kdc-audit.h kdc-plugin.h token_validator_plugin.h csr_authorizer_plugin.h gss_preauth_authorizer_plugin.h
+krb5_HEADERS = kdc-audit.h kdc-plugin.h kdc-accessors.h token_validator_plugin.h csr_authorizer_plugin.h gss_preauth_authorizer_plugin.h
```
nicowilliams, line 247: Add to the dist-related macros too.

kdc/kdc-accessors.h
```diff
@@ -0,0 +1,349 @@
...
+
+/*
+ * HDB *
+ * kdc_request_get_clientdb(astgs_request_t);
+ */
+
+ASTGS_REQUEST_GET_ACCESSOR(HDB *, clientdb)
+
+/*
+ * hdb_entry *
```
nicowilliams, line 274: Can we make this const?

lhoward: Ah yes, that was on my list to do, thanks for reminding me (as well as `struct sockaddr` and a few others).

kdc/Makefile.am
lhoward: Hmm, it's in the same list as the other plugin headers like `kdc-plugin.h`. Should those too be in another list?

kdc/kdc-accessors.h
lhoward: Const-ness for both `HDB *` and `hdb_entry *`? The only reason not to do this is, well, a bunch of HDB functions don't take `const`, e.g. what if a plugin wants to call `hdb_check_client_matches_target_service`; it will need to cast away the const-ness. The solution obviously is to fix the HDB entry points but, that belongs in another patch. What do you think?

nicowilliams:
This is fantastic.  I love it.  Thanks!

PR #944, opened 2022-01-20
assert non-NULL ptrs before calling mem funcs

jaltman:
The definitions of memcpy(), memmove(), and memset() state that
the behaviour is undefined if any of the pointer arguments are
NULL, and some compilers are known to make use of this to
optimise away existing NULL checks in the source.

Change-Id: I489bc256e3eac7ff41d91becb0b43aba73dbb3f9
Link: https://www.imperialviolet.org/2016/06/26/nonnull.html

### DIFF
--- a/lib/asn1/der_copy.c
+++ b/lib/asn1/der_copy.c
@@ -99,11 +99,15 @@ int ASN1CALL
 der_copy_printable_string (const heim_printable_string *from,
 		       heim_printable_string *to)
 {
-    to->length = from->length;
-    to->data   = malloc(to->length + 1);
-    if(to->data == NULL)
+    assert(from->length == 0 || (from->length > 0 && from->data != NULL));
+    to->data = malloc(from->length + 1);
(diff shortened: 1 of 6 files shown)


lib/asn1/der_copy.c
```diff
@@ -99,11 +99,13 @@ int ASN1CALL
 der_copy_printable_string (const heim_printable_string *from,
 		       heim_printable_string *to)
 {
-    to->length = from->length;
-    to->data   = malloc(to->length + 1);
-    if(to->data == NULL)
+    assert(from->length == 0 || (from->length > 0 && from->data != NULL));
+    to->data = malloc(from->length + 1);
+    if (to->data == NULL)
 	return ENOMEM;
```
nicowilliams, line 105: Do we want to initialize `to->length` in the `ENOMEM` case?

jaltman: I've added setting length to zero at all error returns in the next version of the patch.

PR #947, opened 2022-01-21
Rework  `_kdc_find_etype()`

nicowilliams:
`_kdc_find_etype()` has become a monster.  It's time to rewrite it.

I'm splitting it into two parts:

 - `_kdc_find_etype()`, meant for selecting a ticket's session key enctype (though it's also used in GSS pre-auth to pick the enctype for which to derive a reply key, but the considerations are the same in that case)

 - `_kdc_find_key_and_salt()`, which is used to pick _a_ key and salt for use in `PA-ETYPE-INFO` and `PA-ETYPE-INFO2`

The latter is similar in function to `_kdc_get_preferred_key()`, but whereas the `KDC-REQ-BODY`'s `etype` list does not affect service key selection, it does affect client key selection (I _think_ so anyways), so `_kdc_find_key_and_salt()` does call `_kdc_find_etype()`.


kdc/kerberos5.c
```diff
@@ -154,209 +154,271 @@ _kdc_is_anon_request(const KDC_REQ *req)
...
+        princ = r->server;
+
+    if (is_preauth && r->config->preauth_use_strongest_session_key)
+        use_strongest_session_key = TRUE;
+    else if (is_tgs)
+        use_strongest_session_key = r->config->tgt_use_strongest_session_key;
+    else
+        use_strongest_session_key = r->config->svc_use_strongest_session_key;
+
+    /* AFS may still be stuck using 1DES (XXX Make this configurable!) */
```
nicowilliams, line 232: I think this really should be configurable, but I decided to post this PR now for your review.

kdc/kerberos5.c
```diff
@@ -154,209 +154,271 @@ _kdc_is_anon_request(const KDC_REQ *req)
...
-            enctype = ETYPE_DES_CBC_CRC;
-        } else {
-            ret = KRB5KDC_ERR_ETYPE_NOSUPP;
+    ret = _kdc_find_etype(r, KFE_USE_CLIENT | KFE_IS_PREAUTH, &enctype);
+    if (ret)
+        return ret;
+
+    princ = r->client;
+
+    /* XXX So use an alias for the salt? */
```
nicowilliams, line 391: Is this right?

kdc/kerberos5.c
```diff
@@ -154,209 +154,271 @@ _kdc_is_anon_request(const KDC_REQ *req)
...
+        /* A bit of krb5_enctype_valid(), inlined */
+        if ((et = _krb5_find_enctype(outer[o])) == NULL ||
+            (et->flags & F_DISABLED))
+            continue;
+
+        if (!use_strongest_session_key && outer[o] == ETYPE_DES_CBC_CRC)
+            weak_offered = 1;
+
+        for (i = 0; i < ninner; i++) {
+            if (--limit == 0) {
```
nicowilliams, line 292: A bit ugly.

nicowilliams:
`check-des` is broken.

kdc/kerberos5.c
nicowilliams: Never mind.

kdc/kerberos5.c
nicowilliams: Yes.

lib/krb5/crypto.c
```diff
@@ -799,10 +799,63 @@ krb5_checksum_disable(krb5_context context,
...
+    case KRB5_ENCTYPE_AES128_CTS_HMAC_SHA256_128:
+        return &_krb5_enctype_aes128_cts_hmac_sha256_128;
+    case KRB5_ENCTYPE_AES256_CTS_HMAC_SHA384_192:
+        return &_krb5_enctype_aes256_cts_hmac_sha384_192;
+    case KRB5_ENCTYPE_ARCFOUR_HMAC_MD5:
+        return &_krb5_enctype_arcfour_hmac_md5;
+    default:
+        for (i = 0; i < _krb5_num_etypes; i++)
+            if (_krb5_etypes[i]->type == type)
+                return _krb5_etypes[i];
```
lhoward, line 857: Hmm, isn't the compiler going to optimise to this anyway? Is `_krb5_etypes` local to crypto.c? Seems a bit premature optimisation-ish...

nicowilliams: Why would it compile it away?

The problem is this is an O(N) algorithm invoked from loops elsewhere.

lhoward: Can’t it unroll the etype list?

kdc/kerberos5.c
```diff
@@ -154,209 +154,271 @@ _kdc_is_anon_request(const KDC_REQ *req)
...
-	request_princ = r->client_princ;
-    } else {
-	princ = r->server;
-	request_princ = r->server->principal;
+    if (!cached_kerberos_enctypes) {
+        /*
+         * NOTE: This is either all strong enctypes only, or all enctypes,
+         *       including weak ones, depending on configuration.
+         */
+        cached_kerberos_enctypes = krb5_kerberos_enctypes(r->context);
```
lhoward, line 206: Hmm, introduced global shared state when we've been trying to refactor everything to use `r`. Can you cache these in the context instead?

kdc/kerberos5.c
```diff
@@ -154,209 +154,271 @@ _kdc_is_anon_request(const KDC_REQ *req)
...
+                return KRB5KDC_ERR_ETYPE_NOSUPP;
+            }
+            if (use_strongest && inner[i] == ETYPE_DES_CBC_CRC)
+                weak_offered = 1;
+            if (outer[o] != inner[i])
+                continue;
+
+            if (use_client && !is_preauth) {
+                /*
+                 * This is the AS ticket session key negoatiation case, so the
```
lhoward, line 314: spelling

kdc/kerberos5.c
nicowilliams: It's a global in `krb5_kerberos_enctypes()`...

lib/krb5/crypto.c
nicowilliams: I guess a compiler could turn the loop into a switch, yeah.  I will examine the dissassembly, though I'm not sure I want to assume a compiler this smart.

kdc/kerberos5.c
lhoward: Then why do we need to cache it? And why does it take a context argument? (Not in front of computer :))

nicowilliams: Because it's a pair of `static` variables inside that function, `krb5_kerberos_enctypes()`...

lhoward: Hmm, OK, so relying on `krb5_enctype_valid()` being invariant across contexts, which I guess is safe. But, why can't we just call `krb5_enctype_valid()` each time?

nicowilliams: Ah, that: because it does a lot of work on error that we don't need to do in this context.  I'm manually inlining just the piece we need.

PR #949, opened 2022-01-25
lib/hdb: Add a flag to force canonicalization

cryptomilk:
We need a force canonicalization flag for the Samba database module.

### DIFF
--- a/lib/hdb/hdb.h
+++ b/lib/hdb/hdb.h
@@ -76,6 +76,7 @@ enum hdb_lockop{ HDB_RLOCK, HDB_WLOCK };
 #define HDB_F_DELAY_NEW_KEYS	32768	/* apply [hdb] new_service_key_delay */
 #define HDB_F_SYNTHETIC_OK	65536	/* synthetic principal for PKINIT or GSS preauth OK */
 #define HDB_F_GET_FAST_COOKIE  131072	/* fetch the FX-COOKIE key (not a normal principal) */
+#define HDB_F_FORCE_CANON	262144	/* force canonicalization */
 
 /* hdb_capability_flags */
 #define HDB_CAP_F_HANDLE_ENTERPRISE_PRINCIPAL 1


nicowilliams:
Is the `HDB_entry`'s `flags`' `force_canonicalize` insufficient because you need a flag for lookup?

jaltman:
HDB_F_FORCE_CANON will not be used in-tree by Heimdal.  Please provide an example of how it will be used by Samba.

If necessary, perhaps a flag bit or two can be reserved for third-party use.

jaltman:
Related question.  Is HDB_F_CURRENT_KVNO used by Samba?  
The last in-tree use of HDB_F_CURRENT_KVNO was removed in 2011 by 308e53a4a88445973c14f788977883e421914691

josephsutton1:
> Related question. Is HDB_F_CURRENT_KVNO used by Samba?

No, Samba currently doesn't use HDB_F_CURRENT_KVNO anywhere.

cryptomilk:
Samba has a glue layer called SDB. It is very similar to HDB. It is used to either translate to HDB (Heimdal) or KDB (MIT Kerberos). In SDB we have a flag SDB_F_FORCE_CANON which is set to 16384. However as we directly pass HDB flags to SDB and do not translate them, the value 16384 has been assigned to HDB_F_PRECHECK.

So if Heimdal would pass that flag to get_principal() then we would force canonicalization now. In order to avoid that, I would like to define it in Heimdal. We could also use HDB_F_CURRENT_KVNO and rename it to HDB_F_FORCE_CANON if that is preferred.

https://gitlab.com/samba-team/samba/-/blob/master/source4/kdc/db-glue.c#L986

lhoward:
Irrespective of assignments it’s probably a good idea to have a translation layer for flags.

abartlet:
Kia Ora @cryptomilk 

I don't think this is the right approach.  SDB should take two arguments, HDB flags and extra Samba flags, that KDB can use to signal other things that HDB doesn't have a flag for. 

Sorry,

Andrew Bartlett

jaltman:
> > Related question. Is HDB_F_CURRENT_KVNO used by Samba?
> 
> No, Samba currently doesn't use HDB_F_CURRENT_KVNO anywhere.

Thank you for confirming.  We will remove the unused flag before 8.0.

jaltman:
> the value 16384 has been assigned to HDB_F_PRECHECK.

The HDB_F_PRECHECK flag was assigned 16384 more than seven years ago.  

HDB and KDB flag namespaces are private to their respective implementations.    

> SDB should take two arguments, HDB flags and extra Samba flags, that KDB can use to signal other things that HDB doesn't have a flag for.

If its possible to do so, this is a much better implementation because it avoids the possibility of the flag namespaces colliding.

nicowilliams:
So, close this, yes?

PR #951, opened 2022-01-28
remove base dependency from libasn1

lhoward:


nicowilliams (approved):

nicowilliams:
Thanks so much for this!

PR #963, opened 2022-03-02
Introduce HDB_F_FOR_S4U2SELF

metze-samba:
The will be used to allow the addition of special SIDS into the
    PAC:
    
    - SID_AUTHENTICATION_AUTHORITY_ASSERTED_IDENTITY (S-1-18-1):
    
      This is added during the AS-REQ/AS-REP exchange after
      pre-authentication was successful.
    
    - SID_SERVICE_ASSERTED_IDENTITY (S-1-18-2):
      This is added during S4U2Self PAC creation.
      It won't replace a possible
      SID_AUTHENTICATION_AUTHORITY_ASSERTED_IDENTITY
      during S4U2Proxy.

### DIFF
--- a/kdc/mssfu.c
+++ b/kdc/mssfu.c
@@ -110,7 +110,7 @@ validate_protocol_transition(astgs_request_t r)
     EncTicketPart *ticket = &r->ticket->ticket;
     hdb_entry *s4u_client = NULL;
     HDB *s4u_clientdb;
-    int flags = HDB_F_FOR_TGS_REQ;
+    int flags = HDB_F_FOR_TGS_REQ | HDB_F_FOR_S4U2SELF;
     krb5_principal s4u_client_name = NULL, s4u_canon_client_name = NULL;
     krb5_pac s4u_pac = NULL;
     const PA_DATA *sdata;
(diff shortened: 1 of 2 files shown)


lhoward:
Not opposed, but can it be determined by the plug-in interrogating the request for S4U2Self PA data?

lhoward:
Also I’d remove `FOR` from the macro name.

lhoward:
I think it should be fine to call `kdc_request_get_req()` and then `krb5_find_padata(KRB5_PADATA_FOR_USER)` as there are no successful paths where `KRB5_PADATA_FOR_USER` is present and S4U2Self is not performed. Closing for now, please reopen if you disagree.

metze-samba:
> I think it should be fine to call `kdc_request_get_req()` and then `krb5_find_padata(KRB5_PADATA_FOR_USER)` as there are no successful paths where `KRB5_PADATA_FOR_USER` is present and S4U2Self is not performed. Closing for now, please reopen if you disagree.

Yes, that's fine for now, when we add support for the new style PA-S4U-X509-USER S4U2Self version,
but I guess then we could have a 'bool kdc_request_get_s4u2self_used()' for easy access.

PR #964, opened 2022-03-02
allow _kdc_audit_request() to return an error, e.g. HDB_ERR_NOT_FOUND_HERE

metze-samba:
This is (at least) required to trigger HDB_ERR_NOT_FOUND_HERE on
    pre-authentication failures on an RODC.
    
    Read-Only DC's need to pass failing pre-authentications to a Read-Write DC
    
    BUG: https://bugzilla.samba.org/show_bug.cgi?id=14865

### DIFF
--- a/kdc/kerberos5.c
+++ b/kdc/kerberos5.c
@@ -2069,6 +2069,7 @@ _kdc_as_rep(astgs_request_t r)
     krb5_boolean is_tgs;
     const char *msg;
     Key *krbtgt_key;
+    krb5_error_code audit_ret;
 
     memset(rep, 0, sizeof(*rep));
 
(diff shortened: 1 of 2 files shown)


lhoward:
Hmm, this wasn’t really the intent of the audit API but I’ll investigate the right way to do this :)

metze-samba:
> Hmm, this wasn’t really the intent of the audit API but I’ll investigate the right way to do this :)

But was it intended to ignore the return value completely?
Things could fail for other reasons and I think they should also result in an
error for the client.

metze-samba:
> Can you update the error code using an accessor?

We still use the hdb audit hook, but with the new kdc audit it might be possible.

But then the code has to look like this:

```
r->error_code = ret;
_kdc_audit_request(r);
ret = r->error_code;
```

Which also looks strange.

abartlet:
The thing to remember is that the strange HDB_ERR_NOT_FOUND_HERE is an
error return to the caller of the KDC function, not an error return to
the client.

lhoward:
OK, will look into properly once back at desk.

PR #965, opened 2022-03-02
kdc: add missing enctype = p[i] assignments to _kdc_find_etype()

metze-samba:
This is needed when a service provides support for newer entry.etypes
without having a key for that encryption type yet.
It is triggered by svc_use_strongest_session_key=TRUE


lhoward:
This is one for @nicowilliams

metze-samba:
@lhoward I think this should be labled as bug

jaltman (approved):

PR #967, opened 2022-03-02
kdc: provide kdc_request_get_explicit_armor_{clientdb,client,pac}()

metze-samba:
_kdc_fast_check_armor_pac() already checks the PAC of the armor,
but it should also remember it if it's an TGS-REQ with explicit armor.

This will allow the kdc pac hooks to generate a compound identity PAC
with PAC_TYPE_DEVICE_INFO.


lhoward:
Will merge but please take care to update export lists, use bitfields instead of booleans if containing structure uses them, and replace existing local boolean variables with context ones if supplanted.

PR #968, opened 2022-03-02
krb5: add new values and definitions from MS-KILE/MS-SFU

metze-samba:


metze-samba:
> @metze-samba When new ASN1 types are added the Windows function export list must be updated. Fixed by [19b337a](https://github.com/heimdal/heimdal/commit/19b337a0fb6f2c74926734ddbc01ab822b0fb05c)

@jaltman can you also add it to the unix export file, they are missing there too.

It would actually be great if someone could fix the Windows ci run...

PR #969, opened 2022-03-02
lib/krb5: re-allow data->length == 0 in krb5_pac_add_buffer()

metze-samba:
PAC_TYPE_CLIENT_CLAIMS_INFO and PAC_TYPE_DEVICE_CLAIMS_INFO are
of zero length unless any claims are actually defined.

### DIFF
--- a/lib/krb5/pac.c
+++ b/lib/krb5/pac.c
@@ -383,7 +383,7 @@ krb5_pac_add_buffer(krb5_context context, krb5_pac p,
     size_t len, offset, header_end, old_end;
     uint32_t i;
 
-    assert(data->length > 0 && data->data != NULL);
+    assert(data->length == 0 || (data->length > 0 && data->data != NULL));
 
     len = p->pac->numbuffers;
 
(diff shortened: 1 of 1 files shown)


lib/krb5/pac.c
```diff
@@ -383,7 +383,7 @@ krb5_pac_add_buffer(krb5_context context, krb5_pac p,
     size_t len, offset, header_end, old_end;
     uint32_t i;
 
-    assert(data->length > 0 && data->data != NULL);
+    assert(data->length == 0 || (data->length > 0 && data->data != NULL));
```
jaltman, line 386: Our requirement post 190263bb7a56fc775b50a6cd0dc91820d2b2e5eb is that any valid krb5_data must not have a NULL data pointer.   A valid krb5_data that represents a length of zero must have a non-NULL data pointer.   If both the length is zero and the data pointer is NULL, then then krb5_data is unallocated (it might have already been freed.).

lib/krb5/pac.c
```diff
@@ -433,8 +433,9 @@ krb5_pac_add_buffer(krb5_context context, krb5_pac p,
      * copy in new data part
      */
 
-    memcpy((unsigned char *)p->data.data + offset,
-	   data->data, data->length);
+    if (data->length != 0)
```
jaltman, line 436: There is no need to check for non-zero data length since the data pointer is not permitted to be NULL.

jaltman (changes requested):

lhoward:
Accidentally merged, but reverted and fixed properly in 50fb794ef1b049c60671d0c94c11d8954d4a92e0.

PR #970, opened 2022-03-03
Introduce KDC_AUTH_EVENT_HISTORIC_LONG_TERM_KEY

metze-samba:
In Samba we need to avoid incrementing badPwdCount if pre-authentication
fails using a password from the password history of a user.

See https://bugzilla.samba.org/show_bug.cgi?id=14054


kdc/kerberos5.c
```diff
@@ -848,68 +1058,75 @@ pa_enc_ts_validate(astgs_request_t r, const PA_DATA *pa)
...
+	    }
+
+	    hkvno = kvno - hi;
+	    hret = pa_enc_ts_decrypt_kvno(r, hkvno,
+					  &enc_data,
+					  &ts_data,
+					  NULL); /* pa_key */
+	    if (hret == 0) {
+		snprintf(hresult, sizeof(hresult),
+			 ", sucess with kvno=%d",
```
lhoward, line 1104: s/sucess/success/

lhoward:
Looks good in principle although I do not have time to review it now, so @nicowilliams or @jaltman will have to.

I didn't look too closely so this may not be an issue, but please try to maintain existing logging statement syntax, to avoid breaking third-party tools that parse logs.

kdc/kerberos5.c
metze-samba: Thanks!

metze-samba:
> Looks good in principle although I do not have time to review it now, so @nicowilliams or @jaltman will have to.
> 
> I didn't look too closely so this may not be an issue, but please try to maintain existing logging statement syntax, to avoid breaking third-party tools that parse logs.

If it's critical we can keep the output as is, but I thought it would good to print the kvno that failed and a possible
historic kvno that worked, but I can remove that and add some details using kdc_audit_setkv_number
which can be consumed by an audit plugin.

metze-samba:
> > Looks good in principle although I do not have time to review it now, so @nicowilliams or @jaltman will have to.
> > I didn't look too closely so this may not be an issue, but please try to maintain existing logging statement syntax, to avoid breaking third-party tools that parse logs.
> 
> If it's critical we can keep the output as is, but I thought it would good to print the kvno that failed and a possible historic kvno that worked, but I can remove that and add some details using kdc_audit_setkv_number which can be consumed by an audit plugin.

Should I add these?:

+#define KDC_REQUEST_KV_PA_SUCCESS_KVNO         "pa-sucess-kvno"        /* heim_number_t */
+#define KDC_REQUEST_KV_PA_FAILED_KVNO          "pa-failed-kvno"        /* heim_number_t */
+#define KDC_REQUEST_KV_PA_HISTORIC_KVNO                "pa-historic-kvno"      /* heim_number_t */

KDC_REQUEST_KV_PA_HISTORIC_KVNO would always also have
KDC_REQUEST_KV_PA_FAILED_KVNO

lhoward:
Yes, this is a better approach. Suggest the following symbolic names:

```c
#define KDC_REQUEST_KV_PA_SUCCEEDED_KVNO"pa-succeeded-kvno" /* heim_number_t /
#define KDC_REQUEST_KV_PA_FAILED_KVNO "pa-failed-kvno" / heim_number_t /
#define KDC_REQUEST_KV_PA_HISTORIC_KVNO "pa-historic-kvno" / heim_number_t */
```

Please match indentation to `kdc-audit.h`.

metze-samba:
> Yes, this is a better approach. Suggest the following symbolic names:
> 
> ```c
> #define KDC_REQUEST_KV_PA_SUCCEEDED_KVNO "pa-succeeded-kvno" /* heim_number_t */
> #define KDC_REQUEST_KV_PA_FAILED_KVNO "pa-failed-kvno" /* heim_number_t */
> #define KDC_REQUEST_KV_PA_HISTORIC_KVNO "pa-historic-kvno" /* heim_number_t */
>```

Thanks I'll take them.

> Please match indentation to `kdc-audit.h`.

Sure it's always broken within the browser...

nicowilliams:
I'll have to review this in-clone as GitHub is making the diffs difficult, but I'm doing other work at the moment.

kdc/kerberos5.c
```diff
@@ -1038,6 +1038,33 @@ pa_enc_ts_validate(astgs_request_t r, const PA_DATA *pa)
     if (ret == KRB5KDC_ERR_PREAUTH_FAILED) {
 	krb5_error_code ret2;
 	const char *msg = krb5_get_error_message(r->context, ret);
+	krb5_error_code hret = ret;
```
abartlet, line 1094: Could this be renamed historic_key_ret to make this clearer?

abartlet (approved):

abartlet:
I've looked over the changes in a local clone, and they are OK, as far as I can tell, with the assistance of 
```
[diff]
        colorMoved = default
        colorMovedWS = allow-indentation-change
```
in my `.gitconfig`

I would say that it would be easier to validate if some of the variable renames were done upfront, eg pepper1 -> pepper1kdc, as more lines would come over as unchanged or with smaller changes.

PR #972, opened 2022-03-03
Make krb5_init_creds_step() usable at all and match the MIT version of it

metze-samba:
NOTE: commit 1cdc9d5f3cff0288846c29c35ee91b6056a2e2bb
    "krb5: export krb5_init_creds_step()" exported
    krb5_init_creds_step() the first time, but that's
    not in any released version, so it should be fine
    to fix up the prototype in order to make the
    function actually useful for external callers.

All the "lib/krb5: remove unused krb5_krbhst_info argument of" commits are
not strictly required, but they show that the last commit
"lib/krb5: remove dead code from pk_verify_host()"
actually removed dead code. If the logic
in pk_verify_host() should be kept,
a krb5_krbhst_info should be passed in though any useful
way, but passing it to krb5_init_creds_step() is clearly not
the correct way. Maybe a new krb5_init_creds_set_*() function.

### DIFF
--- a/lib/krb5/pkinit.c
+++ b/lib/krb5/pkinit.c
@@ -1365,7 +1365,6 @@ pk_rd_pa_reply_dh(krb5_context context,
 		  const char *realm,
 		  krb5_pk_init_ctx ctx,
 		  krb5_enctype etype,
-		  const krb5_krbhst_info *hi,
 		  const DHNonce *c_n,
 		  const DHNonce *k_n,
                   unsigned nonce,
@@ -1407,7 +1406,7 @@ pk_rd_pa_reply_dh(krb5_context context,
 
     if (host) {
 	/* make sure that it is the kdc's certificate */
-	ret = pk_verify_host(context, realm, hi, ctx, host);
+	ret = pk_verify_host(context, realm, NULL, ctx, host);
 	if (ret)
 	    goto out;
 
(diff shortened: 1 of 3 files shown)


lib/krb5/pkinit.c
```diff
@@ -1092,18 +1091,6 @@ pk_verify_host(krb5_context context,
     if (ret)
 	return ret;
 
-    if (hi) {
-	ret = hx509_verify_hostname(context->hx509ctx, host->cert,
```
lhoward, line 1096: A caller to `krb5_init_creds_step()` could specify `hostinfo`, which would no longer be validated by `hx509_verify_hostname()` with this patch?

metze-samba: In theory yes, but the only caller is krb5_init_creds_get() that that passes NULL,
and there are no external callers because krb5_init_creds_step was luckily never exported
in a released version.

metze-samba:
@lhoward can we at least revert commit 1cdc9d5f3cff0288846c29c35ee91b6056a2e2bb for now,
so that krb5_init_creds_step() doesn't get any external users with the broken prototype?

lhoward:
Sounds like a sensible idea. @nicowilliams? (Sorry I don't have time to work on this right now.)

metze-samba:
> Sounds like a sensible idea. @nicowilliams? (Sorry I don't have time to work on this right now.)

See https://github.com/heimdal/heimdal/pull/974

jaltman:
> > Sounds like a sensible idea. @nicowilliams? (Sorry I don't have time to work on this right now.)
> 
> See #974

@lhoward replied to #974 indicating that there is an existing out of tree caller of the existing krb5_init_creds_step().

nicowilliams:
I'll have to check that OSKT also isn't using this.  If the only out-of-tree caller is @lhoward's then I would be OK with taking this breaking change if @lhoward also is.

lhoward:
I've updated my code, so I am fine for this to be merged (assuming it passes tests etc).

jaltman (approved):

jaltman:
> I've updated my code, so I am fine for this to be merged (assuming it passes tests etc).

I rebased and ran checks before merging.

PR #974, opened 2022-03-05
krb5: don't export krb5_init_creds_step()

metze-samba:
The current prototype is unusable for external callers
and no released version exported it yet.

https://github.com/heimdal/heimdal/pull/972 will hopefully
fix it, but for now we should keep it internal.

### DIFF
--- a/lib/krb5/fast.c
+++ b/lib/krb5/fast.c
@@ -826,7 +826,7 @@ _krb5_fast_anon_pkinit_step(krb5_context context,
 
     anon_pk_ctx = state->anon_pkinit_ctx;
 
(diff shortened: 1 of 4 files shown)


lhoward:
Hmm, PADL/TktBridgeAP uses `krb5_init_creds_step()` as is so I wouldn't be in favour of this change.

lhoward:
So, my out-of-tree caller is not being _used_ by anyone so it's mostly vanity not wanting to break it. But... maybe we should just merge #972 and be done with it? Indeed if you could make `out_realm` permissibly NULL I wouldn't need to change anything at all ;) (But I'm not suggesting you do that.)

jaltman:
https://github.com/heimdal/heimdal/pull/972 has been merged.

PR #977, opened 2022-03-08
lib/krb5: Make parameters to PAC functions 'const'

josephsutton1:
This allows these functions to be used with PACs obtained from KDC accessor functions such as kdc_request_get_pac().


lhoward:
LGTM

PR #979, opened 2022-03-10
lib/krb5: before we create a new FAST armor we should clear the old s…

metze-samba:
…trengthen_key

In a cross-realm situation the client KDC exchange may use on orphaned
strengthen_key (from the previous exchange) if the current KDC
doesn't not support FAST and the previous KDC supported it.

Otherwise init_creds_step() or fast_tgs_strengthen_key()
generate the reply key.

BUG: https://bugzilla.samba.org/show_bug.cgi?id=15005


lhoward:
Looks OK to me.

nicowilliams:
How is cross-realm involved here?

PR #980, opened 2022-03-10
kdc: add enable_fast option (enabled by default)

metze-samba:
This makes it possible to disable fast if needed,
which is very good for regression tests.
Samba disables this in a few test environments with this
option.

BUG: https://bugzilla.samba.org/show_bug.cgi?id=15002
BUG: https://bugzilla.samba.org/show_bug.cgi?id=15005


lhoward:
LGTM

PR #981, opened 2022-03-16
Hard and soft HDB entry aliases

nicowilliams:
At some point we had "hard HDB entry aliases" working, and I broke it when I needed to add something like soft aliases -- I made all aliases soft.  This PR has a fix.

The fix is to distinguish soft and hard aliases by convention.

- Soft aliases are aliases of `WELLKNOWN/REFERRALS/TARGET@${some_realm}` or , `WELLKNOWN/REFERRALS/TARGET/...@${some_realm}`  where the realm is the realm to issue referrals to when a client asks for a ticket for an alias of such a principal.
  The second form is needed because there is a limit to how many aliases an HDB entry can have due to the signed 16-bit "TL-data" length limit imposed by the `kadm5` APIs and MIT kadmin protocol.

- Hard aliases are all others.

`kadmin`/`kadmind`/`httpkadmind`/`kadm5` will not care about the alias type distinction -- we push it onto users.

The `HDB` keytab can't do referrals, so soft aliases are "not found".

The AS will treat client principal soft aliases as referrals.

The TGS will treat service principal soft aliases as referrals.

Hard aliases get used as though they are proper principal entries.

- If the client does not request canonicalization, or if the alias and canonical name are in different realms, then we act as though the requested name were the canonical name of the entry found.
- Conversely, if the client does request canonicalization and the alias and canonical name are in the same realm, then the canonical name will be used in the ticket issued.

### DIFF
--- a/lib/hdb/common.c
+++ b/lib/hdb/common.c
@@ -181,6 +181,7 @@ fetch_entry_or_alias(krb5_context context,
         ret = decode_HDB_EntryOrAlias(value.data, value.length, &eoa, NULL);
     if (ret == 0 && eoa.element == choice_HDB_EntryOrAlias_entry) {
         *entry = eoa.u.entry;
+        entry->aliased = 0;
     } else if (ret == 0 && eoa.element == choice_HDB_EntryOrAlias_alias) {
         krb5_data_free(&key);
 	ret = hdb_principal2key(context, eoa.u.alias.principal, &key);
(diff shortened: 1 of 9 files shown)


lib/hdb/common.c
```diff
@@ -219,11 +223,58 @@ fetch_entry_or_alias(krb5_context context,
...
+ * Soft aliases should be treated as configuration to issue referrals, and they
+ * can only result in referrals to other realms.
+ *
+ * Rather than add a type of aliases, we'll use a convention where the form of
+ * the target of the alias indicates whether the alias is hard or soft.
+ *
+ * TODO We could also use an attribute of the aliased entry.
+ */
+static
+int
```
lhoward, line 242: fold `static int` to single line

PR #982, opened 2022-03-18
kadmin online listing

nicowilliams:
I've had it with `kadmin list` being slow when there are _many_ principals in a database.

This PR adds a `kadm5_iter_principals()` function that uses a callback rather than gathering all [matching] principals into an array in memory.

This PR also adds a variation on the Heimdal kadmin protocol's `LIST` operation that is correspondingly online, implements it in `kadmind`, and makes the `kadmin` client use it.

Lastly, there is a test, which tests listing an HDB w/ 20 principal in chunk sizes of 7 (three chunks), 10 (two chunks), and 50 (one chunk).  Testing listing of an HDB w/ zero principals is... hard and pointless, so no need.

### DIFF
--- a/kadmin/check.c
+++ b/kadmin/check.c
@@ -73,7 +73,7 @@ do_check_entry(krb5_principal principal, void *data)
 	return 1;
 
     memset (&princ, 0, sizeof(princ));
-    ret = kadm5_get_principal(kadm_handle, principal, &princ,
+    ret = kadm5_get_principal(data, principal, &princ,
 			      KADM5_PRINCIPAL | KADM5_KEY_DATA);
     if(ret) {
 	krb5_warn(context, ret, "Failed to get principal: %s", name);
@@ -95,7 +95,7 @@ do_check_entry(krb5_principal principal, void *data)
     }
 
     free(name);
-    kadm5_free_principal_ent(kadm_handle, &princ);
+    kadm5_free_principal_ent(data, &princ);
 
     return 0;
 }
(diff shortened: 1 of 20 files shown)


nicowilliams:
Ahh, there's a re-entrance issue that is difficult to resolve.  Basically, if the callback given to `kadm5_iter_principals()` tries to `kadm5_get_principal()`, then `kadm5` is re-entered and bad things happen because we try to re-use the same open `kadm_handle`.  We need a way to clone a `kadm_handle` so we have a distinct open.  For local operations that means opening a separate `HDB` connection, while for remote operations that means opening a separate connection to `kadmind`.

nicowilliams:
I've added a way for `kadmin` to tell `kadmind` that the user interrupted a listing.  This also adds a way to negotiate protocol versions in the future by having the response to `NOP`s have a version number (current version is 0).

I've also added a `--upto=NUM` option to `kadmin get`/`kadmin list` to exercise interrupting of principal listing.

I'm also making `tests/kdc/check-kadmin` go fast.  No more running `kadmind` in one-operation debug mode.

Lastly, I'm fixing a bug where `kadmind` was not sending error codes on... error (I noticed I was seeing "End of file" errors when `kadmind` was rejecting a request).

nicowilliams:
The protocol for interrupting a long-running listing (e.g., if you have a realm with millions of principals), is that every time `kadmind` sends a chunk of the listing, it checks if the client has sent any data, and if so it will consider the listing interrupted, and will throw away the message.  The interrupt message is just a `NOP`.  Between Nagle's algorithm and the listing being fast, it is possible we'd miss the interrupt, which is why it has to be a NOP: so we can ignore it if it arrives late.

nicowilliams:
The test does run `kadmin list --upto=3 '*'`, but with the server having just 20 principals in its HDB... there's no way to exercise the server-side interruption w/o manually using a debugger (which, I did).

PR #983, opened 2022-03-21
KCM: Memory leak in `handle_read`

fihuer:
The `inmsg` field of the client structure is malloc/realloc'ed in `handle_read` but never free'ed in `maybe_close`.
Seems like Apple already fixed that with this.

### DIFF
--- a/lib/ipc/server.c
+++ b/lib/ipc/server.c
@@ -700,6 +700,7 @@ maybe_close(struct client *c)
     dispatch_release(c->out);
 #endif
     close(c->fd); /* ref count fd close */
+    free(c->inmsg);
     free(c);
     return 1;
 }


nicowilliams:
Looks fine.  Do we also need to `free(c->outmsg)` in case `handle_write()` didn't do it because errors from `write(2)`?

fihuer:
Hmmm, probably. We didn't detect that code path with valgrind but it may
still be a leak.

Does this one, or any other leak, are patched in Apple's Heimdal ?

Le lun. 21 mars 2022 à 17:59, Nico Williams ***@***.***> a
écrit :

nicowilliams:
> Hmmm, probably. We didn't detect that code path with valgrind but it may
still be a leak.

We have a large number of untested error-path-related code paths.  See the `coveralls` build (it's not as bad as it looks because a lot parts of Heimdal that are poorly tested are also not really used; we should really just remove such code).

> Does this one, or any other leak, are patched in Apple's Heimdal ?

I don't often check Apple's fork.  If you found this with valgrind, and the fix looks fine, that's good enough.  I'll merge this now, and if you would like to submit another PR for the `outmsg` error case, that's fine.

PR #985, opened 2022-03-25
kcm: Allow disabling automatic TGT renewal

steffen-kiess:
Add --no-automatic-renewal option to disable automatic TGT renewal.
Renewal can also be disabled by setting automatic_renewal to false
in the [kcm] section of krb5.conf.


nicowilliams:
Thanks!

PR #986, opened 2022-03-26
Some notes on building with VC2019

rodwiddowson:
I recently did a complete build using Visual Studio 2019 and SDK 10.0.19041.0.

There was a bit of a learn cliff and I have tried to capture my experience in the attached patch stream to windows\README.md.

I am **not** proposing that this be accepted as is.  It is more there to inform a rewrite of this page some time (and to act as memo to myself when I next loop back through here)

### DIFF
--- a/windows/README.md
+++ b/windows/README.md
@@ -1,14 +1,14 @@
 Building Heimdal for Windows
 ===================
 
-1. Introduction
----------------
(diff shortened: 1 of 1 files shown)


lhoward:
These LGTM, perhaps you can squash into one commit? cc @jaltman

PR #992, opened 2022-05-05
kdc: support pkinit_kdc_revoke for pkinit anchors

wokis:
This adds support for revocation checking pkinit anchors.

Windows expects `KRB5_KDC_ERR_CLIENT_NOT_TRUSTED` when the anchor certificate is revoked.

BUG: https://bugzilla.samba.org/show_bug.cgi?id=9612

### DIFF
--- a/kdc/kerberos5.c
+++ b/kdc/kerberos5.c
@@ -518,7 +518,11 @@ pa_pkinit_validate(astgs_request_t r, const PA_DATA *pa)
 
     ret = _kdc_pk_rd_padata(r, pa, &pkp);
     if (ret || pkp == NULL) {
-	ret = KRB5KRB_AP_ERR_BAD_INTEGRITY;
+	if (ret == HX509_CERT_REVOKED) {
+	    ret = KRB5_KDC_ERR_CLIENT_NOT_TRUSTED;	
+	} else {
+	    ret = KRB5KRB_AP_ERR_BAD_INTEGRITY;
+	}
 	_kdc_r_log(r, 4, "Failed to decode PKINIT PA-DATA -- %s",
 		   r->cname);
 	goto out;
--- a/kdc/pkinit.c
+++ b/kdc/pkinit.c
@@ -398,6 +398,7 @@ _kdc_pk_rd_padata(astgs_request_t priv,
     hx509_certs trust_anchors;
     int have_data = 0;
     const HDB_Ext_PKINIT_cert *pc;
+    hx509_revoke_ctx revoke_ctx;
 
     *ret_params = NULL;
 
@@ -458,6 +459,20 @@ _kdc_pk_rd_padata(astgs_request_t priv,
     hx509_verify_set_time(cp->verify_ctx, kdc_time);
     hx509_verify_attach_anchors(cp->verify_ctx, trust_anchors);
     hx509_certs_free(&trust_anchors);
+    
+    if (config->pkinit_kdc_revoke != NULL) {
+	hx509_revoke_init(context->hx509ctx, &revoke_ctx);
+	
+	ret = hx509_revoke_add_crl(context->hx509ctx, revoke_ctx, *config->pkinit_kdc_revoke);
+	if (ret) {
+	    hx509_revoke_free(&revoke_ctx);
+	    krb5_set_error_message(context, ret, "failed to add CRL");
+	    goto out;
+	}
+	
+	hx509_verify_attach_revoke(cp->verify_ctx, revoke_ctx);
+	hx509_revoke_free(&revoke_ctx);
+    }
 
     if (config->pkinit_allow_proxy_certs)
 	hx509_verify_set_proxy_certificate(cp->verify_ctx, 1);


jaltman:
@abartlet can Samba run this change through its CI tests?

jaltman:
@abartlet has this change passed the Samba CI tests?

abartlet:
@jaltman I could do a CI run with it but we have no tests for certificate revocation currently.  We hope to extend the python Kerberos testsuite to do that eventually, but nothing in our testsuite would check for certificate revocation paths sadly.

kdc/pkinit.c
```diff
@@ -458,6 +459,20 @@ _kdc_pk_rd_padata(astgs_request_t priv,
     hx509_verify_set_time(cp->verify_ctx, kdc_time);
     hx509_verify_attach_anchors(cp->verify_ctx, trust_anchors);
     hx509_certs_free(&trust_anchors);
+    
+    if (config->pkinit_kdc_revoke != NULL) {
+	hx509_revoke_init(context->hx509ctx, &revoke_ctx);
+	
+	ret = hx509_revoke_add_crl(context->hx509ctx, revoke_ctx, *config->pkinit_kdc_revoke);
```
nicowilliams, line 466: We already do this in `_krb5_pk_load_id()`, which gets called in `krb5_kdc_pk_initialize()`, which gets called in `krb5_kdc_pkinit_config()`, which gets called in `configure()`, which gets called in `main()` in the KDC.

Or was that not working?  And if, do you know why?

kdc/kerberos5.c
```diff
@@ -518,7 +518,11 @@ pa_pkinit_validate(astgs_request_t r, const PA_DATA *pa)
 
     ret = _kdc_pk_rd_padata(r, pa, &pkp);
     if (ret || pkp == NULL) {
-	ret = KRB5KRB_AP_ERR_BAD_INTEGRITY;
+	if (ret == HX509_CERT_REVOKED) {
```
nicowilliams, line 521: LGTM.

kdc/pkinit.c
abartlet: Samba doesn't use the configure() (we start further in), but we do call krb5_kdc_pkinit_config(), so I'm not sure.

nicowilliams: Can you double-check?  Although this is harmless, since `hx509_revoke_add_crl()` will see that the CRL has already been loaded and return w/o doing anything.  If Samba is already calling `krb5_kdc_pkinit_config()` then this bit is not needed at all and I'd rather you drop it.

wokis: Samba indeed calls `krb5_kdc_pkinit_config()`. The missing piece seems to be to use `hx509_verify_attach_revoke(cp->verify_ctx, kdc_identity->revokectx);` so that the revocation context from `_krb5_pk_load_id()` is used here. The only downside is that it would require a restart of Samba to reload the CRL file.

@abartlet Would it be possibile to let Samba redo the `krb5_kdc_pkinit_config()` on reload (for example through `smbcontrol smbd reload-config`?

nicowilliams: The revocation context does get attached in `_krb5_pk_load_id()`.

We should make it possible to reconfigure the KDC w/o restarting it, indeed.  I've been meaning to do something like that, and I'd like even to be able to store configurations in the HDB (and have done some of the work for that) so that we can reconfigure automatically via iprop (though, that does sound a bit scary).

wokis: > The revocation context does get attached in `_krb5_pk_load_id()`.
It indeed does but without adding `hx509_verify_attach_revoke` to `_kdc_pk_rd_padata` I haven't been able to get revocation checking to work.

josephsutton1: `krb5_kdc_pkinit_config()` indeed calls `_krb5_pk_load_id()`, which sets up `verify_ctx` and `revokectx` on the static variable `kdc_identity`. However, as far as I can see, those contexts are never referenced again anywhere else. The verification context used in `_kdc_pk_rd_padata()` is an entirely different one that otherwise would have no revocation context attached.

So we ought either to call `hx509_verify_attach_revoke(cp->verify_ctx, kdc_identity->revokectx)`, as @wokis suggested in a previous comment, or to create a new revocation context as this MR does. But if we go with the latter option, we should add all the CRLs specified in the `pkinit_kdc_revoke` option, not just the first.

IMO the former option is preferable as it means less duplicated code. Reloading CRLs isn’t a problem, for there is already code to handle that: https://github.com/heimdal/heimdal/blob/f62e2f278437ff6c03d2d09bd628381c795bba78/lib/hx509/revoke.c#L812-L823

josephsutton1: Thus my suggested change would be: https://github.com/heimdal/heimdal/commit/aab2f8e5cb537d5a8241fba7bf9b7eb4c2e45fae

jsutton24:
This was merged with commit abd35b246a7ab3b304c3ca918e52240154b87008.

elric1:
Merged, so we can close it?

jsutton24:
Yes, thank you!

PR #993, opened 2022-05-06
Support DIR in hx509_revoke_add_crl

### LINKED ISSUE #991: Support for multiple CRLs in `pkinit_revoke` parameter
**Is your feature request related to a problem? Please describe.**
`pkinit_anchors` supports `DIR:`, however `pkinit_revoke` does not.

**Describe the solution you'd like**
Add `DIR:` support for `pkinit_revoke`.

**Describe alternatives you've considered**
Comma separation of files, such as used in `pkinit_identity` does not work either.

**Additional context**
n/a

wokis:
Resolves #991

### DIFF
--- a/lib/hx509/revoke.c
+++ b/lib/hx509/revoke.c
@@ -697,6 +697,11 @@ hx509_revoke_add_crl(hx509_context context,
 
     if (strncmp(path, "FILE:", 5) == 0) {    
 	path += 5;
+	    
+	for (i = 0; i < ctx->crls.len; i++) {
+		if (strcmp(ctx->crls.val[i].path, path) == 0)
+	    	    return 0
+	}
 	
 	ret = _hx509_revoke_add_crl(context, ctx, path);
 	


lib/hx509/revoke.c
```diff
@@ -628,41 +628,15 @@ load_crl(hx509_context context, const char *path, time_t *t, CRLCertificateList
...
-	return HX509_UNSUPPORTED_OPERATION;
-    }
-
-
-    path += 5;
-
-    for (i = 0; i < ctx->crls.len; i++) {
-	if (strcmp(ctx->crls.val[i].path, path) == 0)
-	    return 0;
-    }
```
jaltman, line 664: The above block prevents the inclusion of duplicate entries in `crls`.  Why should duplicate entries be permitted?

lib/hx509/revoke.c
```diff
@@ -693,6 +667,83 @@ hx509_revoke_add_crl(hx509_context context,
...
+	    hx509_set_error_string(context, 0, HX509_UNSUPPORTED_OPERATION,
+			           "Could not open CRL directory %s", path);
+	    return HX509_UNSUPPORTED_OPERATION;
+	}
+	
+	do {
+	    if ((dir = readdir(dirp)) != NULL) {
+		if (strcmp(dir->d_name, ".") == 0 || strcmp(dir->d_name, "..") == 0)
+		    continue;
+
```
jaltman, line 723: Are there other exclusions that should be enforced?

- Are there a well-known file name extensions that should be required?
- Are there illegal characters that should be excluded?
- Should backup files left by editors be excluded?
- Should files that begin with a '.' be excluded?
- Should directories be excluded?
- Should symlinks be excluded?

jaltman (commented):
This would be useful functionality to add but there are some open questions.

lib/hx509/revoke.c
wokis: They shouldn't and I added back the relevant code block.

lib/hx509/revoke.c
wokis: - Are there a well-known file name extensions that should be required?
  - Perhaps it would be better to just allow .pem and .der?
- Are there illegal characters that should be excluded?
- Should backup files left by editors be excluded?
- Should files that begin with a '.' be excluded?
  - I believe not other Heimdal "DIR:" configuration handle those cases by excluding but I could me wrong. Ultimately it all depends on how well we want to protect the user from misconfiguration.
- Should directories be excluded?
  - We really should skip them since it makes no sense in trying to do a `_hx509_revoke_add_crl` on such entries.
- Should symlinks be excluded?
  - I believe not other Heimdal "DIR:" configuration handle those cases by excluding but I could me wrong. Ultimately it all depends on how well we want to protect the user from misconfiguration.

jaltman: @nicowilliams do you have an opinion on the inclusion / exclusion policy to apply here?

lib/hx509/revoke.c
```diff
@@ -628,41 +628,15 @@ load_crl(hx509_context context, const char *path, time_t *t, CRLCertificateList
...
- * @param ctx hx509 revokation context
- * @param path path to file that is going to be added to the context.
- *
- * @return An hx509 error code, see hx509_get_error_string().
- *
- * @ingroup hx509_revoke
- */
-
 HX509_LIB_FUNCTION int HX509_LIB_CALL
-hx509_revoke_add_crl(hx509_context context,
```
nicowilliams, line 644: Why make this private?  In principle it breaks our ABI.

lib/hx509/revoke.c
nicowilliams: Yes, OpenSSL does hash directories, and opens files in there like so:

```
321             {
322                 BIO_snprintf(b->data, b->max,
323                              "%s%c%08lx.%s%d", ent->dir, c, h, postfix, k);
324             }
```

where `ent->dir` is the configured directory, `c` is the path component separator character (`/` on Unix, `\` on Windows), `h` is the hash, `postfix` is `""` if looking for a trust anchor or `"r"` if looking for a CRL, and `k` is a small integer for when multiple certificates or CRLs share the same 8 hex character hash.

But we're not _searching_ for a particular CRL here.  So taking every file that a) doesn't start with `.` and b) ends in `.pem` or `.der` should be good enough!  Then we don't need to hash/rehash the directory in order to be able to use the CRLs, but on the downside, we'd be reading CRLs we might not need, thus wasting I/O and CPU cycles.  For now, however, this will do.

nicowilliams: I guess we need to decide whether we want `DIR` to work like OpenSSL's, in which case we should check the directory when we need it, and we should just `open()` the relevant hash rather than `readdir()` and load all.  The OpenSSL approach is more performant, but we can always switch to it later, and implementing it now might be too much work.  Thoughts?

PR #997, opened 2022-06-21
updates for xcode 14

dariaphoebe:
5a702ff is one possible implementation tho we should consider using a macro other than fallthrough or just doing it like apple does in xcode 14.


jaltman:
Replaced by https://github.com/heimdal/heimdal/pull/1010

PR #999, opened 2022-07-01
make --shuffle build fixes

trofi:
`git` version of `GNU make` added `make --shuffle` mode in https://savannah.gnu.org/bugs/index.php?62100.

This mode managed to find a few missing build-time dependencies in `heimdal`:

      hdb: add missing build dependency on "hdb-protos.h"
      doc: add dependency on vars.texi
      gssapi: add dependency on gkrb5_err.h

Before the change `make --shuffle -j16` did not survive more than 2 consecutive builds for me.
After the change I saw 15 successful consecutive builds until I gave up.


jaltman (approved):

jaltman:
neat tool.  thank you.

PR #1004, opened 2022-08-05
heimdal: Fix the 32-bit build on FreeBSD

vlendec:
Format string fixes that fail in the Samba build on a 32-bit machine

Signed-off-by: Volker Lendecke <vl@samba.org>


jaltman:
We have to be careful with use of %zd because Windows compilers before VS2015 do not support it.   Must use %Id instead.

Also, we don't want to cast times to (long).   That will only make our existing problems with 32-bit time worse.

PR #1010, opened 2022-09-16
fallthrough logic breaks on macOS 14 SDK with clang 14

jaltman:
Apple clang version 14.0.0 (clang-1400.0.17.3.1) fails the build because stds.h defines `fallthrough` as a macro which is then expanded when base.h evaluates

  # if __has_attribute(fallthrough) && __clang_major__ >= 5

The macOS SDK defines `DISPATCH_FALLTHROUGH` as the macro instead of `fallthrough`.

This change replaces the use of `fallthrough` in the tree with `HEIM_FALLTHROUGH` and updates the declaration in configure logic to define `HEIM_FALLTHROUGH` based upon existing definitions (if any) of `fallthrough` or `DISPATCH_FALLTHROUGH`.


jaltman:
This change has been verified on macOS 13.0 clang 1400.0.17.3.3 arm64-apple-darwin22.1.0

Issue #1011, opened 2022-09-22
KDC: TGS-REQ problem with Windows 11 22H2 client


branciar:
We have an Active Directory domain with a outgoing trust relationship to a Heimdal KDC Kerberos realm.

Users authenticate on the Heimdal realm from AD-joined workstations.

We just encountered a problem with the latest Windows 11 version (22H2):

the TGS-REQ issued by the Windows client to the Heimdal KDC for the krbtgt/AD-DOMAIN@HEIMDAL-DOMAIN principal fails with "Decrypt integrity check failed":

Sep 22 12:24:42 mykdc kdc[683]: Failed to verify authenticator checksum: Decrypt integrity check failed for checksum type rsa-md5, key type aes256-cts-hmac-sha1-96
Sep 22 12:24:42 mykdc kdc[683]: Failed parsing TGS-REQ from IPv6:xx:xx:xx:xx:xx:xx:xx:xx
Sep 22 12:24:42 mykdc kdc[683]: tgs-req: sending error: -1765328353 to client

Under exactly same conditions, previous versions of Windows (up to Win11 21H2) obtain a successful TGS-REP from KDC.

Thorough wireshark inspection of the TGS-REQ packets from various Windows clients highlights only one difference :

the "till" timestamp of the "req-body" section is set to 99990913024805Z (9999-09-13 02:48:05 UTC) by the 22H2 Win11 client, whereas older Windows clients set it to 20370913024805Z (2037-09-13 02:48:05 UTC).

This let me suspect a trivial 32-bit Unix epoch counter overflow in KDC...

If so, this bug only occurs on TGS-REQ: in the logon process, Windows first sends AS-REQ containing the same odd timestamp (99990913024805Z value in the "till" and "rtime" fields of the req-body), which succeeds anyway.

Our KDC version is 7.7.0, on x64 Debian 11 platform (heimdal-kdc 7.7.0+dfsg-2 Debian package).

jaltman:
Heimdal attempted to increase the default requested maximum ticket lifetime and maximum renew lifetime to 0x7FFFFFFF seconds in a013e93e95a826b321870a7e626e2cdc9be13984.   This change had to be reverted because it broke the ability to request tickets.  See issue https://github.com/heimdal/heimdal/issues/220.  

A case should be opened with Microsoft to revert their change.   Requesting a maximum ticket lifetime greater than 9999 years is absolutely nuts.    
  
[...]

PR #1013, opened 2022-09-24
WIP Year 2038 fixes

nicowilliams:
This has two commits so far.  One to the DER decoder to clamp `GeneralizedTime` values post year 2038 to 2^31 - 1.  The other to rename all the `KerberosTime` fields in `lib/asn1/krb5.asn1` so that the compiler can help us find them and distinguish them from the `krb5_timestamp` fields of the same names in `krb5` data structures.  Next up we'll add a macro to clamp `KerberosTime` values whenever we refer to them.  And lastly we'll undo the renaming of those `KerberosTime` fields.

After all of that we should add a `krb5_timestamp64` type and add new structs like `krb5_times` and `krb5_last_req_entry` and new functions.  The kadmin API will be annoying, but for that one we can probably break the ABI (well, maybe not).


lib/asn1/der_get.c
```diff
@@ -511,18 +511,50 @@ generalizedtime2time (const char *s, time_t *t)
...
+         (tm_mon > 1 ||
+          (tm_mon == 1 &&
+           (tm.tm_mday > 19 ||
+            (tm.tm_mday == 19 &&
+             (tm.tm_hour > 3 ||
+              (tm.tm_hour == 3 &&
+               (tm.tm_min > 14 ||
+                (tm.tm_min == 14 &&
+                 tm.tm_sec >= 7))))))))))
+        return 2147483647;
```
lhoward, line 545: I dunno, is 7FFFFFFF maybe clearer?

lib/asn1/timegm.c
```diff
@@ -33,8 +33,6 @@
 
 #include "der_locl.h"
 
-#define ASN1_MAX_YEAR	2000
-
 static int
 is_leap(unsigned y)
```
lhoward, line 37: is `is_leap()` used now?

lib/asn1/timegm.c
```diff
@@ -50,19 +48,46 @@ static const unsigned ndays[2][12] ={
...
+       (tm->tm_mon > 1 ||
+        (tm->tm_mon == 1 &&
+         (tm->tm_mday > 19 ||
+          (tm->tm_mday == 19 &&
+           (tm->tm_hour > 3 ||
+            (tm->tm_hour == 3 &&
+             (tm->tm_min > 14 ||
+              (tm->tm_min == 14 &&
+               tm->tm_sec >= 7))))))))))
+      return 2147483647;
```
lhoward, line 90: Ditto re `0x7fffffff`

jaltman:
@nicowilliams , thank you for putting time into this issue when you had other personal matters to be dealing with.

I'm not sure where my in-line comments ended up.   Perhaps they went to the forked repository.   

To summarize.   I have significant concerns about the first change.   

1. timegm and gmtime have existed forever.   The fact that they were not used and private conversions were used instead is a strong indication to me that leap seconds are not supposed to be taken into account when converting to/from GeneralTime or KerberosTime.   This makes sense because the libraries that marshal and unmarshal are likely to have separate implementations of gmtime and timegm such that the leap second conversions would not be equivalent.   Any change that alters the use of leap seconds should be separate and explicit about what it is doing and why.   It should be backed by evidence that inclusion of leap seconds is the correct behavior.  Preferably referencing RFCs.
2. The `_der_timegm` behavior where UNIX Epoch is returned for any time that is too large is an intentional behavior.   Returning UNIX Epoch means that the requested time is the implementation MAX_TIME.   This proposed change removes that behavior and instead returns 0x7fffffff which might produce a very different result.    Note that MIT's behavior for a too large value in this case is to return an ASN1_BAD_TIMEGM error.   I think the existing Heimdal behavior is preferable from an interop perspective.
3. I do not believe that switching krb5_timestamp to unsigned 32-bit is a worthwhile change.   It isn't a safe API change to make because krb5_timestamp is a public type that is embedded in public structure definitions and in public function declarations.  It is not safe for the library to begin overflowing krb5_timestamp fields with values that will be treated as negative by the application.  Or for applications to assume that krb5_timestamp is unsigned and then be linked against Heimdal libraries where it is signed.    Similar to what was done internally by Apple to maintain the public MIT API Heimdal needs to separate its public types from its internal types.   Sadly I think we need to export an entirely new set of functions that support 64-bit time data structures and a new krb5_timestamp64.  Then use the 64-bit data structures internally and call the 64-bit safe functions from the KDC while exporting the 32-bit functions and structures for existing 32-bit aware applications.
4. Likewise, I think the ASN1 library can and should move away from use of `time_t` any capping that is performed when converting to `krb5_timestamp` should be done within lib/krb5 not within lib/asn1.   

In addition, I'm not convinced that these proposed changes would have any impact on addressing the problem identified by #1011.  The failure in that case is "Failed to verify authenticator checksum: Decrypt integrity check failed for checksum type rsa-md5, key type aes256-cts-hmac-sha1-96" which is more likely to be caused by a failure to perform the checksum on the raw input.

branciar:
In #1011, "Failed to verify authenticator checksum" error occurs because krb5_verify_checksum is not called on the raw input, but on a parsed an re-encoded copy of the req-body. So clamping timestamps to 2038 will absolutely not help in this case.

branciar:
I'm not sure clamping to 2038 in ASN.1 code is the right way. Asn1 is used by hx509 code to check validity timestamps of x509 certificates.
Furthermore, internal timestamps in asn1 seem to be of time_t type, and on my systems (64-bit Linux) time_t is a 64-bit int. Y2038 wrapping only occurs on systems with 32-bit time_t (32-bit Linux suffers from this). Maybe rather replace occurrences of time_t in asn1 by long ints (provided we never use plain <time.h> functions) ? This would avoid depending on architecture's time_t limitations.
Sounds like 2038 clamping should only be done in krb5 code, in places where MIT-compliant API is required.

nicowilliams:
Hmmm, maybe we're just trying too hard.  ABIs where `time_t` is signed 32-bit are just going to have year 2038 problems, and nothing we can do here can truly help.

The minimum changes we must make are:

- have `./configure` and/or `make` warn about building for ABIs where `time_t` is signed 32-bit,
  - and perhaps refuse to build there after some cut-off date like 2032-01-01
- change `FILE`, `KEYRING`, and `KCM` to use _unsigned_ 32-bit times in their formats/messages

For the latter, if the ABI has signed 32-bit `time_t` then year 2038 problems cannot be avoided (see above).

And then we can consider:

- making `lib/asn1/` use a `int64_t` for time _always_, possibly through a `typedef` alias that is not `time_t`, and _never_ use `time_t` itself

No changes are needed for ABIs where `time_t` is 64-bit or unsigned 32-bit.

This is _much_ simpler than the changes in this PR as of this time, and much simpler too than introducing new `krb5` APIs.

I'll spend some time thinking about this.  Feedback welcomed!

nicowilliams:
I'll open a new PR referencing this one when I'm ready.  It will do only the absolute minimal change described in my most recent comment above.

PR #1016, opened 2022-10-26
Fix HDB plugin crashes on 32-bit systems

josephsutton1:
Heimdal's HDB plugin interface, and hence Samba's KDC that depends upon it, doesn't work on 32-bit builds due to structure fields being arranged in the wrong order. This problem presents itself in the form of segmentation faults on 32-bit systems, but goes unnoticed on 64-bit builds thanks to extra structure padding absorbing the errant fields.

This commit reorders the HDB plugin structure fields to prevent crashes and introduces a common macro to ensure every plugin presents a consistent interface.

### DIFF
--- a/lib/hdb/hdb.h
+++ b/lib/hdb/hdb.h
@@ -307,11 +307,9 @@ typedef struct HDB {
 #define HDB_INTERFACE_VERSION	11
 
 struct hdb_method {
-    int			version;
+    HEIM_PLUGIN_FTABLE_COMMON_ELEMENTS(krb5_context);
     unsigned int	is_file_based:1;
     unsigned int	can_taste:1;
-    krb5_error_code	(*init)(krb5_context, void **);
-    void		(*fini)(void *);
     const char *prefix;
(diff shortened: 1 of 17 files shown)


jaltman (approved):

jaltman:
@josephsutton1 please rebase this change on top of the current tree

lib/hdb/hdb.h
```diff
@@ -307,11 +307,9 @@ typedef struct HDB {
 #define HDB_INTERFACE_VERSION	11
 
 struct hdb_method {
-    int			version;
+    HEIM_PLUGIN_FTABLE_COMMON_ELEMENTS(krb5_context);
     unsigned int	is_file_based:1;
```
nicowilliams, line 311: Aha, this is the issue, yes, that I added a bit-field after `version`, while the plugin framework in `lib/base` assumes the first three fields are `version`, `init`, and `fini`.  Good catch!

nicowilliams:
@jaltman proposes that we note that this is an ABI break for sites that run Heimdal from the `master` branch and which have out of tree HDB plugins.

PR #1017, opened 2022-10-27
lib/base: Avoid compiler warning about use-after-free on Fedora 36

abartlet:
While the local stack pointers could be thought of as "only" numbers that are not invalidated by the memory they point at being freed, any use of the pointer after the free is undefined and so warned about (at best).

gcc version 12.2.1 20220819 (Red Hat 12.2.1-1) (GCC)


abartlet:
The same pattern repeats further down the file at line 972.  I'll fix that also and re-push.

PR #1018, opened 2022-10-27
CVE-2022-3437 DES/DES3 overflow fixes

abartlet:
Patches from https://attachments.samba.org/attachment.cgi?id=17561 thanks to @josephsutton1 

Samba BUG: https://bugzilla.samba.org/show_bug.cgi?id=15134


jaltman:
All of the commits except for https://github.com/heimdal/heimdal/pull/1018/commits/93cdadfbc7f8a004890602062497620c733fe52e were included as part of the 7.7.1 | 7.8.0 releases.    The remaining commit was not merged because it breaks older MSVC compilers.    I will push that change separately after adding a proper fix for

PR #1021, opened 2022-11-14
ASN.1 compiler patches for Swift ASN1CodingKit

lhoward:
A few patches to the JSON emitted by `asn1compile` that are useful for [ASN1CodingKit](https://github.com/PADL/ASN1CodingKit).


jaltman:
@lhoward please rebase this change on top of the current tree.   The CI failures should be resolved.

PR #1024, opened 2022-11-16
CVE-2022-42898: lib/krb5: fix _krb5_get_int64 on systems where 'unsigned long' is just 32-bit

metze-samba:
See: https://bugzilla.samba.org/show_bug.cgi?id=15203


jaltman:
We will take https://github.com/heimdal/heimdal/pull/1025 instead since the commit message provides additional details.
Thank you for the submission.

PR #1027, opened 2022-11-16
[7.7.1] lib/krb5: fix _krb5_get_int64 on 32-bit systems

jaltman:
On systems where 'unsigned long' is 32-bits and the 'size' parameter is set to 8 and the bytes are:

  0x78 0x00 0x00 0x00 0x00 0x00 0x00 0x00

When 'i' becomes 4 'v' will be 0 again. As 'unsigned long' is only able to hold 4 bytes.

Change the type of 'v' from 'unsigned long' to 'uint64_t' which matches the type of the output parameter 'value'.

(cherry picked from commit 9d1bfab9882d0aa14ae0981e6667c93db93ffc5d)

Reviewed-by: Ralph Boehme <slow@samba.org>
CVE: CVE-2022-42898
Samba-BUG: https://bugzilla.samba.org/show_bug.cgi?id=15203

### DIFF
--- a/lib/krb5/store-int.c
+++ b/lib/krb5/store-int.c
@@ -49,7 +49,7 @@ KRB5_LIB_FUNCTION krb5_ssize_t KRB5_LIB_CALL
 _krb5_get_int64(void *buffer, uint64_t *value, size_t size)
 {
     unsigned char *p = buffer;
-    unsigned long v = 0;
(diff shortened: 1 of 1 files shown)


jaltman:
The OSX CI failure is well understood.  NO_AFS is defined and unsigned but unused variables generate a fatal warning.

PR #1028, opened 2022-11-16
lib/krb5: krb5_pac_parse mem leak if pac_header_size failure

jaltman:
48 byte memory leak from krb5_pac_parse() each time pac_header_size() fails.

(cherry picked from commit 02f12fc746341f54a514e9e17bc7d315b91129e8)


jaltman:
The macOS CI failure is expected.

PR #1034, opened 2022-11-17
Only test Python in maintainer mode

juikim:
Python is not necessary for normal build.  Test it only in maintainer mode.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -18,7 +18,7 @@ dnl Checks for programs.
 AC_PROG_CC
 AM_PROG_CC_C_O
 AC_PROG_CPP
-AM_PATH_PYTHON
+AM_COND_IF([MAINTAINER_MODE], [AM_PATH_PYTHON])
 AC_CHECK_PROG(CLANG_FORMAT, clang-format, [clang-format], [no])
 test "$CLANG_FORMAT" = no && CLANG_FORMAT=true
 


abartlet:
If this is done, then all of 'make check' needs to be guarded on that as well.

nicowilliams:
I'm pretty sure we don't use Python at all outside `lib/wind`.  In `lib/wind` we use it only when building from not-a-dist-tarball, so maintainer mode isn't quite the condition we need to know whether we need Python.

If you build from a Git clone w/o `--enable-maintainer-mode`, does it work?  I expect it wouldn't.

There's also `lib/hx509/quote.py`, which we don't use _at all_ but which we dist -- I don't think we need that file for anything.

cschuber:
> I'm pretty sure we don't use Python at all outside `lib/wind`. In `lib/wind` we use it only when building from not-a-dist-tarball, so maintainer mode isn't quite the condition we need to know whether we need Python.
> 
> If you build from a Git clone w/o `--enable-maintainer-mode`, does it work? I expect it wouldn't.
> 
> There's also `lib/hx509/quote.py`, which we don't use _at all_ but which we dist -- I don't think we need that file for anything.

Python is used in lib/wind to generate bidi_table.h. The build fails without it.

nicowilliams:
> Python is used in lib/wind to generate bidi_table.h. The build fails without it.

Yup, so, either we commit those outputs or we require Python to build.  If requiring Python to build is too much for some, we should consider committing those outputs.

abartlet:
I really think that python3 is a reasonable expectation for a build system in 2023, and the effort to commit these files and keep them updated would be better spent elsewhere.

Can we just close this? @nicowilliams

PR #1036, opened 2022-11-22
CVE-2022-37966 kdc: Use server keys to select a session key

josephsutton1:
To select a session key etype, we should look at the supported etypes of the server principal rather than the client principal.

Samba BUG: https://bugzilla.samba.org/show_bug.cgi?id=15237

### DIFF
--- a/kdc/kerberos5.c
+++ b/kdc/kerberos5.c
@@ -2205,7 +2205,7 @@ _kdc_as_rep(astgs_request_t r)
      * intersection of the client's requested enctypes and the server's (like a
      * root krbtgt, but not necessarily) etypes from its HDB entry.
      */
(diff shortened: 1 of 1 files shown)


nicowilliams:
Hmm, LGTM but look at the test failing with ` kinit: krb5_get_init_creds: Preauth required but no preauth options send by KDC` -- what's that about?

abartlet:
> Hmm, LGTM but look at the test failing with ` kinit: krb5_get_init_creds: Preauth required but no preauth options send by KDC` -- what's that about?
> 
> EDIT: Ah, I think I know. I bet the correct fix is to make the flags argument `(is_tgs ? KFE_IS_TGS | KFE_USE_CLIENT:0)`.

@josephsutton1 can you take a look at this when you next get a chance?

PR #1038, opened 2022-11-22
hx509 ecdsa support

nicowilliams:
This PR adds support in libhx509 and hxtool for:

 - generating EC keys for use in ECDSA and EdDSA

 - signing certificates and CRLs with ECDSA and EdDSA keys

with support for the following curves:

 - CFRG Ed448 (but not for CMS -- we need to add lib/hcrypto support for `EVP_shake256()` for that)
 - CFRG Ed25519
 - NIST secp256r1
 - NIST secp384r1
 - NIST secp521r1

Notes:

 - We use `EVP_PKEY_CTX_set_ecdh_kdf_type(pctx, EVP_PKEY_ECDH_KDF_NONE)` in PKINIT.
 - We were using the equivalent of `EVP_PKEY_ECDH_KDF_NONE` prior to the OpenSSL 3.0 changes, therefore we must continue to in order to interop with older Heimdal releases.  MIT does not seem to support ECDH for PKINIT.
   If PKINIT used CMS' `KeyAgreeRecipientInfo`, _then_ we would be obligated to use the X9.63 KDF, but PKINIT does not.
 - MIT Kerberos does not seem to support EDCH for PKINIT.

TODO:

 - `make check-valgrind`
 - run `tests/kdc/check-kdc` and `tests/kdc/check-pkinit` modified to run clients and the KDC built with OpenSSL 3.0 and OpenSSL 1.1 to check that Heimdal can interop with itself

TODO later:

 - ~~Test OCSP signing and checking.~~ (we don't have an OCSP responder implementation, though we could use OpenSSL's for testing our client)
 - Research making PKINIT able to use X25519 and X448 for key agreement.

### DIFF
--- a/lib/hx509/crypto-ec.c
+++ b/lib/hx509/crypto-ec.c
@@ -97,6 +97,7 @@ static struct nid2oid_st {
     int key_type;
 } nid2oid[] = {
 #ifdef HAVE_HCRYPTO_W_OPENSSL
+#ifdef HAVE_OPENSSL_30
 #ifdef NID_ED448
     { ASN1_OID_ID_ED448, ASN1_OID_ID_ED448, ASN1_OID_ID_X448,
         "ED448", "X448", NID_X448, NID_ED448, EVP_PKEY_ED448 },
@@ -119,6 +120,7 @@ static struct nid2oid_st {
     { ASN1_OID_ID_ECDSA_WITH_SHA256, ASN1_OID_ID_EC_GROUP_SECP256R1,
         ASN1_OID_ID_EC_GROUP_SECP256R1, "P-256", "P-256", NID_X9_62_prime256v1,
         NID_ecdsa_with_SHA256, EVP_PKEY_EC },
+#endif
 #endif
 #ifdef NID_ecdsa_with_SHA1
     { ASN1_OID_ID_ECDSA_WITH_SHA1, ASN1_OID_ID_EC_GROUP_SECP160R2,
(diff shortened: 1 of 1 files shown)


nicowilliams:
This is failing with:

```
/home/runner/work/heimdal/heimdal/lib/krb5/pkinit-ec.c:152:54: error: comparison between pointer and integer ('void *' and 'int') [-Werror,-Wpointer-integer-compare]
    if (ret == 0 && EC_KEY_generate_key(ctx->u.eckey != 1))
                                        ~~~~~~~~~~~~ ^  ~
/home/runner/work/heimdal/heimdal/lib/krb5/pkinit-ec.c:152:41: error: incompatible integer to pointer conversion passing 'int' to parameter of type 'EC_KEY *' (aka 'struct ec_key_st *') [-Werror,-Wint-conversion]
    if (ret == 0 && EC_KEY_generate_key(ctx->u.eckey != 1))
                                        ^~~~~~~~~~~~~~~~~
/usr/include/openssl/ec.h:936:33: note: passing argument to parameter 'key' here
int EC_KEY_generate_key(EC_KEY *key);
```

which means I'm breaking building with OpenSSL 1.1.

nicowilliams:
Besides the build issues on OS X and Windows, and leaks to fix, there's also sending the wrong `namedCurve` in PKINIT.  Until we implement key agreement algorithm negotiation for PKINIT, I must stick to X9.62.

PR #1041, opened 2022-11-22
OpenSSL 3.0 support backport for Heimdal 7.9

jaltman:


jaltman:
Note that there is no current tester that builds against OpenSSL 3.0.   

A variety of tests currently fail when built against OpenSSL 3.0.

cf/crypto.m4
```diff
@@ -134,6 +134,14 @@ if test "$with_openssl" != "no"; then
...
 fi
 
+if test "$openssl" = "yes"; then
+    AC_CHECK_LIB([crypto],
+                 [OSSL_EC_curve_nid2name],
+                 [AC_DEFINE_UNQUOTED([HAVE_OPENSSL_30], 1,
+                                     [whether OpenSSL is 3.0 or higher])]
+                 )
+fi
+
```
vdukhovni, line 153: Just compile a C program with `#if OPENSSL_VERSION_PREREQ(3,0)`.

lib/hx509/crypto-ec.c
```diff
@@ -127,6 +138,39 @@ parse_ECParameters(hx509_context context,
...
+    if ((&digest_alg->algorithm == &asn1_oid_id_sha256 ||
+         der_heim_oid_cmp(&digest_alg->algorithm, &asn1_oid_id_sha256) == 0))
+        return EVP_sha256();
+    if ((&digest_alg->algorithm == &asn1_oid_id_secsig_sha_1 ||
+         der_heim_oid_cmp(&digest_alg->algorithm, &asn1_oid_id_secsig_sha_1) == 0))
+        return EVP_sha1();
+    if ((&digest_alg->algorithm == &asn1_oid_id_rsa_digest_md5 ||
+         der_heim_oid_cmp(&digest_alg->algorithm,
+                          &asn1_oid_id_rsa_digest_md5) == 0))
+        return EVP_md5();
```
vdukhovni, line 161: This case potentially needs an `# ifndef OPENSSL_NO_MD5` guard. OpenSSL can be built with no MD5 support (perhaps rarely done).

lib/hx509/crypto-ec.c
```diff
@@ -140,6 +184,106 @@ ecdsa_verify_signature(hx509_context context,
...
+                               /* XXX Include the OID in the message */
+                               "Unsupported subjectPublicKey algorithm");
+    if (ret == 0)
+        ret = ECParameters2nid(context, spi->algorithm.parameters, &groupnid);
+    if (ret == 0 && (curve_sn = OBJ_nid2sn(groupnid)) == NULL)
+        hx509_set_error_string(context, 0,
+                               ret = HX509_CRYPTO_SIG_INVALID_FORMAT,
+                               "Could not resolve curve NID %d to its short name",
+                               groupnid);
+    if (ret == 0 && (curve_sn_dup = strdup(curve_sn)) == NULL)
```
vdukhovni, line 217: This is not needed.

lib/hx509/crypto-ec.c
```diff
@@ -140,6 +184,106 @@ ecdsa_verify_signature(hx509_context context,
...
+     * OSSL_DECODER(3).
+     */
+    if (ret == 0) {
+        /*
+         * Apparently there's no error checking to be done here?  Why does
+         * OSSL_PARAM_construct_utf8_string() want a non-const for the value?
+         * Is that a bug in OpenSSL?
+         */
+        params[0] = OSSL_PARAM_construct_utf8_string(OSSL_PKEY_PARAM_GROUP_NAME,
+                                                     curve_sn_dup, 0);
```
vdukhovni, line 237: This is a prototype nuisance, the string is not mutated, and you can pass constant strings (casting away const if need be).

lib/hx509/crypto-ec.c
```diff
@@ -140,6 +184,106 @@ ecdsa_verify_signature(hx509_context context,
...
+                               ret = HX509_CRYPTO_SIG_INVALID_FORMAT,
+                               "Could not set up to parse key for curve %s",
+                               curve_sn);
+
+    /* Finally we can decode the subjectPublicKey */
+    p = spi->subjectPublicKey.data;
+    len = spi->subjectPublicKey.length / 8;
+    if (ret == 0 &&
+        (public = d2i_PublicKey(EVP_PKEY_EC, &template, &p, len)) == NULL)
+        ret = HX509_CRYPTO_SIG_INVALID_FORMAT;
```
vdukhovni, line 258: That's an awful lot of work, instead of just calling `d2i_PUBKEY()`.

cf/crypto.m4
nicowilliams: We need a HAVE_OPENSSL_30 for places where we can't include OpenSSL headers.  Yes, a better autoconf macro would be nice.

lib/hx509/crypto-ec.c
nicowilliams: The `strdup()`?

lib/hx509/crypto-ec.c
nicowilliams: Got it.

lib/hx509/crypto-ec.c
nicowilliams: In my newest commits I do.  See PR #1038.

PR #1044, opened 2022-12-06
bx509d /get-tgts partial authz

nicowilliams:
This PR allows the /get-tgts end-point to handle partial authorization, where a client principal is allowed to get TGTs for some other principals and not others.

Changes:

 - `kdc/simple_csr_authorizer.c` is no more; its functionality has moved into `kdc/test_csr_authorizer.c`
 - `kdc/test_csr_authorizer.c` now has an option to run as an IPC CSR authorizer daemon
 - `kdc/bx509d.c` and `kdc/ipc_csr_authorizer.c` are updated to support partial authorization for `/get-tgts`

@jheiss we should do a walk-through and code review.


kdc/bx509d.c
```diff
@@ -1855,6 +1858,7 @@ authorize_TGT_REQ(struct bx509_request_desc *r)
         ret = kdc_authorize_csr(r->context, "get-tgt", r->req, p);
     krb5_free_principal(r->context, p);
     hx509_request_free(&r->req);
+    r->req = NULL;
```
jheiss, line 1861: Per Nico this is unnecessary, but harmless, since hx509_request_free nulls out req.

jheiss:
I've reviewed this and don't have any further comments. It doesn't look like I can formally approve it, not sure if that's because of the OS X build failure or because I lack permissions. But it has my stamp of approval.

nicowilliams:
Thanks @jheiss!

PR #1047, opened 2022-12-14
Found an error - NULL passed to function and dereferenced there.

le-cha-sever:
General error description:
After having been assigned to NULL value at NAME_FILE, pointer '0' is passed as 1st parameter in call to function 'NAME_FUNCTION', where it is dereferenced.
Solution:
Option 1:   I have declared a local variable. I pass its address to the function NAME_FUNCTION.
Option 2:   I am passing the address of the minor_status variable. This variable is commonly used for this purpose.

### DIFF
--- a/lib/gssapi/krb5/acquire_cred.c
+++ b/lib/gssapi/krb5/acquire_cred.c
@@ -674,7 +674,7 @@ OM_uint32 GSSAPI_CALLCONV _gsskrb5_acquire_cred_from
 				    NULL, time_rec, NULL, actual_mechs);
     if (ret != GSS_S_COMPLETE) {
 	if (handle->mechanisms != NULL)
-	    gss_release_oid_set(NULL, &handle->mechanisms);
+	    gss_release_oid_set(minor_status, &handle->mechanisms);
 	HEIMDAL_MUTEX_destroy(&handle->cred_id_mutex);
 	krb5_free_principal(context, handle->principal);
 	free(handle);
--- a/lib/gssapi/krb5/add_cred.c
+++ b/lib/gssapi/krb5/add_cred.c
@@ -238,6 +238,7 @@ OM_uint32 GSSAPI_CALLCONV _gsskrb5_add_cred_from (
 
 failure:
     if (handle) {
+	OM_uint32 tmp_status;
 	if (handle->principal)
 	    krb5_free_principal(context, handle->principal);
 	if (handle->keytab)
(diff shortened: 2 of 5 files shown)


jaltman:
Thank you for the report but this change is both incorrect because NULL is a valid parameter to pass and because advertisements for the Svace static analyzer are not acceptable in the Heimdal repository history.

le-cha-sever:
I removed advertisements.
Answer to a phrase "NULL is a valid parameter to pass":
1)File lib/gssapi/krb5/add_cred.c line 249 has its original form:
gss_release_oid_set(NULL, &handle->mechanisms);
Then in function gss_release_oid_set in the file lib/gssapi/mech/gss_release_oid_set.c the line 36 will take the next  transformation:
	*minor_status = 0;   ==> *NULL = 0; 
This expression is not defined by the standard.
2) I made a demo at www.onlinegdb.com
It is project with original code https://onlinegdb.com/Qxg3F9Pka
And this is project with modification code https://onlinegdb.com/CNxOP8BRP
In the second case, the program works correctly - it prints both lines:
hello first
hello second

lib/gssapi/krb5/export_sec_context.c
```diff
@@ -247,7 +247,7 @@ _gsskrb5_export_sec_context(
     ret = _gsskrb5_delete_sec_context (minor_status, context_handle,
 				       GSS_C_NO_BUFFER);
     if (ret != GSS_S_COMPLETE)
-	_gss_secure_release_buffer (&minor, interprocess_token);
+	_gsskrb5_release_buffer (minor_status, interprocess_token);
```
nicowilliams, line 250: This change should not be included as it clobbers the possibly-useful value of `*minor_status`.

lib/roken/snprintf-test.c
```diff
@@ -223,7 +223,8 @@ cmp_with_sprintf_float (void)
 static int
 test_null (void)
 {
-    return rk_snprintf (NULL, 0, "foo") != 3;
+    char tmp[256];
+    return rk_snprintf (tmp, sizeof(tmp), "foo") != 3;
```
nicowilliams, line 227: This change should be excluded.  It's perfectly sensible to call `snprintf(NULL, 0, ...)`.

nicowilliams:
Thank you for this PR!  Two of the changes included should be dropped (see other comments).  It is true, all GSS functions that take a `OM_uint32 *minor_status` should never have `NULL` passed in as a minor status pointer!

lib/gssapi/krb5/export_sec_context.c
lhoward: Also, you want to zero it.

le-cha-sever:
Sorry
I misunderstood the logic of how github works.
Is there a way to undo my changes?

le-cha-sever:
I removed two changes

nicowilliams:
I've force-pushed to your branch a rebase of your commits with a single commit and a more appropriate commit subject line.

If you agree, I'll merge this!  Thanks again for your contributions!

le-cha-sever:
I agree. Thanks

PR #1050, opened 2022-12-19
Found an error in file lib/krb5/send_to_kdc.c.

le-cha-sever:
**General error description**:
Variable 'host->fd', which may receive negative value at send_to_kdc.c:499 by calling function 'host_connected' at send_to_kdc.c:758, is used at send_to_kdc.c:403 by calling function 'host_dead' at send_to_kdc.c:772.
**Solution**:
I modified the test conditions on line 760 - I added the expression host->state != DEAD to the condition via the && operation.

### DIFF
--- a/lib/krb5/send_to_kdc.c
+++ b/lib/krb5/send_to_kdc.c
@@ -757,7 +757,7 @@ eval_host_state(krb5_context context,
     if (host->state == CONNECTING && writeable)
 	host_connected(context, ctx, host);
 
-    if (readable) {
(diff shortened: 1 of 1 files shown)


jaltman:
Thank you for the report but advertisements for the Svace analyzer are unacceptable in Heimdal commit history.   Please update the pull request to remove the advertisements.

elric1:
Looks like b02d4d4569 solved this issue in a different way. Thanks for your report!

PR #1054, opened 2022-12-29
only announce enabled pa types

metze-samba:
It confuses clients if the kdc announces support for a pa-type, e.g. KRB5_PADATA_FX_FAST and later ignore/reject it


abartlet (approved):

nicowilliams:
LGTM.  Thanks!

PR #1061, opened 2023-01-08
roken: ROKEN_xxx_ATTRIBUTE macros

lhoward:
Add ROKEN_xxx_ATTRIBUTE macros, derived from krb5-types.h, to aid compiling with compilers that don't have __attribute__ defined.

Not sure if this might break things with cross-compiling though (see the `awk` script changes)


nicowilliams:
LGTM!

jaltman:
cross-compilation does not actually work in Heimdal.   Working cross-compilation requires that a build host version of tools and supporting libraries (including libroken) which are separate and independent of the versions needed for the target platform.   There is no infrastructure to do that at the present time.

jaltman (approved):

PR #1062, opened 2023-01-09
krb5/store_stdio.c: workaround for solaris10/hpux/aix fread/fwrite duplication

manner82:
Observed this issue on solaris10, hpux, aix.
What happens is that an "fwrite following an fread" writes extra characters to the file (probably the read buffer), so the file becomes corrupt. The workaround I have found is to do a seek to the current position before the write. This patch fixes the unit test of store_stdio.

### DIFF
--- a/lib/krb5/store_stdio.c
+++ b/lib/krb5/store_stdio.c
@@ -83,6 +83,29 @@ stdio_store(krb5_storage * sp, const void *data, size_t size)
     ssize_t count;
     size_t rem = size;
 
+#if __sun || _AIX || __hpux
+    /* On solaris10, an fwrite following an fread causes the last character to be duplicated.
+        Example program to reproduce the issue:
+        #include <stdio.h>
+
+    int main() {
+            FILE *f = fopen("my", "w+");  // create a one byte file, containing 'A'
+            char a = 'A';
+            fwrite(&a, 1, 1, f);
+            fclose(f);
+
+            f = fopen("my", "r+");
+            char v;
+            fread(&v, 1, 1, f);    // Read the 'A' character
+            // fseeko(f, 0, SEEK_CUR); // -> this is a workaround, seek to where we are already
+            char b = 'B';
+            fwrite(&b, 1, 1, f);   // -> the file content becomes "AAB", despite we only write 'B'
+            fclose(f);
+            return 0;
+    }
+     */
+    fseeko(F(sp), 0, SEEK_CUR);
+#endif
     /* similar pattern to net_write() to support pipes */
     while (rem > 0) {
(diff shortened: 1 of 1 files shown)


manner82:
Since the issue is weird, I leave here an easy way to try it: solaris can run on x86_64, you can install it with vagrant:

```
# download and and enter into a solaris 11.4 virtualbox machine:
vagrant init cprince/solaris11_4
vagrant up
vagrant ssh

# add a working gcc compiler (run these commands inside the ssh session):
sudo pkgadd -d http://get.opencsw.org/now
export PATH=$PATH:/opt/csw/bin
pkgutil -U
pkgutil -y -i gcc5core

# create and compile the example program:
cat test.c
    #include <stdio.h>

    int main() {
            FILE *f = fopen("my", "w+");  // create a one byte file, containing 'A'
            char a = 'A';
            fwrite(&a, 1, 1, f);
            fclose(f);

            f = fopen("my", "r+");
            char v;
            fread(&v, 1, 1, f);    // The read
            // fseeko(f, 0, SEEK_CUR); // -> this is a workaround, seek to where we are already
            char b = 'B';
            fwrite(&b, 1, 1, f);   // -> the file content becomes "AAB"
            fclose(f);
            return 0;
    }

gcc -o test test.c

# run and check the result:
./test && cat my
    AAB
```

nicowilliams:
Thanks!

nicowilliams:
https://stackoverflow.com/questions/23843587/fwrite-doesnt-work-directly-after-fread

nicowilliams:
Hmm, but where were we using `"r+"`, `"a+"`, or `"w+"` that would break like this?

manner82:
Thanks for the quick actions, I really appreciate it!

So it seems this bug is ANSI C standard. :) Even present in fopen manual[1], altough on my system it is only mentioned as "good practice". I am quite amazed by this, sounds like documenting a bug instead of fixing it. However this does not seem to cause any trouble on linux, macos and freebsd.

```
[1] When a file is opened with update mode ( '+' as the second or third character in the mode argument), both input and output may be performed on the associated stream. However, the application shall ensure that output is not directly followed by input without an intervening call to fflush() or to a file positioning function fseek(), fsetpos(), or rewind(), and input is not directly followed by output without an intervening call to a file positioning function, unless the input operation encounters end-of-file.
```

lhoward:
If it’s good practice then let’s do it unconditionally?

nicowilliams:
> If it’s good practice then let’s do it unconditionally?

Already done with 829f07eed30 `krb5: Always fseek before fwrite in storage_stdio` :)

PR #1064, opened 2023-01-09
base: Add atomic CAS macros/functions

nicowilliams:


lib/base/heimbase-atomics.h
```diff
@@ -151,6 +164,10 @@ heim_base_exchange_64(uint64_t *p, uint64_t newval)
     return val;
 }
 
+#define heim_base_cas_pointer(t,e,d) 	compare_and_swaplp((atomic_l)(t), (e), (d))
+#define heim_base_cas_32(t,e,d)		compare_and_swap  ((atomic_l)(t), (e), (d))
```
lhoward, line 168: Believe this should be `atomic_p` for 32-bit (see `heim_base_atomic_dec_32()`)

lhoward: https://www.ibm.com/docs/en/aix/7.2?topic=c-compare-swap-kernel-services

lib/base/heimbase-atomics.h
```diff
@@ -164,6 +181,10 @@ heim_base_exchange_64(uint64_t *p, uint64_t newval)
 #define heim_base_exchange_32(t,v)	((ULONG)InterlockedExchange((LONG volatile *)(t), (LONG)(v)))
 #define heim_base_exchange_64(t,v)	((ULONG64)InterlockedExchange64((LONG64 violatile *)(t), (LONG64)(v)))
 
+#define heim_base_cas_pointer(t,e,d) 	InterlockedCompareExchangePointer((atomic_l)(t), (d), (e))
+#define heim_base_cas_32(t,e,d)		InterlockedCompareExchange  ((atomic_l)(t), (d), (e))
+#define heim_base_cas_64(t,e,d)		InterlockedCompareExchange64((atomic_l)(t), (d), (e))
+
```
lhoward, line 187: Windows will want `PVOID volatile *`, `LONG volatile *`, `LONG64 volatile *` respectively. If they even need casts?

nicowilliams: Oops, good catch!

A test that uses these (for what?) would expose this.

nicowilliams:
I've reworked this a fair bit and tested `<stdatomic.h>` and GCC atomic builtins.  There is now also a test so we'll know that this works on Windows.  The test is not very good in that it doesn't race threads, but it should elicit compilation errors/warnings if `<heimdal-atomics.h>` has errors.

nicowilliams:
I stopped the UBSAN clang build to save runner time allocation.

PR #1065, opened 2023-01-09
Seqnum window (WIP)

nicowilliams:


nicowilliams:
I should use `heim_base_atomic_load()` for reading and not assume atomic reads.

PR #1066, opened 2023-01-10
hx509: Add certificates DIR: hashing support

nicowilliams:
This will make certificate verification faster.

TBD:

 - Add a test.  The test should create some CA certs, some EE certs, hash a DIR of trust anchors, and test validation.

   To hash the directory we'll have to use openssl(1) because we don't have that functionality in Heimdal yet.


nicowilliams:
Currently our `DIR:` iterates every file that's a certificate store.  We need a new `HASHDIR:` or something to only iterate hashed certs.

PR #1067, opened 2023-01-10
roken/parse_bytes: fix test for >= terabyte units on 32 bit systems

manner82:
On 32 bit systems, sizeof(ssize_t) and sizeof(unsigned long aka UL) is 32 bits which is not able to hold the value of a terabyte.

### DIFF
--- a/kcm/config.c
+++ b/kcm/config.c
@@ -36,6 +36,8 @@
 #include <getarg.h>
 #include <parse_bytes.h>
 
+#define MAX_REQUEST_MAX 67108864ll /* 64MB, the maximum accepted value of max_request */
+
 static const char *config_file;	/* location of kcm config file */
 
 size_t max_request = 0;		/* maximal size of a request */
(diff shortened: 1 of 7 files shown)


manner82:
For a test one can compile with ./configure --build=i686-pc-linux-gnu CFLAGS=-m32 CXXFLAGS=-m32 LDFLAGS=-m32

The big number scenarios in check-gen-template/test_default also fails, but that seems to be harder, I was not able to figure out why.

```
$ ./lib/asn1/check-gen-template
encoding of TESTDefault 0 failed 1859794436
encoding of TESTDefault 2 failed 1859794436
test_default() failed!
```

nicowilliams:
Thanks for this!  Maybe we just don't want to support parsing values too large for `size_t`/`ssize_t` on any platform, so maybe we should `#if` out those table entries that are too large on 32-bit systems.

nicowilliams:
OK, I think we'll take this, but can you change the call sites to `parse_bytes()` to use `int64_t` for the variable they assign the result to?  And where those callers are parsing allocation sizes, we should have some reasonable limit (e.g., 64MB).

manner82:
> Thanks for this! Maybe we just don't want to support parsing values too large for `size_t`/`ssize_t` on any platform, so maybe we should `#if` out those table entries that are too large on 32-bit systems.

I have added a commit for that:
asn1/check-gen.c: disable some non working bignum tests on 32 bit systems

nicowilliams:
Yes, this looks pretty good!  Thanks so much.

abartlet (approved):
This got a positive comment earlier from @nicowilliams should it be merged?

PR #1069, opened 2023-01-12
Update krb5.conf

ChristianBoehm:
I would like to introduce a comment which mentions the ability including either configuration files or directory's. The situation in some Linux distributions are somewhat unclear, in terms of having a default include directory or not in krb5.conf e.g. /etc/krb5.conf.d/ 
Having a comment like this, gives the package maintainers the ability to define a default include or handover the situation to the specific software, which expects this folder. Thanks!


nicowilliams:
Thanks!

PR #1070, opened 2023-01-13
Multi master iprop (WIP)

nicowilliams:
This PR adds multi-master iprop.  This is very much a work in progress, though the sketch is complete and a test exists (though it's not complete).  Do not use for now.

There is now a [`lib/kadm5/README.md`](https://github.com/heimdal/heimdal/blob/95a53b06cfe58d6326ee66768c425f441cc9a5ad/lib/kadm5/README.md) that documents details of the iprop protocol and theory of operation.

STATUS:

 - it mostly works
    - the mechanism for preventing endless circulation of incremental updates works
 - there is a test
    - the test is not complete:
       - it needs to test split-brain operation
       - it needs to test conflict resolution
 - not yet passed through valgrind, nor coverity for that matter
 - `scan-build make` clean
 - `make check-valgrind` clean (there's a few leaks of no consequence)

TODO BEFORE THIS IS READY:

 - Coverity.  (For reasons we generally want to wait until this lands before we run Coverity.)
 - Check that all uses of `local_context` and `server_context` in `lib/kadm5/ipropd_slave.c` are sensible
 - Check that `ipropd-slave` receiving incremental updates correctly locks and unlocks the HDB and local iprop log around each entry
 - Make `ipropd-slave` merging upstream full props unlock the local iprop log and HDB every N records (that way other local transactions can get a chance to go through while receiving a potentially large full prop)

Theory (see also the new `lib/kadm5/README.md`):

 - we add to `HDB_entry` (and, therefore, to iprop log records) an indication of
    - origin KDC,
    - time of write,
    - list of KDCs that have seen this entry
 - `ipropd-slave` will drop any iprop records it receives where the included `HDB_entry`'s iprop information lists the receiving KDC as having seen the record
 - `ipropd-slave` will add the local KDC's hostname to the list of KDCs that have seen the received record before writing the record to the local iprop log
 - `ipropd-slave` will keep multiple iprop logs:
    - one representing the local HDB state, including locally-originated transactions as well as ones received from other KDCs
    - one uber-block-only log per-upstream KDC in which to record the iprop version information needed to catch up to that KDC in the future
 - we now include an `HDB_entry` in `kadm_delete` records so that those also have iprop information
 - with all of that we can:
    - prevent infinite looping of iprop records because every record will have enough information that `ipropd-slave` can decide whether to accept and reply a record or drop it
    - perform a modicum of conflict resolution
 - full props received in multi-master mode are merged in as if the send-complete was just a very large incremental update, thus if a KDC gets cut off from the rest for a long time and iprop logs get truncated, the realm will recover as one would expect 

In multi-master mode `iprod-slave` no longer continuously holds a lock on the local HDB and iprop log, thus no longer excluding local transactions.

Any iprop topology should work provided that:

 - every KDC that runs `ipropd-master` also runs `ipropd-slave`
 - there is a path from every KDC that runs `ipropd-master` to every other KDC that runs `ipropd-master`

Thus this PR also adds support for hierarchical iprop.

Note that not every KDC must run `ipropd-slave` in multi-master mode.  Thus one could run one or two masters per-site, and one or two "connections" between sites in a circular, tree, or star topology, with most KDCs being secondaries.

Currently the configuration of which KDCs follow which KDCs is static (one `ipropd-slave` invocation per upstream KDC), not dynamic.  Eventually we could store configuration in the HDB and allow for dynamic reconfiguration of KDC iprop services.

NOTE: Of the iprop log entry payload formats for `create`, `modify`, `rename`, and `delete`, only `delete` was extensible before this PR, and now it will no longer be extensible.  That's because all iprop log entry payloads that carry an encoded `HDB_entry` carry it last and without a length field, so the length is always implied by the length of the whole payload minus things written before the encoded `HDB_entry`.

### DIFF
--- a/kadmin/init.c
+++ b/kadmin/init.c
@@ -137,6 +137,7 @@ init(struct init_options *opt, int argc, char **argv)
     int i;
     HDB *db;
     krb5_deltat max_life = 0, max_rlife = 0;
+    time_t now;
 
     if (!local_flag) {
 	krb5_warnx(context, "init is only available in local (-l) mode");
@@ -165,7 +166,8 @@ init(struct init_options *opt, int argc, char **argv)
 	krb5_warn(context, ret, "hdb_open");
 	return 1;
     }
-    ret = kadm5_log_reinit(kadm_handle, 0);
+    (void) krb5_us_timeofday(context, &now, NULL);
+    ret = kadm5_log_reinit(kadm_handle, 0, now);
     if (ret) {
         krb5_warn(context, ret, "Failed iprop log initialization");
         return 1;
--- a/kadmin/load.c
+++ b/kadmin/load.c
@@ -417,10 +417,12 @@ doit(const char *filename, int mergep)
     char *p;
     int lineno;
     int flags = O_RDWR;
+    time_t now;
     struct entry e;
     hdb_entry ent;
     HDB *db = _kadm5_s_get_db(kadm_handle);
 
+    (void) krb5_us_timeofday(context, &now, NULL);
     f = fopen(filename, "r");
     if (f == NULL) {
 	krb5_warn(context, errno, "fopen(%s)", filename);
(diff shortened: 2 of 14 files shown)


nicowilliams:
Ah, it's very close now.

PR #1071, opened 2023-01-13
Some portability fixes

manner82:
roken: provide a getline implementation from libedit if needed
spnego/context_storage: undef sc_flags (for hpux)
heimdal/asn1: do not throw error when trying to allocate 0 bytes of memory
heimbase-atomics.h: replace heim_base_atomic_barrier with syntax valid noop

### DIFF
--- a/lib/asn1/template.c
+++ b/lib/asn1/template.c
@@ -3015,7 +3015,7 @@ _asn1_copy(const struct asn1_template *t, const void *from, void *to)
 	    unsigned int i;
 
 	    tel->val = calloc(fel->len, ellen);
-	    if (tel->val == NULL)
+	    if (tel->val == NULL && fel->len > 0)
 		return ENOMEM;
 
 	    tel->len = fel->len;
--- a/lib/base/heimbase-atomics.h
+++ b/lib/base/heimbase-atomics.h
@@ -357,7 +357,7 @@ heim_base_cas_64(heim_base_atomic(uint64_t) *target, uint64_t expected,uint64_t 
 #endif
 
 #ifndef heim_base_atomic_barrier
-#define heim_base_atomic_barrier()
+static inline void heim_base_atomic_barrier(void) { return; }
 #endif
 
 #ifndef heim_base_atomic_load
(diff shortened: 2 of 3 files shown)


nicowilliams:
Ha, as it happens #1065 has a `getdelim()`/`getline()` cribbed from `lib/libedit` (with some changes): https://github.com/heimdal/heimdal/pull/1065/commits/dd5db177ff6253d02c89e2f5efbf8affe97e879e

lib/asn1/template.c
```diff
@@ -3015,7 +3015,7 @@ _asn1_copy(const struct asn1_template *t, const void *from, void *to)
 	    unsigned int i;
 
 	    tel->val = calloc(fel->len, ellen);
-	    if (tel->val == NULL)
+	    if (tel->val == NULL && fel->len * ellen > 0)
```
nicowilliams, line 3018: Let's avoid the potential for overflowing here and not multiply.  `ellen` will never be zero, so all we have to check for is that `fel->len` is not zero.

lib/base/heimbase-atomics.h
```diff
@@ -357,7 +357,7 @@ heim_base_cas_64(heim_base_atomic(uint64_t) *target, uint64_t expected,uint64_t
 #endif
 
 #ifndef heim_base_atomic_barrier
-#define heim_base_atomic_barrier()
+#define heim_base_atomic_barrier()  0
```
nicowilliams, line 360: Perhaps this should just be:

```
static inline void heim_base_atomic_barrier(void) { return; }
```

nicowilliams:
> This affects solaris and hpux builds, where getline is not present.

That's... really ancient Solaris.  Solaris 10 shipped with `getdelim()`/`getline()` 18 years ago.  Just FYI.  It doesn't change your need.

manner82:
> Ha, as it happens #1065 has a `getdelim()`/`getline()` cribbed from `lib/libedit` (with some changes): [dd5db17](https://github.com/heimdal/heimdal/commit/dd5db177ff6253d02c89e2f5efbf8affe97e879e)

Thanks, this looks a lot nicer. I have removed our patch in hope this one finds its way to the master instead. Or should I cherry-pick it?

lib/asn1/template.c
manner82: thanks, applied

lib/base/heimbase-atomics.h
manner82: thanks, it's a lot nicer, applied

nicowilliams:
Thanks.  LGTM now.

nicowilliams:
Thanks!

Issue #995, opened 2022-05-20
gss_unwrap_iov with CFX krb5 fails to unwrap if wrap token includes nonzero RRC


sthaber:
**Describe the bug**

There's really two bugs here, as there are two separate code paths affected.

First, unwrapping when not sealed (no encryption) is unimplemented as far as I can tell. Take a look at:

https://github.com/heimdal/heimdal/blob/master/lib/gssapi/krb5/cfx.c#L602
https://github.com/heimdal/heimdal/blob/master/lib/gssapi/krb5/cfx.c#L968

I believe this should be supported, as RFC4121 states:

```
The rotation count value is chosen by the sender based on
implementation details.  The receiver MUST be able to interpret all
possible rotation count values, including rotation counts greater
than the length of the token.
```

We've observed that, for example, Windows Server LDAP likes to send signed responses with a rotation applied. Furthermore, Heimdal's own `gss_wrap` implementation applies a rotation to signed messages. The IOV wrap equivalent does not, however! This means that if you take any of the unit tests in test_context.c and round a trip a `gss_wrap` followed by a `gss_unwrap_iov`, with sealing turned off, it will fail in the code path I linked to above.

The second bug is that unwrapping when sealed (with encryption) also seems to be broken. There exists a code path that is supposed to unrotate an IOV (`iov_unrotate()`, but it doesn't seem to work correctly. I haven't debugged exactly why, but the same set of circumstances described above (`gss_wrap` + `gss_unwrap_iov`) will produce the failure.

As a workaround, we've discovered it is sufficient to unrotate the buffers ourselves before passing them to the IOV unwrap in Heimdal. However, this is less than ideal, as it means we're interpreting RFC4121 data structures directly in our application, and we'd rather than Heimdal do this for us.

**To Reproduce**
1. Set up Kerberos using AES256/SHA1 as the encryption/checksumming type. This is standard if you use the Linux KDC.
2. Set up a GSS context using a Kerberos ticket from the KDC. (`gss_init_sec_context()` and `gss_accept_sec_context()`)
3. Wrap any message of any length using `gss_wrap()`. We used "abc", for example.
4. (Optional) Provide CONF_REQ_ENCRYPTION to seal the message when wrapping.
5. Take the wrapped message and attempt to unwrap it using `gss_unwrap_iov()`. This requires you set up the header, data, padding, and trailer appropriate per the IOV interface.
6. Observe a GSS error is returned.

**Expected behavior**
[...]

PR #1081, opened 2023-02-13
Implement `gss_get_mic_iov` and `gss_verify_mic_iov`

sthaber:
With some limitations:
- I couldn't for the life of me get the test fixture working at all. It's failing lots of stuff on master for me even without my changes. It's tested out our test environment and it's working fine but that's probably not very helpful for you. I'll include an example test that I want to include at the bottom.
- Only works for krb5 mechanism; errors for others.
- Only works for non-DCE style messages, if that's a thing for get/verify MIC. I assume it errors with DCE but didn't try b/c we don't use that in our environment.
- Probably only is actually zero copy for SHA1 -- this is also true of wrap/unwrap IOV variants today (SHA1 appears zero-copy, others do not). But should work for all checksum flavors.
- Defined the IOV interface myself to match our use case: `get` takes a `header`, `padding`, and `trailer` and returns a checksum in the `data` body. Verify takes a `data` body and errors if the verification fails. This might not match what other GSS libraries do for these methods with IOV parameters.
- Could probably share more code but I kept it isolated to make it easy to apply to various versions of the library.

Here's the start of a test I would write in your framework if I could get it working:
```
static void
getverifymic_iov(gss_ctx_id_t cctx, gss_ctx_id_t sctx, gss_OID mechoid)
{
	OM_uint32 min_stat, maj_stat;
	gss_iov_buffer_desc iov[4];
	int iov_len;

	char message_data[16] = "0123456789abcdef";

	memset(iov, 0, sizeof(iov));
	memset(&iov, 0, sizeof(iov));
	iov_len = sizeof(iov)/sizeof(iov[0]);

	iov[0].type = GSS_IOV_BUFFER_TYPE_HEADER | GSS_IOV_BUFFER_FLAG_ALLOCATE;
	iov[1].type = GSS_IOV_BUFFER_TYPE_PADDING | GSS_IOV_BUFFER_FLAG_ALLOCATE;
	iov[2].type = GSS_IOV_BUFFER_TYPE_TRAILER | GSS_IOV_BUFFER_FLAG_ALLOCATE;

	iov[3].type = GSS_IOV_BUFFER_TYPE_DATA;
	iov[3].buffer.value = message_data;
	iov[3].buffer.length = 16;

	maj_stat = gss_get_mic_iov(&min_stat, cctx, iov, iov_len);
	if (maj_stat != GSS_S_COMPLETE) {
		errx(1, "gss_get_mic_iov failed: %s", gssapi_err(maj_stat, min_stat, mechoid));
	}

	// XXX Steven: verify and assert, using both IOV and non-iov.
}
```

### DIFF
--- a/lib/gssapi/gssapi/gssapi.h
+++ b/lib/gssapi/gssapi/gssapi.h
@@ -956,6 +956,14 @@ GSSAPI_LIB_FUNCTION OM_uint32 GSSAPI_LIB_CALL
 gss_get_mic_iov(OM_uint32 *, gss_const_ctx_id_t, gss_iov_buffer_desc *, int);
 
 GSSAPI_LIB_FUNCTION OM_uint32 GSSAPI_LIB_CALL
+gss_verify_mic_iov(
+	OM_uint32 *,
+	gss_const_ctx_id_t,
+	gss_iov_buffer_desc *,
+	int,
+	const gss_buffer_t);
+
+GSSAPI_LIB_FUNCTION OM_uint32 GSSAPI_LIB_CALL
 gss_wrap_iov(OM_uint32 *, gss_ctx_id_t, int, gss_qop_t, int *,
 	     gss_iov_buffer_desc *, int);
 
--- a/lib/gssapi/gssapi/gssapi.h
+++ b/lib/gssapi/gssapi/gssapi.h
@@ -953,6 +953,9 @@ gss_decapsulate_token(gss_const_buffer_t /* input_token */,
  */
 
 GSSAPI_LIB_FUNCTION OM_uint32 GSSAPI_LIB_CALL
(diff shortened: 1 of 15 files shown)


sthaber:
From discussion on [this issue](https://github.com/heimdal/heimdal/issues/995) with @lhoward, sounded like my IOV work might be useful. If not, feel free to close. Thanks.

lhoward:
Thank you! I'll take a look when I have a moment!

lhoward:
@jaltman / @nicowilliams please don't merge until I've had a chance to review

sthaber:
Thanks. Yeah this is definitely a WIP so I for sure wouldn't merge it without some knowledgeable eyes on and possible unit tests in the library itself.

PR #1083, opened 2023-03-08
kdc-plugin: Split updating a PAC out of PAC verification

josephsutton1:
Up to now krb5plugin_kdc_pac_verify() has served to both verified and updated a PAC. There are cases, however, when we only want to retrieve and verify a PAC, but don't want to modify it. This is the case with the PAC from a FAST armor ticket.
    
Therefore, add a new plugin function, pac_update(), that will update a PAC obtained using pac_verify(). pac_verify() now only deals with verifying a PAC, while pac_update() handles any necessary updates to it.


josephsutton1:
I just addressed a flaw that I noticed in the `_kdc_pac_update()` API. I had callers pass in `is_trusted` to indicate whether or not the PAC in the TGT was trusted, but this still left us with no way to indicate whether the _armor_ PAC should be trusted. I'm now using a different method to convey this.

abartlet (approved):
This is an important part of what Samba AD needs for implementing AD Claims correctly in the PAC, and so getting to Windows 2012 FL support.

PR #1084, opened 2023-04-05
cf: Honour OpenSSL library and include paths

dankm:
Check for OpenSSL features while LDFLAGS and CFLAGS are set with the paths provided in the configure command line. This allows detecting a non-default OpenSSL's version correctly, such as on FreeBSD with one of the OpenSSL 3.0 ports.


nicowilliams:
Thanks!

Issue #790, opened 2021-07-01
Implicit declaration of functions in configure tests


ryandesign:
**Describe the bug**
Using the heimdal 7.5.0, 7.7.0, or even the latest code available in master today (2ee8834696f6d5c732f73005de78631005f1f527), some configure tests use functions without declaring them and/or without including the headers that declare them, resulting in incorrect determinations about system capabilities being made if the implicit declaration of functions is considered an error.

For example, some failing configure checks are:

```
checking for socket... no
checking for gethostbyname... no
```

From config.log, the reason for the failures is not that `socket` or `gethostbyname` don't exist (they do exist on macOS); you just forgot to include the right headers:

```
configure:20234: checking for socket
configure:20263: ccache /usr/bin/clang  -o conftest -pipe -Os -Werror=implicit-function-declaration -arch x86_64  -I/opt/local/include -D_LARGE_FILES= -L/opt/local/lib -Wl,-headerpad_max_install_names -arch x86_64 conftest.c      >&5
conftest.c:106:1: error: implicit declaration of function 'socket' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
socket()
^
1 error generated.
```
```
configure:20328: checking for gethostbyname
configure:20357: ccache /usr/bin/clang  -o conftest -pipe -Os -Werror=implicit-function-declaration -arch x86_64  -I/opt/local/include -D_LARGE_FILES= -L/opt/local/lib -Wl,-headerpad_max_install_names -arch x86_64 conftest.c      >&5
conftest.c:106:1: error: implicit declaration of function 'gethostbyname' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
gethostbyname()
^
1 error generated.
```

You need to `#include <sys/socket.h>` to find `socket` and `#include <netdb.h>` to find `gethostbyname`, etc.

Using `grep \ implicit` on my config.log, there are many such occurrences:

```
conftest.c:60:1: error: implicit declaration of function 'db_create' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
conftest.c:60:1: error: implicit declaration of function 'db_create' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
conftest.c:60:1: error: implicit declaration of function 'db_create' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
conftest.c:60:1: error: implicit declaration of function 'db_create' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
[...]

PR #1085, opened 2023-04-13
Avoid implicit function declarations in configure, for C99 compatibility

fweimer-rh:
This should fix issue #790.

Related to:

* https://fedoraproject.org/wiki/Changes/PortingToModernC
* https://fedoraproject.org/wiki/Toolchain/PortingToModernC

### DIFF
--- a/cf/find-func-no-libs2.m4
+++ b/cf/find-func-no-libs2.m4
@@ -21,7 +21,7 @@ if eval "test \"\$ac_cv_func_$1\" != yes" ; then
 		*) ac_lib="-l$ac_lib" ;;
 		esac
 		LIBS="$6 $ac_lib $5 $ac_save_LIBS"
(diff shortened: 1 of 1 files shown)


nicowilliams:
Ay, I suspect that there was a good reason that Heimdal has its `AC_FIND_FUNC_NO_LIBS` and `AC_FIND_FUNC_NO_LIBS2` m4 macros that take a headers parameter, but I'm inclined to take this.

abartlet:
A good way to help make folks feel like merging this would be to attach the config.h from before the change, before the change with the new compiler, and after the change, and show the diffs.  (At least, if this was my project - Samba - this is the proof I would require).

jaltman:
This change eliminates the -Wimplicit-function-declaration errors on Fedora 40 with gcc (GCC) 14.2.1 20240801 (Red Hat 14.2.1-1)

Issue #1091, opened 2023-05-17
Interop failures between Heimdal master and AD regarding PAC (code 0x3c)


nicowilliams:
When a cross-realm TGT is issued by a Heimdal KDC running `master` branch code with a target realm that is an Active Directory realm, the AD KDC will respond with a `KRB-ERROR` with a generic error and will log a `0x3c` code which is [described](https://learn.microsoft.com/en-us/windows/security/threat-protection/auditing/event-4769) variously as:

```
Group membership has overloaded the PAC.
Multiple recent password changes have not propagated.
Crypto subsystem error caused by running out of memory.
SPN too long.
SPN has too many parts.
```

Workaround: set `disable_pac = true` in the Heimdal KDC's `krb5.conf` in `[realms]` under the KDC's realm.

nicowilliams:
@iboukris @lhoward @josephsutton1 any clues?  Does this mean we should create a PAC large enough that the AD KDC can scribble SIDs onto it?

nicowilliams:
Or maybe it's a problem with FAST?

lhoward:
Not sure. @stevesyfuhs, the Heimdal KDC always issues PACs, even when not functioning as a Samba DC, in order to support constrained delegation and also to simplify the code paths for when it *is* functioning as a Samba DC.

This seems to be causing problems with cross-realm MIT trusts with Windows. Any ideas? Perhaps Windows isn't expecting this? The PACs Heimdal is issuing in this case omit `PAC_LOGON_INFO` but they do have `PAC_SERVER_CHECKSUM`, `PAC_PRIVSVR_CHECKSUM`, `PAC_LOGON_NAME`,`PAC_UPN_DNS_INFO`, `PAC_ATTRIBUTES_INFO`, and `PAC_TICKET_CHECKSUM` as far as I can tell from reading the code.

nicowilliams:
I've tried excluding `PAC_UPN_DNS_INFO`; that did not help.

josephsutton1:
Could it be because `PAC_LOGON_INFO` is missing? Samba at least requires that buffer to be present in the PACs it receives.

lhoward:
Quite possible. Windows seems to be OK without `PAC_LOGON_INFO` in the same domain (i.e. when pretending to be a RODC à la Azure AD or TktBridgeAP, including only `PAC_REQUESTOR_SID`). But likely the MIT cross-realm code paths are different in Windows.

[...]

PR #1092, opened 2023-05-18
kdc: Honor no-auth-data-reqd on cross-real TGTs

nicowilliams:
Nowadays we use PACs instead of AD-SIGNEDPATH, so we want a PAC on every TGT, but we don't necessarily want PACs on cross-realm TGTs.

Specifically, we don't interop well yet with AD when issuing cross-realm TGTs with AD realms as the destination realm (see #1091).


lhoward:
Sounds good but, does this mean you'll have to explicitly set that bit in the HDB for cross-realm TGSes? Or are you doing this implicitly in HDB? I guess ideally we want things just to work where Samba cross-realm TGTs have PACs included, and vanilla Kerberos realm TGTs don't.

We should understand any interaction with resource-based constrained delegation (RBCD) which I admit is not something I am across.

PR #1093, opened 2023-05-19
kdc: Pass ticket into PAC verification function

josephsutton1:
The plugin may want to verify the ticket is actually a TGT, and not a kpasswd ticket. To do this, it needs access to the ticket to check that its remaining lifetime is greater than two minutes.

### DIFF
--- a/kdc/kdc-plugin.c
+++ b/kdc/kdc-plugin.c
@@ -145,6 +145,7 @@ struct verify_uc {
     hdb_entry *client;
     hdb_entry *server;
     hdb_entry *krbtgt;
+    EncTicketPart *ticket;
     krb5_pac *pac;
 };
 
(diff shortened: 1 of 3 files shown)


lhoward:
Change the ABI, change the plugin version

lhoward:
`s/KRB5_PLUGIN_KDC_VERSION_10/KRB5_PLUGIN_KDC_VERSION_11/`

nicowilliams:
@josephsutton1 does the plugin need to know the `Ticket` envelope's `sname` as well?

nicowilliams:
Oh, I see, #1083 has the version bump.

josephsutton1:
> @josephsutton1 does the plugin need to know the `Ticket` envelope's `sname` as well?

No, I don’t believe so.

abartlet (approved):
How should we proceed with this and #1083 ?

josephsutton1:
> How should we proceed with this and #1083 ?

[lorikeet-heimdal-202305172147](https://git.samba.org/?p=lorikeet-heimdal.git;a=shortlog;h=refs/heads/lorikeet-heimdal-202305172147) should contain these two plugin changes arranged properly (but there this commit is named “kdc: Check lifetime of correct ticket”).

PR #1094, opened 2023-05-19
hdb: Add armor principal flag

josephsutton1:
This is useful to be able to tell whether a principal is being looked up just to verify a FAST armor ticket, which lets the HDB plugin avoid some unnecessary work.

### DIFF
--- a/kdc/fast.c
+++ b/kdc/fast.c
@@ -835,10 +835,9 @@ _kdc_free_fast_state(KDCFastState *state)
 }
 
 krb5_error_code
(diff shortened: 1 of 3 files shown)


lhoward:
LGTM

abartlet (approved):
This is (as you would expect) useful for Samba in implementing the AD DC, because while we want to apply authentication policies to client lookups, we won't want to apply them to the lookup of the armor principal, we just need the key to decrypt it. 

I've been hesitant with the historical Heimdal upstream membership rights, but would it be OK if I clicked 'merge' on this given the LGTM from @lhoward ?

PR #1096, opened 2023-05-19
kdc: Allow plugin to override KDC reply e-data

josephsutton1:
This is useful in order to produce custom error responses, as Windows does for NTSTATUS codes.


kdc/libkdc-exports.def
```diff
@@ -64,6 +64,7 @@ EXPORTS
 	kdc_request_set_reply_key
 	kdc_request_set_server_princ
 	kdc_request_set_sname
+	kdc_set_e_data
```
lhoward, line 67: Shouldn't this be `kdc_request_set_edata`

kdc/kerberos5.c
```diff
@@ -467,6 +467,27 @@ _kdc_set_e_text(astgs_request_t r, const char *fmt, ...)
     kdc_log(r->context, r->config, 4, "%s", e_text);
 }
 
+/*
+ * Override the e-data field to be returned in an error reply. The data will be
+ * owned by the KDC and eventually will be freed with krb5_data_free().
+ */
+krb5_error_code
+kdc_set_e_data(astgs_request_t r, heim_octet_string e_data)
```
lhoward, line 475: should be in `kdc-plugin.c`

lib/base/heimbase-svc.h
```diff
@@ -68,6 +68,7 @@
     char *cname;                                                \
     char *sname;                                                \
     const char *e_text;                                         \
+    heim_octet_string *e_data;                                  \
```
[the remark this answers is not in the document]
josephsutton1: I guess I thought it might be useful to be able to force the e-data field to be empty (by calling set_e_data() with zero-length data), but the way you suggest is probably better.

josephsutton1:
Rebased on master.

abartlet:
@nicowilliams It would be very helpful to get this merged, we think we have addressed all the concerns raised by @lhoward.  This change is quite important to Samba because if we don't get the error replies matching windows exactly we get clients retrying in a way that locks them out twice as fast under "bad password lockout".

abartlet:
The recent update by @josephsutton1 is related to https://gitlab.com/samba-team/samba/-/merge_requests/3377 - without that change Heimdal clients against a Windows DC that is imposing a Silo / authentication policy restriction gets a very confusing ` Miscellaneous failure (see text): Failed to decode METHOD-DATA (host/winclient@ADDOM.SAMBA.EXAMPLE.COM)`

lib/krb5/init_creds_pw.c
```diff
@@ -3144,19 +3144,36 @@ init_creds_step(krb5_context context,
...
+			_krb5_debug(context, 5, N_("Failed to decode METHOD-DATA", ""));
+		    }
+		} else if (len != ctx->error.e_data->length) {
+		    /* Trailing data — just ignore the error. */
+		    free_KERB_ERROR_DATA(&kerb_error_data);
+		} else {
+		    /* OK. */
+		    free_KERB_ERROR_DATA(&kerb_error_data);
 		}
 	    }
```
josephsutton1, line 3179: Just below this we have:
```
	    /*
	     * Unwrap KRB-ERROR, we are always calling this so that
	     * FAST can tell us if your peer KDC suddenly dropped FAST
	     * wrapping and its really an attacker's packet (or a bug
	     * in the KDC).
	     */
	    ret = _krb5_fast_unwrap_error(context, ctx->nonce, &ctx->fast_state,
					  &ctx->md, &ctx->error);
```

and a similar call in the get_cred_kdc() case.

The problem is that KERB-ERROR-DATA is not wrapped in FX-FAST, but `_krb5_fast_unwrap_error()` will notice if the reply is supposed to be FAST‐armored (and give the potentially confusing error message “FAST fast response is missing FX-FAST”).

Is it safe to bypass this call and treat a KERB-ERROR-DATA response as authoritative when we can’t guarantee that the KDC really issued the response?

josephsutton1:
Now Samba will produce error messages that are a bit more helpful:

`Miscellaneous failure (see text): KDC policy rejects request (NT status code 0xc0000413) (host/winclient@ADDOM.SAMBA.EXAMPLE.COM)`

josephsutton1:
Sigh, MSVC appears to violate the C standard by supporting `uint32_t` but not `PRIx32`.

josephsutton1:
Instead, MSVC forces us to use `"%I32x"` instead.

abartlet:
That could be put into a define in roken if we wanted.

lib/krb5/init_creds_pw.c
```diff
@@ -3173,7 +3173,31 @@ init_creds_step(krb5_context context,
...
+
+		    /* log the failure */
+		    if (_krb5_have_debug(context, 5)) {
+			const char *str = krb5_get_error_message(context, ret);
+			_krb5_debug(context, 5, "krb5_get_init_creds: KRB-ERROR %d/%s", ret, str);
+			krb5_free_error_message(context, str);
+		    }
+
 		    free_KERB_ERROR_DATA(&error_data);
+		    goto out;
```
josephsutton1, line 3200: This won’t work. We need to continue to the `ret == KRB5KDC_ERR_KEY_EXP && ctx->runflags.change_password == 0 && ctx->runflags.change_password_prompt` case so that the user can be prompted to change their password if it has expired.

lib/krb5/init_creds_pw.c
josephsutton1: I checked the FAST RFC, which says:

“If the Kerberos FAST padata is included in the request but not included in the error reply, it is a matter of the local policy on the client to accept the information in the error message without integrity protection. However, the client SHOULD process the KDC errors as the result of the KDC's inability to accept the AP_REQ armor and potentially retry another request with a different armor when applicable” — that’s not the case here. “The Kerberos client MAY process an error message without a PA-FX-FAST-REPLY, if that is only intended to return better error information to the application, typically for trouble-shooting purposes.”

https://datatracker.ietf.org/doc/html/rfc6113#section-5.4.4

jsutton24:
> We should include this in 8.0. Can you rebase?

This needs a bit more work before it’s ready to be merged (see https://github.com/heimdal/heimdal/pull/1096#pullrequestreview-1731080418).

nicowilliams:
@abartlet @jsutton24 can you tell me which PRs you need to be included in a Heimdal 8.0 release?  I'd like to make a release soon!

lib/krb5/get_cred.c
```diff
@@ -719,12 +719,54 @@ get_cred_kdc(krb5_context context,
...
 
 	if (rep.error.e_data) {
-	    ret = decode_METHOD_DATA(rep.error.e_data->data,
-				     rep.error.e_data->length,
-				     &md, NULL);
+	    KERB_ERROR_DATA error_data;
+
+	    memset(&error_data, 0, sizeof(error_data));
+
+	    /* First try to decode the e-data as KERB-ERROR-DATA. */
```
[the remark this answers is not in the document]
jsutton24: I’m not sure if there’s any particular rhyme or reason to how Windows decides what to return.

Issue #1100, opened 2023-05-25
Missing -d argument to yacc for header file in lib/asn1, lib/hx509


riastradh:
**Describe the bug**
lib/asn1/Makefile and lib/hx509/Makefile invoke yacc without the -d flag, but expect the header file, asn1parse.h and sel-gram.h, to be created.  The -d flag is required to generate a header file, so the build fails after yacc has run with:

```
../../../../master/lib/asn1/lex.l:50:10: fatal error: asn1parse.h: No such file or directory
 #include "asn1parse.h"
          ^~~~~~~~~~~~~
```

**To Reproduce**
./configure && make, with BSD yacc (e.g., /usr/bin/yacc on NetBSD 9)

**Expected behavior**
successful build

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - OS: NetBSD
 - Browser: N/A
 - Version 9

**Smartphone (please complete the following information):**
N/A

**Additional context**
N/A

PR #1101, opened 2023-05-25
Pass -d to yacc(1) so it generates the header file.

riastradh:
Bison must generate this unsolicited, but BSD yacc does not.

The makefiles for lib/com_err and lib/sl already did this; this change just adds it to lib/asn1 and lib/hx509 too.

fix https://github.com/heimdal/heimdal/issues/1100


nicowilliams:
Ah, let's see if that works with bison.

nicowilliams:
Thanks!

PR #1116, opened 2023-05-26
if a CSR has multiple extRequests, keep only last

mcr:
The code already keeps the last entry, but it leaks the previous copies.  @nicowilliams noticed this.


nicowilliams:
Which of these options is best in this case:

- keep the first
- keep the last (but don't leak)
- error out

?

Issue #1139, opened 2023-05-29
check-gssmask test deadlocks reliably


riastradh:
**Describe the bug**
```
$ ps -A -l -d -www
  UID   PID  PPID    CPU PRI NI      VSZ    RSS WCHAN   STAT TTY          TIME COMMAND
...
 1000 17863  9364    476  80 10    22344   1884 wait    IN+  pts/152   0:00.00 | |                         `-- /bin/sh ./check-gssmask
 1000  7023 17863      0  80 10    59592   5948 netio   SN+  pts/152   0:00.13 | |                           |-- /home/riastradh/crypto/krb/heimdal/build/x86_64--netbsd9/appl/gssmask/.libs/gssmask -p 8889 --spn=host/n1.test.h5l.se@TEST.H5L.SE --logfile=n1.log --moniker=n1
 1000  7303 17863    742  80 10    59580   5884 netio   IN+  pts/152   0:00.08 | |                           |-- /home/riastradh/crypto/krb/heimdal/build/x86_64--netbsd9/appl/gssmask/.libs/gssmask -p 8891 --spn=host/n3.test.h5l.se@TEST.H5L.SE --logfile=n3.log --moniker=n3
 1000 19943 17863    866  80 10    59584   5884 netio   IN+  pts/152   0:00.07 | |                           |-- /home/riastradh/crypto/krb/heimdal/build/x86_64--netbsd9/appl/gssmask/.libs/gssmask -p 8890 --spn=host/n2.test.h5l.se@TEST.H5L.SE --logfile=n2.log --moniker=n2
 1000 20196 17863      0  80 10    59312   4308 netio   SN+  pts/152   0:00.02 | |                           `-- /home/riastradh/crypto/krb/heimdal/build/x86_64--netbsd9/appl/gssmask/.libs/gssmaestro --slaves=localhost:8889 --slaves=localhost:8890 --slaves=localhost:8891 --principals=user1@TEST.H5L.SE:u1
...
```

All of these processes are single-threaded and they all appear to be waiting for reads.

If I hit ^C, it leaves a kdc process running, which causes subsequent test runs to spuriously fail because of address-already-in-use.  The line https://github.com/heimdal/heimdal/blob/9cbafd7320b8023ea8c6ea89643803bd021adda5/tests/gss/check-gssmask.in#L99 should perhaps have HUP INT TERM after EXIT.

**To Reproduce**
[...]

Issue #1133, opened 2023-05-28
Suspicious non-const static buffer in krb5 get_in_tkt.c set_ptypes


riastradh:
**Describe the bug**
https://github.com/heimdal/heimdal/blob/9cbafd7320b8023ea8c6ea89643803bd021adda5/lib/krb5/get_in_tkt.c#L316-L319
https://github.com/heimdal/heimdal/blob/9cbafd7320b8023ea8c6ea89643803bd021adda5/lib/krb5/get_in_tkt.c#L321
https://github.com/heimdal/heimdal/blob/9cbafd7320b8023ea8c6ea89643803bd021adda5/lib/krb5/get_in_tkt.c#L337

Unclear if there's anything actually wrong with this but it's suspicious to return a non-const pointer to a static buffer to a caller.

**To Reproduce**
code inspection

nicowilliams:
Indeed, this is a bug, though in a deprecated function's code path.

Issue #1132, opened 2023-05-28
Suspicious non-const static buffer in krb5_digest_request


riastradh:
**Describe the bug**
https://github.com/heimdal/heimdal/blob/9cbafd7320b8023ea8c6ea89643803bd021adda5/lib/krb5/digest.c#L689-L692

Doesn't compile with const added, unclear whether this is a problem, just leaving this here to come back to later.

**To Reproduce**
code inspection

**Expected behavior**
const static data, or non-static data, or thread synchronization

Issue #1131, opened 2023-05-28
Thread-unsafe logic in krb5_checksum_disable, krb5_enctype_disable, krb5_enctype_enable, and krb5_allow_weak_crypto


riastradh:
**Describe the bug**
The functions krb5_checksum_disable, krb5_enctype_disable, and krb5_allow_weak_crypto read/modify/write global data in struct _krb5_checksum_type and struct _krb5_encryption_type objects.

And, despite having krb5_context arguments, their effects aren't limited to the context.

**To Reproduce**
code inspection

**Expected behavior**
Since the list of supported checksum and encryption types is statically determined, krb5_context could instead have a small private bit mask of which ones are enabled.

nicowilliams:
We should make sure this is thread-safe if we want to keep the global behavior.  We probably don't need this to not be global behavior, but there's a much easier way to deal with this: just have a setting in the `krb5_context` and check that when looking up cryptographic algorithms.

(This is interesting because something very similar happens in OpenSSL, but their approach to making this thread-safe killed performance in OpenSSL 3.0.)

PR #1140, opened 2023-05-29
Sprinkle const throughout libkrb5

riastradh:
This cuts down on a large amount of .data in libkrb5.so, moving it to .rodata, and, more importantly, substantially reduces the number of non-const globals that must be audited for thread safety.  I tracked all this down by iteratively running:

```
nm --defined-only lib/krb5/.libs/libkrb5.a \
| awk '$1 ~ /:/ { f = $1 } $2 ~ /[bBdD]/ { if (f) { printf("\n"); print f; f = "" }; print }'
```

to print all .data and .bss symbols and what file they came from, until I eliminated every unnecessarily mutable global I could, with some exceptions I didn't understand on cursory inspection (notably https://github.com/heimdal/heimdal/issues/1132 and https://github.com/heimdal/heimdal/issues/1133).

One more commit along these lines in my tree is queued up, to constify the checksum/encryption/string-to-key type structures.  However, it is blocked on https://github.com/heimdal/heimdal/issues/1131, so I'll do it separately.

All tests except check-gssmask pass with this on NetBSD 9.  (check-gssmask hangs, but it hung already without these changes, so that's not a regression: https://github.com/heimdal/heimdal/issues/1139)

I did not examine the other libraries than libkrb5.  My focus is on making it safe for clients to use this.  Next step would be libgssapi, which, from the nm | awk incantation, looks like it too has a lot of unnecessary mutable globals.

### DIFF
--- a/lib/krb5/constants.c
+++ b/lib/krb5/constants.c
@@ -35,7 +35,7 @@
 
 #include "krb5_locl.h"
 
-KRB5_LIB_VARIABLE const char *krb5_config_file =
+KRB5_LIB_VARIABLE const char *const krb5_config_file =
 #ifdef KRB5_DEFAULT_CONFIG_FILE
 KRB5_DEFAULT_CONFIG_FILE
 #else
@@ -56,12 +56,12 @@ SYSCONFDIR "/krb5.conf" PATH_SEP
 #endif /* KRB5_DEFAULT_CONFIG_FILE */
 ;
 
-KRB5_LIB_VARIABLE const char *krb5_defkeyname = KEYTAB_DEFAULT;
+KRB5_LIB_VARIABLE const char *const krb5_defkeyname = KEYTAB_DEFAULT;
 
-KRB5_LIB_VARIABLE const char *krb5_cc_type_api = "API";
-KRB5_LIB_VARIABLE const char *krb5_cc_type_file = "FILE";
-KRB5_LIB_VARIABLE const char *krb5_cc_type_memory = "MEMORY";
-KRB5_LIB_VARIABLE const char *krb5_cc_type_kcm = "KCM";
-KRB5_LIB_VARIABLE const char *krb5_cc_type_scc = "SCC";
-KRB5_LIB_VARIABLE const char *krb5_cc_type_dcc = "DIR";
-KRB5_LIB_VARIABLE const char *krb5_cc_type_keyring = "KEYRING";
+KRB5_LIB_VARIABLE const char *const krb5_cc_type_api = "API";
+KRB5_LIB_VARIABLE const char *const krb5_cc_type_file = "FILE";
+KRB5_LIB_VARIABLE const char *const krb5_cc_type_memory = "MEMORY";
+KRB5_LIB_VARIABLE const char *const krb5_cc_type_kcm = "KCM";
+KRB5_LIB_VARIABLE const char *const krb5_cc_type_scc = "SCC";
+KRB5_LIB_VARIABLE const char *const krb5_cc_type_dcc = "DIR";
+KRB5_LIB_VARIABLE const char *const krb5_cc_type_keyring = "KEYRING";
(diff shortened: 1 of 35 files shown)


nicowilliams:
image

nicowilliams:
And we need to fix our OS X build in GHA :(

nicowilliams:
Note that when you go through the logs of failing Windows builds the error messages at the very end may not be useful, and you may have to search for `error` in the middle of the logs (which won't load in the browser UI unless you scroll through it all, but you can download the logs).

abartlet:
I dropped this patch into Samba's build and it still builds, which is good.

I realise this needs to be a one-step-at-a-time thing, but it would be awesome if we could address the remaining const warnings, perhaps in a new PR.

abartlet:
These may not show up in a default Heimdal build, but in Samba (which imports Heimdal, and from where my interest comes) we compile with compile flags including -Wdiscarded-qualifiers -Wcast-qual and -Werror.

We actually have to set
```
heimdal_no_error_flags = ['-Wno-error=discarded-qualifiers',
                          '-Wno-error=cast-qual',
                          '-Wno-error=missing-field-initializers',
                          '-Wno-error=shadow',
                          '-Wno-error=implicit-fallthrough',
                          '-Wno-error=enum-compare',
                          '-Wno-error=unused-but-set-variable',
                          '-Wno-error=unused-const-variable',
                          '-Wno-error=unused-variable',
                          '-Wno-error=unused-result',
                          '-Wno-error=old-style-definition']
```
to mask the impact of -Werror and allow a build of Heimdal in Samba.  

Now none of this is 'your problem', but I would suggest that a good way to ensure that your const data changes are complete (they tickle down to all the callers etc) is to set strict compiler flags in Heimdal as far as possible, and still have it build.

lib/krb5/pac.c
```diff
@@ -2032,8 +2032,8 @@ _krb5_pac_get_attributes_info(krb5_context context,
     return 0;
 }
 
-static unsigned char single_zero = '\0';
-static krb5_data single_zero_pac = { 1, &single_zero };
+static const unsigned char single_zero = '\0';
+static const krb5_data single_zero_pac = { 1, rk_UNCONST(&single_zero) };
```
abartlet, line 2036: shouldn't this be krb5_const_data?

lib/krb5/constants.c
```diff
@@ -56,12 +56,12 @@ SYSCONFDIR "/krb5.conf" PATH_SEP
...
+KRB5_LIB_VARIABLE const char *const krb5_defkeyname = KEYTAB_DEFAULT;
 
-KRB5_LIB_VARIABLE const char *krb5_cc_type_api = "API";
-KRB5_LIB_VARIABLE const char *krb5_cc_type_file = "FILE";
-KRB5_LIB_VARIABLE const char *krb5_cc_type_memory = "MEMORY";
-KRB5_LIB_VARIABLE const char *krb5_cc_type_kcm = "KCM";
-KRB5_LIB_VARIABLE const char *krb5_cc_type_scc = "SCC";
-KRB5_LIB_VARIABLE const char *krb5_cc_type_dcc = "DIR";
-KRB5_LIB_VARIABLE const char *krb5_cc_type_keyring = "KEYRING";
+KRB5_LIB_VARIABLE const char *const krb5_cc_type_api = "API";
```
abartlet, line 61: Are "" strings not const by default anyway?

abartlet (commented):
This is pretty good, but see my comments. 

I'm confused by the need to add another 'const' to "" strings.  I thought the compiler did that, eg gcc 4.0 removed -fwritable-strings https://gcc.gnu.org/gcc-4.0/changes.html

lib/krb5/constants.c
abartlet: Ahh, that makes sense.

lib/krb5/pac.c
abartlet: Sorry, there are so many other krb5_const types these days I simply assumed!  

This change seems correct then.

nicowilliams:
Thanks!

Issue #1154, opened 2023-06-06
Simple option to prevent all DNS queries


riastradh:
**Is your feature request related to a problem? Please describe.**
Various code paths in Heimdal rely on DNS lookups -- for domain-to-realm mapping, for realm-to-KDC mapping, for KDC/kadmind/kpasswdd hostname address lookups.  These are all potential privacy leaks, especially once Heimdal can make all connections to KDC/kadmind/kpasswdd through a [SOCKS or HTTP CONNECT proxy](https://github.com/heimdal/heimdal/issues/1151).

**Describe the solution you'd like**
A single `[libdefaults] prevent_dns = true` option, or `[libdefaults] dns_allowed = false`, that systematically disables all DNS lookups in Heimdal -- both in tools like kinit and in libkrb5, libgssapi, &c., so anything using Heimdal won't lead to DNS lookups within Heimdal.  All paths to getaddrinfo or gethostbyname or equivalent should be gated on this option.  Obviously DNS lookups would remain enabled by default, but a single krb5.conf option should be able to prevent them once and for all.

**Describe alternatives you've considered**
There are some options for disabling DNS for individual purposes:
 - `[libdefaults] dns_lookup_kdc = false` for the realm -> KDC mapping
 - `[libdeafults] dns_lookup_realm = false` for the domain -> realm mapping
 - `[libdefaults] name_canon_rules = as-is:` to plug the [CNAME-chasing vulnerability](https://github.com/heimdal/heimdal/issues/1130)

However, none of them systematically inhibits the use of DNS, and even if you track down all the individual options defined right now, Heimdal might sprout new ways to leak user activity through DNS queries that aren't covered by those options.

**Additional context**
Relevant to https://github.com/heimdal/heimdal/issues/1151 to ensure no privacy leaks through DNS when connections are made via hostnames, not addresses, in SOCKS or HTTP CONNECT.

Issue #1151, opened 2023-06-05
SOCKS and/or HTTP CONNECT proxy


riastradh:
**Is your feature request related to a problem? Please describe.**
kinit, kgetcred, gss_init_sec_context, and applications that use these via libkrb5 or libgssapi or equivalent all do DNS lookups and talk in the clear directly to a KDC over the internet.

This leaks two things:
1. the realm you're logging into and the principal you're logging in as, to anyone on the network between you and the KDC, such as eavesdroppers sitting across the coffee shop; and
2. your location (IP address), to the KDC.

**Describe the solution you'd like**
krb5.conf should have an option for connecting to the KDC via TCP through a SOCKS proxy and/or an HTTP CONNECT proxy, leaving all name resolution to the proxy.

This way, all these connections can be done through Tor, for any application that is able to use Tor for connections to the service, or through an ssh -D forward.  (Of course, it requires domain-to-realm and realm-to-KDC mapping to be hard-coded in the local krb5.conf, and DNS mapping to be disabled.)

**Describe alternatives you've considered**
Heimdal currently has an option `http_proxy` but it is only used for KDCs reached through HTTP anyway, which I believe is off by default.

Heimdal currently has an option `dns_proxy` but (a) it only covers the DNS queries (like the service name canonicalization which [shouldn't be happening anyway](https://github.com/heimdal/heimdal/issues/1130), domain-to-realm mapping, and KDC SRV/A/AAAA records), not communication with the KDC itself, and (b) it is a bizarre protocol that came out of nowhere and probably isn't very useful.

**Additional context**
[Portrait of Heimdall apparently wearing some kind of socks, with a chicken on his head](https://en.wikipedia.org/wiki/File:Heimdal_(1907)_by_J._T._Lundbye.jpg)

lhoward:
IAKERB would be better, no?Sent from my iPhoneOn 6 Jun 2023, at 08:53, riastradh ***@***.***> wrote:﻿
Is your feature request related to a problem? Please describe.
kinit, kgetcred, gss_init_sec_context, and applications that use these via libkrb5 or libgssapi or equivalent all do DNS lookups and talk in the clear directly to a KDC over the internet.
This leaks two things:

the realm you're logging into and the principal you're logging in as, to anyone on the network between you and the KDC, such as eavesdroppers sitting across the coffee shop; and
your location (IP address), to the KDC.

Describe the solution you'd like
[...]

PR #1155, opened 2023-06-09
New option [libdefaults] block_dns = true to block all DNS queries.

riastradh:
fix https://github.com/heimdal/heimdal/issues/1154

(draft pending automatic tests)


nicowilliams:
@riastradh yes: `lib/krb5/test_cc.c`, which uses `krb5_set_config()` to set a configuration as if it'd been read from a file.

kadmin/kadm_conn.c
```diff
@@ -246,6 +246,11 @@ start_server(krb5_context contextp, const char *port_str)
 	hints.ai_flags    = AI_PASSIVE;
 	hints.ai_socktype = SOCK_STREAM;
 
+	if (krb5_config_get_bool(context, NULL, "libdefaults", "block_dns",
+		NULL)) {
+	    hints.ai_flags &= ~AI_CANONNAME;
```
nicowilliams, line 251: There is no `getaddrinfo()` flag for disabling DNS lookups.  Is this change specifically for NetBSD?

kdc/altsecid_gss_preauth_authorizer.c
```diff
@@ -167,6 +167,13 @@ ad_connect(krb5_context context,
     } *s, *servers = NULL;
     size_t i, num_servers = 0;
 
+    if (krb5_config_get_bool(context, NULL, "libdefaults", "block_dns",
+	    NULL)) {
+	ret = KRB5KDC_ERR_SVC_UNAVAILABLE;
+        krb5_set_error_message(context, ret, "DNS blocked when finding AD DC");
+	return ret;
+    }
+
```
nicowilliams, line 176: @lhoward do you want a config param here for specifying LDAP servers to use?

lib/krb5/get_addrs.c
```diff
@@ -50,6 +50,14 @@ gethostname_fallback (krb5_context context, krb5_addresses *res)
     char hostname[MAXHOSTNAMELEN];
     struct hostent *hostent;
 
+    if (krb5_config_get_bool(context, NULL, "libdefaults", "block_dns",
```
nicowilliams, line 53: Is this needed here?  I expect in many cases there will be an `/etc/hosts` entry for the local host's name.

lib/krb5/krbhst.c
```diff
@@ -718,6 +728,13 @@ plugin_get_hosts(krb5_context context,
 {
     struct plctx ctx = { type, kd, 0 };
 
+    /*
+     * XXX Need a way to pass this through -- unsure if any of this is
```
nicowilliams, line 732: The plugins get the `krb5_context`, so they get to check this parameter.  I say let's trust them.

nicowilliams:
I'm not sure that trying to get `getaddrinfo()` and `gethostbyname()` to not use DNS is a worthy goal.  We can stop at not doing `SRV` and `TXT` (and, someday, `URI`) resource record lookups.

nicowilliams:
> Why would getaddrinfo with AI_NUMERICHOST and no AI_CANONNAME ever do DNS lookups? Are there any serious implementations that do more than just parse the IP address (i.e., inet_pton) in that case?

If you don't pass an IP to `getaddrinfo()` I still expect it to look up the given name's addresses.

> > We can stop at not doing SRV and TXT (and, someday, URI) resource record lookups.

> No, that defeats the purpose -- my goal for https://github.com/heimdal/heimdal/issues/1154 is to guarantee that Heimdal leaks nothing into DNS queries. This is critical for location privacy. (Not sufficient -- that will also require https://github.com/heimdal/heimdal/issues/1151 -- but necessary.)

What I'm saying is that is that leaving out `AI_CANONNAME` and adding `AI_NUMERICHOST` and `AI_NUMERICSERV` is not enough: you have to also make sure that the "name" passed to `getaddrinfo()` is an IP address.  Granted, anyone who configures `block_dns = true` probably would also configure KDCs in `/etc/krb5.conf` using only IPs.  But maybe you're right that `inet_pton()` is the right option here.

nicowilliams:
OK, thanks, that resolves my `getaddrinfo()` questions.

nicowilliams:
Is this ready?  It's marked draft.

nicowilliams:
To test this sort of thing we'd really need to use containers in the tests (unless already running in a container) so we can start a DNS server.

nicowilliams:
> Not really -- a good test would be to make some krb5_* or GSS-API calls in an executable program that defines the getaddrinfo symbol to call abort() if it doesn't have AI_NUMERICHOST|AI_NUMERICSERV set (and to call inet_pton otherwise to implement the usual functionality).
> 
> It's not quite the same audit as ktracing and filtering for networking syscalls, but it should work on any elven platform.

I'm not sure that symbol interposition will work reliably on all platforms (Windows/MSYS/Cygwin?), but, yes, it's an idea.

The fact that we generally can't rely on DNS servers on non-standard ports when using the system resolver means we eventually have to aim for a container-ish test scheme.  An alternative would be to have our own resolver (which is a scary prospect, though it would have a number of benefits, but mainly it would be a tremendous time suck when we already don't have the time for it).

nicowilliams:
Maybe we can have a very simple interposition-based test after all.  If interposition fails then no problem: the test passes, and if interposition succeeds then the test will work.  And we can add a test that interposition does work so that we can know from the test logs.

appl/test/auditdns.c
```diff
@@ -0,0 +1,96 @@
...
+
+    if (hints == NULL ||
+	(hints->ai_flags & AI_NUMERICHOST) == 0 ||
+	(hints->ai_flags & AI_CANONNAME) != 0) {
+	fprintf(stderr, "DNS leak: %s %s:%s\n",
+	    __func__, hostname, servname);
+	abort();
+    }
+
+    if ((sym = dlsym(RTLD_NEXT, __func__)) == NULL) {
```
nicowilliams, line 87: Excellent.  My only question is whether we should `#ifdef` everything out here in static builds.

nicowilliams:
@riastradh it's been merged.  Arguably `tests/kdc/` would have been a better place for this test since you're not modifying GSS code, but that's ok.

lib/krb5/krbhst.c
```diff
@@ -795,7 +795,9 @@ kdc_get_next(krb5_context context,
 	return KRB5_KDC_UNREACH;
     }
 
-    if(context->srv_lookup) {
+    if (krb5_config_get_bool(context, NULL, "libdefaults", "block_dns",
+	    NULL) &&
+	context->srv_lookup) {
```
jsutton24, line 800: Why are we running the following code only if `block_dns` is `true`? Correct me if I’m wrong, but shouldn’t these conditions be the other way round?

jsutton24: I’ll submit a PR to fix this.

PR #1159, opened 2023-06-19
Some miscellaneous fixes

josephsutton1:
Just a collection of minor fixes that I had lying around.

### DIFF
--- a/kdc/fast.c
+++ b/kdc/fast.c
@@ -799,8 +799,10 @@ _kdc_fast_strengthen_reply_key(astgs_request_t r)
 
 	ret = krb5_generate_random_keyblock(r->context, r->reply_key.keytype,
 					    &r->strengthen_key);
-	if (ret)
-	    krb5_abortx(r->context, "random generator fail");
+	if (ret) {
+	    kdc_log(r->context, r->config, 0, "failed to prepare random keyblock");
+	    return ret;
+	}
 
 	ret = _krb5_fast_cf2(r->context,
 			     &r->strengthen_key, "strengthenkey",
--- a/lib/krb5/changepw.c
+++ b/lib/krb5/changepw.c
@@ -695,7 +695,7 @@ krb5_change_password (krb5_context	context,
 		      int		*result_code,
 		      krb5_data		*result_code_string,
 		      krb5_data		*result_string)
-    KRB5_DEPRECATED_FUNCTION("Use X instead")
+    KRB5_DEPRECATED_FUNCTION("Use krb5_set_password instead")
 {
     struct kpwd_proc *p = find_chpw_proto("change password");
 
--- a/kdc/kerberos5.c
+++ b/kdc/kerberos5.c
@@ -2760,7 +2760,7 @@ out:
 				 r->rep.padata,
 			         r->armor_crypto,
 			         &req->req_body,
-			         r->error_code,
+			         r->error_code ? r->error_code : ret,
 			         r->client_princ,
 			         r->server_princ,
 			         NULL, NULL,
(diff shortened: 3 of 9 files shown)


kdc/fast.c
```diff
@@ -799,8 +799,10 @@ _kdc_fast_strengthen_reply_key(astgs_request_t r)
 
 	ret = krb5_generate_random_keyblock(r->context, r->reply_key.keytype,
 					    &r->strengthen_key);
-	if (ret)
-	    krb5_abortx(r->context, "random generator fail");
+	if (ret) {
+	    kdc_log(r->context, r->config, 0, "random generator fail");
+	    return ret;
+	}
```
abartlet, line 805: Should we say why it failed - given the reason we are not abort()ing is that it can fail for reasons unrelated to the RNG?

abartlet (commented):
Other than clarifying further why the RNG failed after stopping it aborting, I'm happy with these.

lib/krb5/changepw.c
```diff
@@ -695,7 +695,7 @@ krb5_change_password (krb5_context	context,
 		      int		*result_code,
 		      krb5_data		*result_code_string,
 		      krb5_data		*result_string)
-    KRB5_DEPRECATED_FUNCTION("Use X instead")
+    KRB5_DEPRECATED_FUNCTION("Use krb5_set_password instead")
```
abartlet, line 698: I've carefully checked all the Use X instead changes match the comments.

abartlet (approved):

kdc/fast.c
```diff
@@ -799,8 +799,10 @@ _kdc_fast_strengthen_reply_key(astgs_request_t r)
 
 	ret = krb5_generate_random_keyblock(r->context, r->reply_key.keytype,
 					    &r->strengthen_key);
-	if (ret)
-	    krb5_abortx(r->context, "random generator fail");
```
nicowilliams, line 803: I'm of the opinion that the KDC should definitely `abort()` if there's no RNG.  The RNG shouldn't fail though.

lib/base/dict.c
```diff
@@ -52,8 +52,10 @@ dict_dealloc(void *ptr)
 {
     heim_dict_t dict = ptr;
     struct hashentry **h, *g, *i;
+    size_t j;
 
-    for (h = dict->tab; h < &dict->tab[dict->size]; ++h) {
```
nicowilliams, line 56: `&dict->tab[dict->size]` is perfectly safe and defined behavior.

kdc/fast.c
nicowilliams: `krb5_generate_random_block()` does `krb5_abortx()` if `krb5_generate_random()` fails, and `krb5_generate_random_keyblock()` calls `krb5_generate_random_block()`.  We should rationalize this a bit.

nicowilliams: `krb5_generate_random_block()` is `void`-returning.  We should deprecate it and not use it in-tree.

nicowilliams:
I think [f3ad53c1d](https://github.com/heimdal/heimdal/pull/1159/commits/f3ad53c1d125372da265685f715ba8990dbe30aa) is not needed but not harmful, so I'm inclined to just take it, though if you have time to remove that commit that'd be great.

josephsutton1:
> I think [f3ad53c1d](https://github.com/heimdal/heimdal/pull/1159/commits/f3ad53c1d125372da265685f715ba8990dbe30aa) is not needed but not harmful, so I'm inclined to just take it, though if you have time to remove that commit that'd be great.

OK, removed.

nicowilliams:
Thanks!

PR #1160, opened 2023-06-19
kadmin/check: move error messages to stderr

svmhdvn:
This helps with certain automation tools, such as ansible, that usually expect failure to be visible in stderr.

Other minor changes:
* replace "doesn't" with "does not" to keep messages consistent and avoid escaping quotes in grep, etc.
* add newlines


nicowilliams:
Thanks!

Issue #1164, opened 2023-06-20
Undefined behaviour in `ktutil add`


riastradh:
**Describe the bug**
https://github.com/heimdal/heimdal/blob/da9cad2047b13d781afb7789d79599b5f3218bfa/admin/add.c#L177-L190

1. If `malloc` succeeds, then on entry to line 187, `p` and `res` are both (valid) pointers into the object allocated by it.
2. After at least one loop iteration so that `p == end` is possible, if `realloc` succeeds on line 187, then object allocated in (1) by `malloc` no longer exists, and any pointers into it are indeterminate under C99 Sec. 6.2.4 paragraph 2:
   > The value of a pointer becomes indeterminate when the object it points to reaches the end of its lifetime.
3. On line 189, both `p` and `res` are indeterminate, which under C99 3.17.2 may be trap representations, arithmetic on which is undefined behaviour.

**To Reproduce**
gcc12 with `-Wuse-after-free`

**Expected behavior**
no need to run to the corner store to buy holy kleenex any time I run ktutil

nicowilliams:
Oy, that's a typo.  It should have been `alloced >> 1`, not `alloced > 1`.  That's a pretty funny typo!

PR #1165, opened 2023-06-20
ktutil: Avoid even arithmetic on res after realloc(res, ...).

riastradh:
Under C99, Sec 6.2.4, paragraph 2:

> The value of a pointer becomes indeterminate when the object it points to reaches the end of its lifetime.

'Indeterminate' (3.17.2) includes a trap representation, and any reference to a trap representation is undefined behaviour.  Thus, after realloc(res, ...) succeeds, any reference to res (or p) is undefined behaviour.

So, instead of using `p - res` after res has been freed, use the existing name for the value we know it has now: len.  (We could also use alloced because p == end in this branch, and end = res + alloced, and p = res + len.  Of course, we would have to move it up a line to before we update alloced to have a different value.)

fix https://github.com/heimdal/heimdal/issues/1164

### DIFF
--- a/admin/add.c
+++ b/admin/add.c
@@ -187,7 +187,7 @@ read_file(FILE *f)
             if ((tmp = realloc(res, alloced + (alloced > 1))) == NULL)
                 err(1, "Out of memory");
             alloced += alloced > 1;
-            p = tmp + (p - res);
+            p = tmp + len;
             res = tmp;
             end = res + alloced;
         }


admin/add.c
```diff
@@ -186,8 +186,8 @@ read_file(FILE *f)
 
             if ((tmp = realloc(res, alloced + (alloced > 1))) == NULL)
                 err(1, "Out of memory");
+            p = tmp + alloced;
```
abartlet, line 189: This really doesn't look right.  Shouldn't it be `p = tmp + len`?

abartlet (changes requested):

admin/add.c
abartlet: Yes, it would be much clearer, as we can clearly see len incrementing on line 195

nicowilliams:
Thanks!

Issue #1105, opened 2023-05-26
gssmask.c build fails with -Werror=format-truncation


riastradh:
**Describe the bug**
```
../../../../master/appl/gssmask/gssmask.c: In function ‘handleGetVersionAndCapabilities’:
../../../../master/appl/gssmask/gssmask.c:659:39: error: ‘%s’ directive output may be truncated writing up to 255 bytes into a region of size between 0 and 255 [-Werror=format-truncation=]
      snprintf(name, sizeof(name), "%s-%s-%s",
                                       ^~
        ut.sysname, ut.version, ut.machine);
                    ~~
../../../../master/appl/gssmask/gssmask.c:659:6: note: ‘snprintf’ output between 3 and 768 bytes into a destination of size 256
      snprintf(name, sizeof(name), "%s-%s-%s",
      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
        ut.sysname, ut.version, ut.machine);
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Note: On NetBSD, each member of struct utsname is 256 bytes long.

**To Reproduce**
```
./configure --enable-developer
make
```

**Expected behavior**
successful build

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - OS: NetBSD
 - Browser: N/A
 - Version: 9
 - Compiler: gcc 7.4.0
 
**Smartphone (please complete the following information):**
N/A

**Additional context**
N/A

Issue #1102, opened 2023-05-25
unused-but-set-variable key_data in hdb-mitdb.c


riastradh:
**Describe the bug**
The variable `key_data` in the `mdb_seq` function in lib/hdb/hdb-mitdb.c is assigned, but unused, triggering gcc's `-Wunused-but-set-variable` warning.  With `--enable-developer`, this warning is fatal and the build fails.

```
  CC       hdb-mitdb.lo
../../../../master/lib/hdb/hdb-mitdb.c: In function 'mdb_seq':
../../../../master/lib/hdb/hdb-mitdb.c:772:15: error: variable 'key_data' set but not used [-Werror=unused-but-set-variable]
     krb5_data key_data, data;
               ^~~~~~~~
```

It's unclear to me what the intent of this variable is: should it be used, or can it be discarded altogether?  The code came in a lump all at once in 2009, so there's no revision history here to check for obvious mistakes, and I'm not familiar enough with this code to know what it's trying to do.

For now, to make progress, I've added `(void)key_data` to the function to suppress the warning.

**To Reproduce**
```
./configure --enable-developer && make
```

**Expected behavior**
successful build

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - OS: NetBSD
 - Browser: N/A
 - Version: 9

**Smartphone (please complete the following information):**
N/A

**Additional context**
N/A

PR #1166, opened 2023-06-21
Fix some compiler warnings.

riastradh:
fix https://github.com/heimdal/heimdal/issues/1102
fix https://github.com/heimdal/heimdal/issues/1105

### DIFF
--- a/appl/gssmask/gssmask.c
+++ b/appl/gssmask/gssmask.c
@@ -646,7 +646,7 @@ static int
 HandleOP(GetVersionAndCapabilities)
 {
     int32_t cap = HAS_MONIKER;
-    char name[256] = "unknown", *str;
+    char *name = NULL, *str = NULL;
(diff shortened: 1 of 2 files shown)


appl/gssmask/gssmask.c
```diff
@@ -656,20 +656,24 @@ HandleOP(GetVersionAndCapabilities)
     {
 	struct utsname ut;
 	if (uname(&ut) == 0) {
-	    snprintf(name, sizeof(name), "%s-%s-%s",
-		     ut.sysname, ut.version, ut.machine);
+	    if (asprintf(&name, "%s-%s-%s",
+		    ut.sysname, ut.version, ut.machine) == -1) {
```
nicowilliams, line 660: As I understand it one has to check whether `asprintf()` returns `-1` or the variable it was expected to set is null.

Issue #1171, opened 2023-06-21
Makefile.am for lib/asn1, lib/com_err, and lib_sl sets YFLAGS instead of AM_YFLAGS 


riastradh:
**Describe the bug**
Makefile.am is supposed to set AM_YFLAGS and reserve YFLAGS for the user, according to [the automake manual](https://www.gnu.org/software/automake/manual/html_node/Flag-Variables-Ordering.html).

However, in master as of 9569ee1a3b4ba8fa142b07e2bad150f0681916db, only lib/hx509/Makefile.am correctly sets AM_YFLAGS instead of YFLAGS:
```
$ git grep YFLAGS
lib/asn1/Makefile.am:YFLAGS = -d -o asn1parse.c -t
lib/com_err/Makefile.am:YFLAGS = -d -o parse.c
lib/hx509/Makefile.am:AM_YFLAGS = -d -o sel-gram.c
lib/sl/Makefile.am:YFLAGS = -d -o slc-gram.c
```

**To Reproduce**
code inspection
run make with extra options in YFLAGS

**Expected behavior**
Makefile.am only sets AM_YFLAGS, not YFLAGS

**Additional context**
I have not checked for abuse of user variables other than YFLAGS, which are listed in [the automake manual](https://www.gnu.org/software/automake/manual/html_node/Flag-Variables-Ordering.html).

PR #1173, opened 2023-06-21
Makefile.am: Set AM_YFLAGS and AM_LFLAGS, not YFLAGS or LFLAGS.

riastradh:
YFLAGS and LFLAGS are reserved for the user to set, not for makefiles to set:
https://www.gnu.org/software/automake/manual/html_node/Flag-Variables-Ordering.html

fix https://github.com/heimdal/heimdal/issues/1171


nicowilliams:
Thanks!

PR #1175, opened 2023-06-22
Add HDB entry auth-data-reqd flag

nicowilliams:
One site needs to disable PACs altogether for service tickets, but might need to enabled them on a case by case basis.

With these patches we now default to using PACs, which can be turned off on a per-service principal basis with the `no-auth-data-reqd` flag, or one can disable PACs for all but root `krbtgt`s and then enable PACs on a case by case basis (including cross-realm `krbtgt`s) with the new `auth-data-reqd` HDB entry flag.

If both flags get set, then you end up with `no-auth-data-reqd` set and `auth-data-reqd` reset:
image

### DIFF
--- a/kdc/misc.c
+++ b/kdc/misc.c
@@ -348,6 +348,10 @@ _kdc_include_pac_p(astgs_request_t r)
     }
     if (r->server->flags.no_auth_data_reqd)
 	return FALSE;
+    if (r->server->flags.auth_data_reqd)
+	return TRUE;
+    if (r->config->disable_pac)
+        return FALSE;
 
     return !!(r->pac_attributes & (KRB5_PAC_WAS_REQUESTED | KRB5_PAC_WAS_GIVEN_IMPLICITLY));
 }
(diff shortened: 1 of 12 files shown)


kdc/misc.c
```diff
@@ -331,15 +331,27 @@ _kdc_verify_checksum(krb5_context context,
...
+        if (krb5_principal_is_root_krbtgt(r->context, r->server->principal) ||
+            !r->server->flags.no_auth_data_reqd)
+            return TRUE;
+    }
+    if (r->server->flags.no_auth_data_reqd)
 	return FALSE;
+    if (r->server->flags.auth_data_reqd)
+	return TRUE;
+    if (r->config->disable_pac)
+        return TRUE;
```
lhoward, line 354: Shouldn't `disable_pac` return FALSE?

lhoward (changes requested):

kdc/misc.c
nicowilliams: Oy, yes.

PR #1176, opened 2023-06-24
Use perl module JSON:PP, part of core, instead of JSON package

bkuhls:
This patch removes the need for an external package.

Ported from
https://github.com/openwrt/packages/blob/master/net/samba4/patches/105-perl-json-pp.patch https://github.com/openwrt/packages/commit/402f4ba4eff65b80a9deaa6085256112bec4d67b#diff-208d4e0345c9d29fbec23d6f655ba794afd3052f5cb8dd73944db72ce81b847b

### DIFF
--- a/cf/make-proto.pl
+++ b/cf/make-proto.pl
@@ -4,7 +4,7 @@
 use Getopt::Std;
 use File::Compare;
 
-use JSON;
+use JSON::PP
 
 my $comment = 0;
 my $doxygen = 0;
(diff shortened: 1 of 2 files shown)


nicowilliams:
So then we can remove `sudo cpanm install JSON` (and installation of `cpanm`) from GitHub Actions workflows?

bkuhls:
> So then we can remove `sudo cpanm install JSON` (and installation of `cpanm`) from GitHub Actions workflows?

I do not know, the topic came up with the buildroot while cross-compiling samba. We need a host version of compile_et for this and want to avoid adding the JSON package to the host version of perl.

nicowilliams:
> I do not know, the topic came up with the buildroot project while cross-compiling samba. We need a host version of compile_et for this and want to avoid adding the JSON package to the host version of perl.

So the answer is "yes, if you have a recent enough Perl5", I think.  I'll try it.  Thanks!

PR #1177, opened 2023-07-05
pkinit: Correctly pad Diffie-Hellman key

josephsutton1:
If the Diffie-Hellman key was ‘n’ bytes too short, we would shift it to the right ‘n’ places, padding it out to the correct length to compute the reply key.

Unfortunately, we forgot to increase the size of the key accordingly, so ‘n’ trailing key bytes would be discarded. This could mean failure to decrypt a reply when interoperating with a Kerberos implementation without this bug.

### DIFF
--- a/kdc/pkinit.c
+++ b/kdc/pkinit.c
@@ -230,6 +230,7 @@ generate_dh_keyblock(krb5_context context,
 	    size -= dh_gen_keylen;
 	    memmove(dh_gen_key + size, dh_gen_key, dh_gen_keylen);
 	    memset(dh_gen_key, 0, size);
+	    dh_gen_keylen += size;
 	}
     } else if (client_params->keyex == USE_ECDH) {
 	if (client_params->u.ecdh.public_key == NULL) {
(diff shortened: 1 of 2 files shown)


kdc/pkinit.c
```diff
@@ -230,6 +230,7 @@ generate_dh_keyblock(krb5_context context,
 	    size -= dh_gen_keylen;
 	    memmove(dh_gen_key + size, dh_gen_key, dh_gen_keylen);
 	    memset(dh_gen_key, 0, size);
+	    dh_gen_keylen += size;
```
abartlet, line 233: Does the buggy client mean we need to try both possibilities now?

josephsutton1: Yes, if we want to interoperate with buggy Heimdal clients without sporadic failures. For the other way round, interoperating with buggy Heimdal servers, the client probably would need to keep regenerating the key until it got a working one.

jaltman: Back in 2010 there were interop problems with MIT which resulted in the submission of both 7639f83561a5d8967094e7e76caebb6d1890791f and then 6e05462c1e83a7dc02995430dffb4213132fe373.  Does the existing code successfully interoperate with MIT?

josephsutton1: >  Does the existing code successfully interoperate with MIT?

I don’t know. I haven’t tested that.

mattwoodyard:
The client side of this patch seems to have fixed client side pkinit failures we were seeing against an AD KDC.

PR #1178, opened 2023-07-05
krb5: Clarify documentation for ‘pkinit_revoke’ parameter

josephsutton1:
If multiple valid CRLs are specified for a particular issuer, only the first will be checked. The documentation didn’t really hint at this.


abartlet (approved):
Thanks so much for improving the documentation.  @nicowilliams can we merge this please or are you OK if I do so?

PR #1184, opened 2023-08-14
Fix code spelling

josephsutton1:
A few spelling fixes that I’ve accumulated from working on the KDC and X.509 areas of the codebase.


abartlet (approved):
Thanks so much for doing this.

It would be great to merge this into upstream Heimdal not just as an improvement generally, but because for Samba, we need these correct as we have started to use `codespell` and it is nice to be able to include the our vendored heimdal in that eventually. 

Additionally, this would be an easy win to reduce the diff between master and our lorikeet-heimdal tree.

abartlet:
@jaltman @nicowilliams this seems safe to merge, can I or would you like to?

PR #1185, opened 2023-08-14
Unicode transcoding fixes

josephsutton1:
Reject invalid UTF-8 encodings, and add support for transcoding non-BMP codepoints (in the range U+10000 to U+0x10FFFF, inclusive).


lib/wind/utf8.c
```diff
@@ -87,9 +97,16 @@ utf8toutf32(const unsigned char **pp, uint32_t *out)
 	    return WIND_ERR_INVALID_UTF8;
 	}
     } else {
-	*out = c;
+	out_val = c;
     }
 
+    /* Allow unpaired surrogates (in the range 0xd800–0xdfff). */
```
nicowilliams, line 103: Do we have to?

lib/wind/utf8.c
```diff
@@ -220,8 +237,8 @@ wind_ucs4utf8(const uint32_t *in, size_t in_len, char *out, size_t *out_len)
             default:
                 break;
 	    }
+	    out += len;
```
nicowilliams, line 240: We already `out += len` above.  The one below that you moved here looks suspicious to me.

nicowilliams, line 240: What we could have done instead is `if (!out) continue;`, then deindent all the code in this `if` block, which would make it more readable, but this will do.

lib/wind/test-utf8.c
```diff
@@ -50,21 +50,32 @@ static const char *failing_testcases[] = {
     "\xF7",
     "\xC0\x01",
     "\xC0\x7F",
+    "\xC0\x80",
```
nicowilliams, line 53: Right, this is an over-long sequence for `U+0000`.

lib/wind/test-utf8.c
```diff
@@ -50,21 +50,32 @@ static const char *failing_testcases[] = {
...
+    "\xC0\x81",
```
nicowilliams, line 54: And this is an overlong sequence for `U+0001`.

lib/wind/test-utf8.c
```diff
@@ -50,21 +50,32 @@ static const char *failing_testcases[] = {
...
     "\xC0\xFF",
     "\xC0\x80\x80",
+    "\xC1\x80",
```
nicowilliams, line 57: This is an overlong sequence for `U+0040`.

lib/wind/test-utf8.c
```diff
@@ -50,21 +50,32 @@ static const char *failing_testcases[] = {
...
     "\xE0\x01",
     "\xE0\x7F",
     "\xE0\x80",
     "\xE0\xFF",
     "\xE0\x80\x20",
     "\xE0\x80\xFF",
+    "\xE0\x80\x80",
```
nicowilliams, line 64: And so on.

lib/wind/utf8.c
```diff
@@ -412,15 +429,30 @@ wind_utf8ucs2(const char *in, uint16_t *out, size_t *out_len)
 	if (ret)
 	    return ret;
 
-	if (u & 0xffff0000)
-	    return WIND_ERR_NOT_UTF16;
```
nicowilliams, line 416: We should have an error code `WIND_ERR_NOT_BMP` or something.

lib/wind/utf8.c
josephsutton1: Yeah, I thought we might want to continue to allow unpaired surrogates in case something depends on their use. I’ve known Windows to produce them, for example.

lib/wind/utf8.c
```diff
@@ -412,15 +429,30 @@ wind_utf8ucs2(const char *in, uint16_t *out, size_t *out_len)
...
+		low_ten_bits = u & 0x3ff;
+
+		out[o] = 0xd800 | high_ten_bits;
+		out[o+1] = 0xdc00 | low_ten_bits;
+	    }
+	    o += 2;
+	} else {
+	    if (out) {
+		if (o >= *out_len)
+		    return WIND_ERR_OVERRUN;
```
nicowilliams, line 451: `WIND_ERR_OVERRUN` seems wrong (but so was `WIND_ERR_NOT_UTF16`.

lib/wind/utf8.c
nicowilliams: This was the only place where we used `WIND_ERR_NOT_UTF16`, so maybe we should rename it to `WIND_ERR_NOT_BMP`.

nicowilliams: We already have `WIND_ERR_INVALID_UTF16` for use in decoding of UTF-16, so we really don't need `WIND_ERR_NOT_UTF16`, so let's rename this to `WIND_ERR_NOT_BMP`, and let's use it below.

nicowilliams:
Thanks!

josephsutton1:
Thanks for merging this. Is there anything more I have to address regarding error codes?

nicowilliams:
> Is there anything more I have to address regarding error codes?

No, there's nothing to be done.

josephsutton1:
> No, there's nothing to be done.

OK, thanks!

PR #1187, opened 2023-09-21
asn1: Do not emit extra typedefs nor extra ASN.1 C commentary

nicowilliams:


lib/asn1/gen.c
```diff
@@ -1694,13 +1711,15 @@ declare_type(const Symbol *s, Type *t, int typedefp)
 {
     char *newbasename = NULL;
 
-    if (typedefp)
+    if (typedefp && !s->emitted_declaration)
```
nicowilliams, line 1714: This change is not necessary.  I'll remove it.

lib/asn1/gen.c
```diff
@@ -1903,10 +1925,13 @@ generate_type_header (const Symbol *s)
      * member fields are not OPTIONAL/DEFAULTed.
      */
     generate_subtypes_header(s);
-    fprintf(headerfile, "/*\n");
-    fprintf(headerfile, "%s ::= ", s->name);
-    define_asn1 (0, s->type);
-    fprintf(headerfile, "\n*/\n\n");
+    if (!s->emitted_asn1) {
```
nicowilliams, line 1928: For every type we emit it's ASN.1 definition in a C comment above it, but we recurse just above (`generate_subtypes_header(s);`) and we were emitting the same commentary over and over.

lib/asn1/gen.c
```diff
@@ -1963,9 +1988,22 @@ generate_type_header (const Symbol *s)
...
-    fprintf(headerfile, "typedef ");
-    define_type(0, s->gen_name, s->gen_name, NULL, s->type, TRUE,
-                preserve_type(s->name) ? TRUE : FALSE);
+    if (!s->emitted_declaration) {
+        fprintf(headerfile, "typedef ");
+        define_type(0, s->gen_name, s->gen_name, NULL, s->type,
+                    DEF_TYPE_TYPEDEFP | DEF_TYPE_EMIT_NAME |
+                    (preserve_type(s->name) ? DEF_TYPE_PRESERVE : 0));
+    } else if (s->type->type == TType) {
+        /* This is a type alias and we've already declared it */
```
nicowilliams, line 1997: We've already emitted all the forward declarations, but since for type aliases the definition is the same as the declaration, we should just emit nothing here.

nicowilliams: This is the case of an alias of the form `Foo := Bar`.

lib/asn1/gen.c
```diff
@@ -1963,9 +1988,22 @@ generate_type_header (const Symbol *s)
...
+    } else if (s->type->type == TTag &&
+               s->type->subtype != NULL &&
+               s->type->subtype->symbol != NULL) {
+        /* This is a type alias and we've already declared it */
```
nicowilliams, line 2001: As above.  This is the case of `Foo ::= [APPLICATION 12345] Bar`.

lib/asn1/gen.c
```diff
@@ -1963,9 +1988,22 @@ generate_type_header (const Symbol *s)
...
+    } else {
+        define_type(0, s->gen_name, s->gen_name, NULL, s->type,
```
nicowilliams, line 2003: This is the case where we've not emitted a forward declaration and this should be a `struct`, `enum`, `union`, etc.  We don't emit `typedef`, and we don't want to emit the name of the type at the end.

lib/asn1/gen.c
```diff
@@ -937,7 +937,13 @@ getnewbasename(char **newbasename, int typedefp, const char *basename, const cha
 	err(1, "malloc");
 }
 
-static void define_type(int, const char *, const char *, Type *, Type *, int, int);
+typedef enum define_type_options {
```
nicowilliams, line 940: `define_type()` had two boolean arguments.  I couldn't stomach one more.

lib/asn1/gen.c
```diff
@@ -937,7 +937,13 @@ getnewbasename(char **newbasename, int typedefp, const char *basename, const cha
...
+    DEF_TYPE_NONE = 0,
+    DEF_TYPE_PRESERVE = 1,
+    DEF_TYPE_TYPEDEFP = 2,
```
nicowilliams, line 943: I don't know what to call this one :(

lib/asn1/symbol.c
```diff
@@ -171,3 +171,9 @@ emitted_tag_enums(const Symbol *s)
 {
     ((Symbol *)(uintptr_t)s)->emitted_tag_enums = 1;
 }
+
+void
+emitted_asn1(const Symbol *s)
+{
+    ((Symbol *)(uintptr_t)s)->emitted_asn1 = 1;
```
nicowilliams, line 178: This is just unconsting.

lib/asn1/gen.c
jaltman: bit flags are good

lib/asn1/gen.c
jaltman: I'm not sure

lib/asn1/gen.c
jaltman: I tested without this change and I concur.

lib/asn1/gen.c
jaltman: ack

jaltman (changes requested):
Please update with the indicated change reverted.

jaltman (approved):

PR #1190, opened 2023-11-01
Support later Visual Studio releases better

fanc999-1:
Hi,

This attempts to make the Heimdal builds against later Visual Studio better, by:

*  Updating `include\config.w32` to define that `stdint.h` and `inttypes.h` are available, when Visual Studio 2010 or later and Visual Studio 2013 or later are used, respectively, since these headers are shipped with these versions of Visual Studio.
*  Update `windows\NTMakefile.config` and `lib\roken\NTMakefile`  to reflect on the availability of `stdint.h` and `stdbool.h`, which are available since Visual Studio 2010 and Visual Studio 2013, repsectively.  As a consequence, the compatibility headers for these headers that comes in `lib\roken` is not copied in cases where Visual Studio ships with them).   This will fix builds that use Heimdal that is horribly broken when when the system `inttypes.h` is included, as things in the compatibility `stdint.h` will clash with the system's `inttypes.h` badly.
* Update the `makeinfo.exe` call to instead use whatever that that is passed in throught envvars or the NMake command line.

With blessings, thank you!

### DIFF
--- a/include/config.h.w32
+++ b/include/config.h.w32
@@ -577,19 +577,23 @@ static const char *const rcsid[] = { (const char *)rcsid, "@(#)" msg }
 /* #undef HAVE_INNETGR */
 
 /* Define to 1 if the system has the type `int16_t'. */
-/* #define HAVE_INT16_T 1 */
+#if (_MSC_VER >= 1600)
+#define HAVE_INT16_T 1
 
 /* Define to 1 if the system has the type `int32_t'. */
-/* #define HAVE_INT32_T 1 */
+#define HAVE_INT32_T 1
 
 /* Define to 1 if the system has the type `int64_t'. */
-/* #define HAVE_INT64_T 1 */
+#define HAVE_INT64_T 1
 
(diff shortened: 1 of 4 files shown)


abartlet:
It seems there are bigger issues with the windows build sadly.

sfhacker:
Hi.
Can you post either here or raise a new issue listing the error(s) you have encountered? I would like to reproduce them locally.
Thanks in advance.

abartlet:
[logs_3289.zip](https://github.com/heimdal/heimdal/files/13342777/logs_3289.zip)

The logs are available to me when logged in, but likely not guests.  So I attach the zip file.

nicowilliams:
image

nicowilliams:
The problem may be this in `windows/NTMakefile.w32`:

```
{}.hin{$(INCDIR)}.h:
        $(CP) $< $@
```

which I think builds `include/stdint.h` from `lib/roken/stdint.h`.  Where MSVC has its own `stdint.h` we must then not build our own `stdint.h`.

PR #1191, opened 2023-11-01
hx509: Add support for CSRs w/ BasicConstraints

nicowilliams:
bx509d / kx509 will reject CSRs that have an `extReq` w/ a `BasicConstraints`, but it should really either ignore it or reject `cA` requests.

As well it should ignore non-critical unknown `Extension`s in `extReq`.

This comes up with CSRs issued by Dell iDRACs, which include a non-critical `BasicConstraints` extension w/ `cA` set to false.  Nothing in RFCs 2985 (PKCS#9) or 2986 (PKCS#10) says anything about CSRs having to include `BasicConstraints` or not, but evidently some do so we need to support that.

 - [x] do not reject CSRs that have non-critical unknown or unsupported `extReq` extensions
 - [x] support `BasicConstraints` as an extension in CSRs
 - [x] default to not authorizing CSRs that request a CA certificate
 - [x] ~call `hx509_request_get_cA()` in kx509/bx509d and reject requests that have that set~
       (defaults to not issuing CA certs anyways)
 - [x] write a test of this for bx509d

`libhx509` changes:

 - when parsing a CSR check for `BasicConstraints` in the `extReq`
 - count critical and non-critical unsupported `extReq`s separately
 - add `hx509_request_set_cA()` to mark a CSR as wanting a CA cert (this has an argument for setting the optional `pathLenConstraint`)
 - add `hx509_request_set_eE()` to mark a CSR as wanting an EEcert (but this will make the CSR to be output include `BasicConstraints`, just with `cA` set to `FALSE`, and therefore absent)
 - add `hx509_request_get_cA()` and `hx509_request_get_cA_pathLenConstraint()` getters
 - add `hx509_request_authorize_cA()` (and nothing in-tree uses it)

`hxtool` changes:

 - `request-create` sub-command gets three new options:
   - `--ca`
   - `--ca-path-length`
   - `--ee`

### DIFF
--- a/lib/hx509/req.c
+++ b/lib/hx509/req.c
@@ -46,11 +46,15 @@ struct hx509_request_data {
     KeyUsage ku;
     ExtKeyUsage eku;
     GeneralNames san;
+    BasicConstraints bc;
     struct abitstring_s authorized_EKUs;
     struct abitstring_s authorized_SANs;
-    uint32_t nunsupported;  /* Count of unsupported features requested */
-    uint32_t nauthorized;   /* Count of supported   features authorized */
+    uint32_t nunsupported_crit; /* Count of unsupported critical features requested */
+    uint32_t nunsupported_opt;  /* Count of unsupported optional features requested */
+    uint32_t nauthorized;       /* Count of supported   features authorized */
+    uint32_t ca_is_authorized:1;
     uint32_t ku_are_authorized:1;
+    uint32_t include_BasicConstraints:1;
 };
 
 /**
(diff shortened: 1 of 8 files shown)


nicowilliams:
@jaltman reviewed.

lib/hx509/req.c
```diff
@@ -1515,6 +1666,14 @@ hx509_request_print(hx509_context context, hx509_request req, FILE *f)
             break;
         }
     }
+    if (req->nunsupported_crit) {
+        fprintf(f, "  unsupported_critical_extensions_count: %u\n",
+                (unsigned)req->nunsupported_crit);
+    }
+    if (req->nunsupported_crit) {
```
josephsutton1, line 1673: Coverity noticed that this looks like a copy‐paste error — `req->nunsupported_crit` should presumably read `req->nunsupported_opt`.

nicowilliams:
> Coverity noticed that this looks like a copy‐paste error — `req->nunsupported_crit` should presumably read `req->nunsupported_opt`.

Thanks!

PR #1193, opened 2023-11-14
kadm5: allow setting password_lifetime to 0 to clear

dariaphoebe:
when [kadmin] password_lifetime is set to 0, it should clear the time rather than making the freshly-changed password expire immediately


dariaphoebe:
I have made the requested change to update the documentation and updated the PR

nicowilliams:
LGTM.

PR #1200, opened 2023-12-15
fix build with llvm16+ where --no-undefined-version is the default

svmhdvn:
LLVM 16+ enables the `--no-undefined-version` flag in LLD by default in commit https://github.com/llvm/llvm-project/commit/241dbd310599e3c1a0f1b0c9ced14c8b8760539e. This causes many build errors of the form: `error: version script assignment of <VERSION> to symbol <SYMBOL> failed: symbol not defined` in heimdal.

This patch enables preprocessing of the `version-script.map` files so that we can conditionally export symbols in version scripts based on `#define` values in `config.h`.


nicowilliams:
Hey, this is very neat!  I need to spend some time reviewing this, but yeah, the idea is sound, and we're almost certainly going to need to be doing this in the longer term as the linker-editors become more strict.

svmhdvn:
Sounds good, I do have some bandwidth to push this PR to submission so let me know whenever you get the time.

lib/hx509/version-script.mapin
```diff
@@ -131,7 +129,6 @@ HEIMDAL_X509_1.2 {
 		hx509_certs_free;
 		hx509_certs_info;
 		hx509_certs_init;
-		hx509_certs_iter;
```
nicowilliams, line 134: I suppose this (and a few others) could be `#ifdef __BLOCKS__`...

lib/roken/version-script.mapin
```diff
@@ -1,23 +1,14 @@
...
-		cgetclose;
-		cgetmatch;
-		cgetnum;
-		cgetset;
-		cgetustr;
 		ct_memcmp;
-		err;
-		errx;
 		free_getarg_strings;
-		fseeko;
```
nicowilliams, line 14: Interesting.  `fseeko()`/`ftello()` are really only here for compatibility with Windows, but we build these on Unix too as empty translation units (because `HAVE_FTELLO` is defined).  I'm inclined to remove them from this mapfile as you're doing, but also from `lib/roken/Makefile.am`; alternatively `#ifdef` them here.

lib/roken/version-script.mapin
```diff
@@ -1,23 +1,14 @@
...
-		ftello;
 		get_window_size;
 		getarg;
 		getnameinfo_verified;
-		hex_decode;
```
nicowilliams, line 19: These need to be left in.

lib/roken/version-script.mapin
```diff
@@ -61,38 +56,74 @@ HEIMDAL_ROKEN_2.0 {
...
+#ifndef HAVE_FNMATCH
 		rk_fnmatch;
+#endif
 		rk_free_environment;
+#ifndef HAVE_FREEADDRINFO
 		rk_freeaddrinfo;
+#endif
+#ifndef HAVE_FREEHOSTENT
 		rk_freehostent;
-		rk_freeifaddrs;
```
nicowilliams, line 69: This needs to be `#ifdef``'ed somehow.

lib/roken/version-script.mapin
```diff
@@ -61,38 +56,74 @@ HEIMDAL_ROKEN_2.0 {
...
-		rk_gai_strerror;
```
nicowilliams, line 70: Ditto.

lib/roken/version-script.mapin
```diff
@@ -61,38 +56,74 @@ HEIMDAL_ROKEN_2.0 {
...
+#ifndef HAVE_INET_NTOP
 		rk_inet_ntop;
+#endif
+#ifndef HAVE_INET_PTON
 		rk_inet_pton;
+#endif
 		rk_injectauxv;
+#ifndef HAVE_LOCALTIME_R
 		rk_localtime_r;
-		rk_memmem;
```
nicowilliams, line 88: We do use `memmem()`/`rk_memmem()`.

lib/roken/version-script.mapin
```diff
@@ -219,8 +301,6 @@ HEIMDAL_ROKEN_2.0 {
 		rtbl_set_flags;
 		rtbl_set_prefix;
 		rtbl_set_separator;
-		signal;
```
nicowilliams, line 222: Do we even need the `signal()` wrapper?  And why isn't it called `rk_signal()`?  @jaltman.  Maybe we should just remove it.

lib/roken/version-script.mapin
```diff
@@ -229,13 +309,9 @@ HEIMDAL_ROKEN_2.0 {
 		unparse_time_approx;
 		unparse_units;
 		unparse_units_approx;
-		verr;
```
nicowilliams, line 232: We don't even build `lib/roken/err.c` on non-Windows, so maybe we should also not refer to `err.h` in `lib/roken/Makefile.am`.  (NOTE: I'm not asking OP to make any change in this respect.)

lib/roken/version-script.mapin
```diff
@@ -229,13 +309,9 @@ HEIMDAL_ROKEN_2.0 {
...
-		verrx;
-		vwarn;
-		vwarnx;
-		warn;
-		warnx;
-		writev;
+#ifndef HAVE_WRITEV
```
nicowilliams, line 312: We don't even build `lib/roken/writev.c` on non-Windows.

nicowilliams:
I think this is pretty close.  Thanks for doing this work!

lib/hx509/version-script.mapin
svmhdvn: Thanks, fixed.

lib/roken/version-script.mapin
svmhdvn: Fixed, I just wrapped them in ifdefs

lib/roken/version-script.mapin
svmhdvn: Thanks, fixed.

lib/roken/version-script.mapin
svmhdvn: I deleted this because `memmem` is redirected to `rk_smemmem`. Take a look at https://github.com/heimdal/heimdal/blob/cb9a13032285ee9a20507f21c44b67a547a9996c/lib/roken/roken.h.in#L505.

Is this intentional? I don't find any other references to `rk_memmem` directly.

lib/roken/version-script.mapin
svmhdvn: For the sake of keeping this PR focused, this should be fine to keep in here right? We can remove `writev.c` in a separate commit?

lib/roken/version-script.mapin
nicowilliams: Yes, it gets renamed to `rk_memmem()`, so the three uses in `kdc/connect.c` are actually uses of `rk_memmem()` (but only when the OS/C library doesn't have a `memmem()`).  That's intentional.

lib/roken/version-script.mapin
nicowilliams: Great question.  I'm not sure, but I think you're right.  @jaltman may know.

lib/roken/version-script.mapin
nicowilliams: Just wrapping the `writev` symbol with `#ifndef HAVE_WRITEV` should suffice, though on Unix systems it will always be that we `HAVE_WRITEV`, so you can also just remove it from `lib/roken/version-script.mapin`.

lib/roken/version-script.mapin
svmhdvn: Sorry for the annoyance and my possible misunderstanding of your comment, but could you point me to the definition of `rk_memmem`? Here's my understanding:

* Renaming of `memmmem` to `rk_smemmem` happens here: https://github.com/heimdal/heimdal/blob/cb9a13032285ee9a20507f21c44b67a547a9996c/lib/roken/roken.h.in#L506-L507

* Definition of `memmem` (which gets renamed again to `rk_smemmem` by the previous macro) is here: https://github.com/heimdal/heimdal/blob/cb9a13032285ee9a20507f21c44b67a547a9996c/lib/roken/memmem.c#L38-L41

* All calls of `memmem` happen in `kdc/connect.c`

Where does the renaming happen for `memmem` to `rk_memmem`? Note the difference of the letter `s` between `rk_memmem` and `rk_smemmem`.

lib/roken/version-script.mapin
svmhdvn: I'll keep it removed for now then

lib/roken/version-script.mapin
svmhdvn: Thanks, fixed.

lib/roken/version-script.mapin
svmhdvn: Thanks, removing it then.

lib/roken/version-script.mapin
nicowilliams: Argh, where did that `s` come from?  Well, that's clearly a typo.  Now I'm curious how this works on Windows.

nicowilliams:
image

image

svmhdvn:
Can you give repro instructions for those errors? They look unrelated to my changes. Things are building successfully on my local system with LLVM 17.

Quick thing I noticed, I've accidentally added libedit's `chartype.h` local changes to this PR, deleting this now and repushing.

nicowilliams:
@svmhdvn you can see those in the Windows build errors in the checks that are being run on this PR.  Just click on `Details` for each failed build, click on the failed action, click the search field, type `error`, then search backwards through the logs.  There's also a log download option.  Note that the full logs are not loaded into the page, so you can't reliably use the browser's search feature.

svmhdvn:
How can I trigger the Windows build again? My PR doesn't seem to trigger it anymore.

nicowilliams:
> How can I trigger the Windows build again? My PR doesn't seem to trigger it anymore.

See `.github/workflows/windows.yml`:

```
    pull_request:
      paths:
         - '!docs/**'
         - '!**.md'
         - '!**.[1-9]'
         - '**.[chly]'
         - '**.hin'
         - '**.in'
         - '**.pl'
         - '**.py'
         - '**.asn1'
         - '**.opt'
         - '**.w32'
         - '**/NTMakefile'
         - '**/COPYING'
         - '**/INSTALL'
         - '**/README*'
         - '.github/workflows/windows.yml'
         - '!appveyor.yml'
         - '!.travis.yml'
```

svmhdvn:
If the latest version of my PR doesn't trigger any windows builds due to the windows filter, then it shouldn't affect the windows build in any way right? Is this review otherwise complete?

nicowilliams:
I'll do a final review and merge.  Thanks!

nicowilliams:
Can you rebase this?

Issue #1205, opened 2024-01-02
Use-after-free in krb5_sendto with debug logs, HTTP transport to KDC, and http_proxy


riastradh:
**Describe the bug**
When

1. debug-level logging is enabled,
2. the KDC transport is HTTP, and
3. http_proxy is configured,

submit_request will allocate an addrinfo with getaddrinfo and store it at host->ai, and then free it with freeaddrinfo before returning, but later on in advancing the krb5_sendto state machine, subsequent debug_host calls will continue to use host->ai to format debug messages.

**To Reproduce**
1. set `KRB5_TRACE=0-100/STDERR`
2. configure an HTTP-based KDC
3. configure http_proxy
5. run kinit

(Alternatively: code inspection.  Alternatively: observe the same bug while drafting SOCKS4a proxy support.)

**Expected behavior**
no use after free

PR #1206, opened 2024-01-02
krb5: Fix use-after-free when debug is enabled with http_proxy.

riastradh:
Instead of freeing host->ai on return from submit_request in the http_proxy path, stash the proxy's ai in another member host->freeai which will be freed when host is freed.

Assumption: All hosts created in submit_request cease to be used before any of them are freed, so it's safe to pick one host to hang the proxy's ai on and free the ai when that host is freed.

fix https://github.com/heimdal/heimdal/issues/1205

### DIFF
--- a/lib/krb5/send_to_kdc.c
+++ b/lib/krb5/send_to_kdc.c
@@ -328,6 +328,7 @@ struct host_fun {
 struct host {
     enum host_state { CONNECT, CONNECTING, CONNECTED, WAITING_REPLY, DEAD } state;
     krb5_krbhst_info *hi;
+    struct addrinfo *freeai;
     struct addrinfo *ai;
     rk_socket_t fd;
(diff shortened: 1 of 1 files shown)


lib/krb5/send_to_kdc.c
```diff
@@ -393,6 +394,8 @@ deallocate_host(void *ptr)
     if (!rk_IS_BAD_SOCKET(host->fd))
 	rk_closesocket(host->fd);
     krb5_data_free(&host->data);
+    free(host->freeai);
```
nicowilliams, line 397: Shouldn't this be `freeaddrinfo()` instead of `free()`?  Should this be contingent on `host->freeai` not being `NULL`?

nicowilliams:
LGTM.  Thanks!

Issue #1186, opened 2023-09-05
`libatomic` not linked for 64-bit atomics


jpalus:
**Describe the bug**
`heimdal` started to use 64-bit atomics which are not supported natively on all targets. I see there is a toggle `HEIM_BASE_ATOMICS_FALLBACK` which set (manually?) provides some fallbacks for atomics. However on most platforms `stdatomic.h` + linking against `libatomic` is enough and I think that would be preferred. The latter part is missing though -- currently `heimdal` does not check if linking `libatomic` is necessary

**To Reproduce**
1. Try to build `heimdal` for `i386`, `i486`, `armv6` or `mips`

**Expected behavior**
Build works fine.

**Actual behavior**
Build fails on missing symbols due to lack of `-latomic`.

**Desktop (please complete the following information):**
 - OS: Linux

nicowilliams:
What targets don't provide atomics?  Also, yes, you can `#define` `HEIM_BASE_ATOMICS_FALLBACK` to get around this.  Ideally we should have autoconf checks for these.

jpalus:
Don't have any comprehensive list but some notable examples of platforms lacking support for 8-byte (64-bit) atomics:
* i386/i486 used as a target in some OSes for 32-bit x86
* ARMv6 (or lower) which I think is still the primary target in main Linux distribution for Raspberry Pis
* 32-bit MIPS used on many routers

More exotic examples include 32-bit PPC and SPARC.

Created #1208 with an attempt to fix this issue.

PR #1208, opened 2024-01-04
cf: link libatomic if needed

jpalus:
heimdal uses 8-byte (64-bits) atomic operations for a while now (ie in lib/krb5/krcache.c). however some platforms (ie i386, i486, ARM < 7) don't support them natively and need to be linked against libatomic in order to implement functionallity from <stdatomic.h>.

if the header was found check 64-bit atomics work without any additional libraries. if not try to add linking with libatomic. if that fails as well fallback to -DHEIM_BASE_ATOMICS_FALLBACK

no change for platfroms without <stdatomic.h>

Fixes: #1186

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -433,7 +433,6 @@ AC_CHECK_HEADERS([\
 	signal.h				\
 	strings.h				\
 	stropts.h				\
-	stdatomic.h				\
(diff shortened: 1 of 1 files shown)


nicowilliams:
Thanks!

PR #1209, opened 2024-01-05
Source code documentation `/**` unification.

kai-burghardt:
- This commit focuses on structure. It does not change the numerous typos or spelling mistakes.
- Consistently rely on "autobrief" feature: Remove `@brief`.
- Replace `@value` with `@param`.
- Replace `XXX` with `@todo`.
- Replace unstructured `Inputs:` text with `@param[in]` and `Outputs:` with `@param[out]`.

### DIFF
--- a/lib/base/array.c
+++ b/lib/base/array.c
@@ -68,9 +68,9 @@ struct heim_type_data array_object = {
 };
 
 /**
- * Allocate an array
+ * Allocate an array.
  *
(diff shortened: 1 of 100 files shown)


nicowilliams:
Thank you!  This looks very nice, though it's also very big and will take some time to review, but as long as we can see it's only commentary I think the review should be fast enough anyways.  My only concern is conflicts with other PRs, but I can always hand-fix the conflicts.

PR #1211, opened 2024-01-08
cf/make-proto.pl: make JSON:PP optional

metze-samba:


nicowilliams:
Oh, and we never use the `-x` option...

nicowilliams:
Thanks!

Issue #1212, opened 2024-01-09
DNS audit test harness might not work with static linking or symbol renaming


riastradh:
**Describe the bug**
The new test harness for auditing efficacy of the new block_dns option relies on dlsym(RTLD_NEXT, "getaddrinfo") to return the real getaddrinfo function pointer for resolving numeric hosts.

This could cause trouble in two scenarios:

1. if Heimdal is built purely statically linked, or on a system where ELF-style symbol interposition doesn't work; and/or
2. if the system's getaddrinfo function has a different symbol, e.g. declared in netdb.h with `int getaddrinfo(...) asm("__getaddrinfo50")` (like [NetBSD uses for some things](https://wiki.NetBSD.org/symbol_versions/), though not at the moment for getaddrinfo).

**To Reproduce**
critical thinking

**Expected behavior**
maybe slightly more robust to weird edge cases like this

PR #1213, opened 2024-01-09
Use AI_NUMERICSERV if block_dns, and use local getaddrinfo to audit.

riastradh:
This change has two parts:

1. Provide our own local implementation of numeric-only getaddrinfo in auditdns.c used to audit for DNS leaks, rather than deferring to dlsym(RTLD_NEXT, "getaddrinfo"), in terms of inet_pton.

   To keep review and implementation simple, this is limited to AI_NUMERICHOST _and_ AI_NUMERICSERV -- this requires that we arrange to pass AI_NUMERICSERV in callers too.

2. Wherever we implement block_dns, set AI_NUMERICSERV in addition to AI_NUMERICHOST as needed by the new auditdns.c getaddrinfo.

   (In principle this might also avoid other network leaks -- POSIX guarantees no name resolution service will be invoked, and gives NIS+ as an example.)

   One tiny semantic change to avoid tripping over the auditor: kadmin(8) now uses the string "749" rather than the string "kerberos-adm".  (Currently we don't audit kadmin(8) for DNS leaks but let's avoid leaving a rake to step on.)  Every other caller I found is already guaranteed to pass a numeric service rather than named service to getaddrinfo.

fix https://github.com/heimdal/heimdal/issues/1212

### DIFF
--- a/appl/test/auditdns.c
+++ b/appl/test/auditdns.c
@@ -28,8 +28,11 @@
 #include <config.h>
 #endif
 
-#include <dlfcn.h>
+#include <arpa/inet.h>
+#include <assert.h>
+#include <errno.h>
 #include <netdb.h>
+#include <netinet/in.h>
 #include <stdio.h>
 #include <stdlib.h>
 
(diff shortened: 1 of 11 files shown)


appl/test/auditdns.c
```diff
@@ -65,32 +68,308 @@ gethostbyname2(const char *name, int af)
...
+     * Check whether a service is specified at all.
+     */
+    if (servname == NULL) {
+	/*
+	 * No service specified.  Use the wildcard port 0.
+	 */
+	port = 0;
+    } else {
+	/*
+	 * Service specified.  Parse it as a nonnegative integer, at
```
nicowilliams, line 181: One might think that `strtoul()` would refuse to parse negative numbers, but, no, it doesn't.  What I do for this sort of thing is use `strtol()`/`strtoll()`/`strtoimax()` as appropriate, check that the the result is valid, not negative, then use the result, and if I have to support numeric values larger than the signed result type's largest integer then I explicitly look for a `-` sign.  What a pain.  Not saying you have to here -- this is a test program and there will not be negative numbers here.  I'm just... ranting.

appl/test/auditdns.c
```diff
@@ -65,32 +68,308 @@ gethostbyname2(const char *name, int af)
...
+		addrlen[0] = sizeof(addr->sin6);
+		goto have_addr;
+	    }
+	}
+    }
+
+    /*
+     * No hostname, or hostname can't be parsed.
+     */
+    error = EAI_NONAME;
```
nicowilliams, line 308: This isn't reachable, right?

nicowilliams:
Thanks!

appl/test/auditdns.c
nicowilliams: You don't have to unless you want to.

appl/test/auditdns.c
nicowilliams: No need.  Negative numbers will be rejected anyways just because they would be larger than 65535 anyways.  I merely felt the need to whine about `strtoul()` :)

Issue #1214, opened 2024-01-09
block_dns as implemented fails to block reverse DNS (PTR) lookups


riastradh:
**Describe the bug**
Heimdal may trigger DNS queries even though block_dns is enabled, via getnameinfo.

This is so because when drafting https://github.com/heimdal/heimdal/pull/1155 to block all paths to getaddrinfo without AI_NUMERICHOST (and now, since https://github.com/heimdal/heimdal/pull/1213, AI_NUMERICSERV), I neglected to consider reverse lookups via getnameinfo.

**To Reproduce**
- code inspection
- call krb5_sock_to_principal

**Expected behavior**
Setting `[libdefaults] block_dns = true` actually blocks all DNS queries.

Need to:
- [ ] Go through all the calls to getnameinfo and ensure they pass NI_NUMERICHOST|NI_NUMERICSERV (and maybe also NI_NUMERICSCOPE, out of paranoia).
- [ ] Update auditdns.c to provide stubs for getnameinfo and gethostbyaddr that abort except in the numeric-only cases.

nicowilliams:
Can we not just mostly get rid of `getnameinfo()`?

 - `krb5_sock_to_principal()` is not used and should be removed (I doubt we'd be breaking any apps, but we could just leave a stub that stringifies the address and be done)
 - `ipropd-slave` uses `getnameinfo()` for debug messages, but it's pointless because we should know the upstream's name anyways (it's given on the command line and it has to be a hostname because we do call `krb5_sname_to_principal()` on it), so we don't need to look up the IP's name here ever
 - `add_locate()`'s use of `getnameinfo()` is suspect and I haven't grokked it yet, but there's no attack here
 - `debug_host()` uses `getnameinfo()`, but always with debug level 5, so we could just leave that as-is

I think that's all uses of `getnameinfo()`.  I believe we do not otherwise use `gethostbyaddr()` or `getipnodebyaddr()` (except via roken's `rk_getnameinfo()` wrapper.

nicowilliams:
Maybe we should just have roken wrappers for reverse lookups that simply fail...

nicowilliams:
> I have no attachment to the existing uses, except that if we keep getnameinfo in debug_host, I'd like to make it respect block_dns so you don't have to worry about cranking up debug logs causing network privacy leaks.

Agreed.

> I started to draft a change along the lines I sketched, but I can also just delete these uses instead.

[...]

PR #1215, opened 2024-01-10
Prevent network leaks of reverse DNS lookups with [libdefaults] block_dns = yes.

riastradh:
1. Unconditionally pass NI_NUMERICHOST|NI_NUMERICSERV|NI_NUMERICSCOPE to most getnameinfo calls.
2. Predicate the remaining call, in krb5_sock_to_principal, on block_dns = no.
3. Add getnameinfo and gethostbyaddr stubs to auditdns.c to help check for leaks.

While here: tidy up the minor issues left over from https://github.com/heimdal/heimdal/pull/1213 (https://github.com/heimdal/heimdal/pull/1213#pullrequestreview-1811770450, https://github.com/heimdal/heimdal/pull/1213#pullrequestreview-1811778956, and a harmless stray semicolon).

fix https://github.com/heimdal/heimdal/issues/1214

### DIFF
--- a/appl/test/auditdns.c
+++ b/appl/test/auditdns.c
@@ -35,8 +35,10 @@
 #include <netinet/in.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <string.h>
 
 #include "resolve.h"
+#include "roken.h"
 
 struct rk_dns_reply *
 rk_dns_lookup(const char *domain, const char *type_name)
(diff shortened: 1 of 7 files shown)


appl/test/auditdns.c
```diff
@@ -66,6 +68,24 @@ gethostbyname2(const char *name, int af)
 
 #endif	/* HAVE_GETHOSTBYNAME2 */
 
+struct hostent *
+gethostbyaddr(const void *addr, socklen_t len, int af)
+{
+    const socklen_t maxlen[] = {
+	[AF_INET] = sizeof(struct in_addr),
```
nicowilliams, line 75: This won't work for Windows.  Did I miss anything similar earlier?  Anyways, the reason is that we still support older MSFT tools that don't fully support C99.

nicowilliams:
Thanks!

PR #1217, opened 2024-01-10
krb5: Fix DNS SRV name collision handling

nicowilliams:
I believe the code we have in Heimdal to deal with TLD name collisions is not quite right because the `target` of a DNS `SRV` RR has to have a trailing `'.'`.  This should fix it, though... we have no test.


jaltman:
@nicowilliams I do not believe the checks for malformed responses is required but you are correct that the current code is broken.  the strchr() check should be != NULL instead of == NULL.

PR #1228, opened 2024-01-27
Fix spelling

fel1x-developer:
Translated files in `/po` folder are not modified.


lhoward:
Best not to change the IETF drafts, as much as I like correct spelling :)

lhoward:
Ditto any RFCs and SQLite or any other imported code.

fel1x-developer:
Restored changes for drafts, RFCs, and SQlite.

Issue #1201, opened 2023-12-23
Compile with autoconf 2.72 fails


heitbaum:
**Describe the bug**

build error with Heimdal master

**To Reproduce**
- Run configure using autoconf 2.72

```
checking for /var/media/DATA/home-rudi/LibreELEC.tv/build.LibreELEC-Generic.x86_64-12.0-devel/toolchain/bin/host-gcc  options needed to detect all undeclared functions... cannot detect
configure: error: in '/var/media/DATA/home-rudi/LibreELEC.tv/build.LibreELEC-Generic.x86_64-12.0-devel/build/heimdal-3253c49544eacb33d5ad2f6f919b0696e5aab794/.x86_64-linux-gnu':
configure: error: cannot make /var/media/DATA/home-rudi/LibreELEC.tv/build.LibreELEC-Generic.x86_64-12.0-devel/toolchain/bin/host-gcc  report undeclared builtins
See 'config.log' for more details
```

**Expected behavior**

successful configuration. Configuration is successful with 2.71

nicowilliams:
Are you trying to cross-compile?  Cross-compilation of Heimdal is broken and incomplete.

PR #1229, opened 2024-02-10
cf/largefile.m4: Fix build with autoconf-2.72

bkuhls:
Fixes https://github.com/heimdal/heimdal/issues/1201


bkuhls:
Patch was inspired by https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=53ad9b81118ada34df80b4a85208e4ec2d04a0bb

PR #1234, opened 2024-03-01
Allow GSS_C_CHANNEL_BOUND_FLAG on the client

metze-samba:
Allow an application to provide GSS_C_CHANNEL_BOUND_FLAG on the client side and
force KERB_AP_OPTIONS_CBT.

See https://github.com/krb5/krb5/pull/1329


metze-samba:
See also https://github.com/krb5/krb5/pull/1329

PR #1241, opened 2024-05-09
kdc: Enforce hardware authentication for accounts requiring it

jsutton24:


lhoward:
LGTM, of course you can do PKINIT without hardware, but I too treated them as identical in XAD.

PR #1243, opened 2024-06-03
Fix gss_display_name output type for krb5 WELLKNOWN/ANONYMOUS@WELLKNOWN:ANONYMOUS plus warnings

jaltman:


lhoward:
LGTM (checked `krb5_principal_is_anonymous()` usage).

PR #1246, opened 2024-06-17
Fix handling of FILE ccache removed creds

jaltman:
For compatibility with MIT both endtime and authtime must be modified in order for the cred to be ignored when iterating the ccache contents.

Heimdal fcc_next_cred() must ignore removed credentials when iterating and not pass them back to the application.    Returning a cred with endtime < authtime violates the spirit of RFC4120 which requires that a KDC not return a cred with an endtime < the starttime or authtime.

### DIFF
--- a/lib/krb5/fcache.c
+++ b/lib/krb5/fcache.c
@@ -1088,6 +1088,9 @@ cred_delete(krb5_context context,
      */
     cred->times.endtime = 0;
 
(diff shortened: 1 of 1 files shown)


jaltman:
MIT updated their file ccache cred_removed() test in https://github.com/krb5/krb5/commit/4c0838bb4c232866b95c9f2f72a55bf77cfc1308

PR #1248, opened 2024-06-20
Gssapi sequence fixes

SimonWilkinson:
A couple of fixes for issues with checking sequence ordering in GSSAPI

### DIFF
--- a/lib/gssapi/krb5/arcfour.c
+++ b/lib/gssapi/krb5/arcfour.c
@@ -740,14 +740,14 @@ OM_uint32 _gssapi_unwrap_arcfour(OM_uint32 *minor_status,
 	return GSS_S_BAD_MIC;
     }
 
+    if (conf_state)
+	*conf_state = conf_flag;
+
     HEIMDAL_MUTEX_lock(&context_handle->ctx_id_mutex);
     omret = _gssapi_msg_order_check(context_handle->order, seq_number);
     HEIMDAL_MUTEX_unlock(&context_handle->ctx_id_mutex);
     if (omret)
 	return omret;
-
-    if (conf_state)
-	*conf_state = conf_flag;
 
     *minor_status = 0;
     return GSS_S_COMPLETE;
@@ -1375,15 +1375,15 @@ _gssapi_unwrap_iov_arcfour(OM_uint32 *minor_status,
 	}
     }
 
+    if (pconf_state) {
+	*pconf_state = conf_state;
+    }
+
     HEIMDAL_MUTEX_lock(&ctx->ctx_id_mutex);
     ret = _gssapi_msg_order_check(ctx->order, seq_number);
     HEIMDAL_MUTEX_unlock(&ctx->ctx_id_mutex);
     if (ret != 0) {
 	return ret;
-    }
-
-    if (pconf_state) {
-	*pconf_state = conf_state;
     }
 
     *minor_status = 0;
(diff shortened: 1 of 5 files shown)


lib/gssapi/krb5/sequence.c
```diff
@@ -188,7 +188,7 @@ _gssapi_msg_order_check(struct gss_msg_order *o, OM_uint32 seq_num)
     for (i = 0; i < o->length - 1; i++) {
 	if (o->elem[i] == seq_num)
 	    return GSS_S_DUPLICATE_TOKEN;
-	if (o->elem[i + 1] < seq_num && o->elem[i] < seq_num) {
```
jaltman, line 191: This has been broken since the original commit f24cb76b05 ("sequence number checks, order and replay")

jaltman:
[RFC2743 1.2.3: Per-Message Replay Detection and Sequencing](https://www.rfc-editor.org/rfc/rfc2743.html#page-21) describes the compliant behavior implemented by this pull request.

lib/gssapi/krb5/arcfour.c
```diff
@@ -740,15 +740,15 @@ OM_uint32 _gssapi_unwrap_arcfour(OM_uint32 *minor_status,
 	return GSS_S_BAD_MIC;
     }
 
+    if (conf_state)
+	*conf_state = conf_flag;
+
     HEIMDAL_MUTEX_lock(&context_handle->ctx_id_mutex);
     omret = _gssapi_msg_order_check(context_handle->order, seq_number);
     HEIMDAL_MUTEX_unlock(&context_handle->ctx_id_mutex);
     if (omret)
```
jaltman, line 749: I think minor_status needs to be set to 0 before this return.  Otherwise there might be random garbage used by the caller.   We might consider setting *minor_status = 0 upon entry to this function instead of setting it at each error return site.

lib/gssapi/krb5/arcfour.c
```diff
@@ -1375,17 +1375,17 @@ _gssapi_unwrap_iov_arcfour(OM_uint32 *minor_status,
...
     }
 
+    if (pconf_state) {
+	*pconf_state = conf_state;
+    }
+
     HEIMDAL_MUTEX_lock(&ctx->ctx_id_mutex);
     ret = _gssapi_msg_order_check(ctx->order, seq_number);
     HEIMDAL_MUTEX_unlock(&ctx->ctx_id_mutex);
     if (ret != 0) {
```
jaltman, line 1385: I think minor_status needs to be set to 0 before this return.  Otherwise there might be random garbage used by the caller.   We might consider setting *minor_status = 0 upon entry to this function instead of setting it at each error return site.

lib/gssapi/krb5/unwrap.c
```diff
@@ -175,10 +175,10 @@ unwrap_des
     return GSS_S_BAD_MIC;
   }
 
-  ret = _gssapi_msg_order_check(context_handle->order, seq_number);
-  if (ret) {
+  seq_err = _gssapi_msg_order_check(context_handle->order, seq_number);
+  if (seq_err == GSS_S_FAILURE) {
```
jaltman, line 179: I think minor_status needs to be set to 0 before this return.  Otherwise there might be random garbage used by the caller.   We might consider setting *minor_status = 0 upon entry to this function instead of setting it at each error return site.

jaltman (changes requested):
This change looks good to me with the exception that there are several functions where *minor_status is left unset.  As a result the caller will receive a non-GSS_S_COMPLETE with potentially random garbage for minor_status.

lib/gssapi/krb5/arcfour.c
SimonWilkinson: This is a pre-existing issue, that isn't affected by this change

lib/gssapi/krb5/unwrap.c
SimonWilkinson: Again, not setting minor_status is a pre-existing issue that isn't affected by using 'seq_err' rather than 'ret' here.

jaltman (approved):

jaltman (approved):

PR #1250, opened 2024-07-02
kdc: Check return value of pa_enc_ts_decrypt_kvno()

jsutton24:


kdc/kerberos5.c
```diff
@@ -919,6 +919,9 @@ pa_enc_ts_validate(astgs_request_t r, const PA_DATA *pa)
 	goto out;
     }
     free_EncryptedData(&enc_data);
+    if (ret) {
```
jaltman, line 922: @jsutton24 I am confused.  How could the flow of control arrive at this line with 'ret' non-zero?  If ret is non-zero at line 895 it will enter the block above that either executes a goto try_next_key or goto out at line 919.

jsutton24: Sorry, this was a Samba‐only issue, introduced by https://git.samba.org/?p=lorikeet-heimdal.git;a=patch;h=05819c3ae5c6395135dd6611021ac9fe70f3cf53.

PR #1255, opened 2024-08-29
krb5/fast.c: fix copying source/dest for copy_KrbFastArmor

manner82:
if using armoring service.

(The source and destination is exchanged in the call, which makes the "value" part (fxarmor.armor_value) not filled up, so the armored ticket request will not be encrypted correctly and will fail.)

### DIFF
--- a/lib/krb5/fast.c
+++ b/lib/krb5/fast.c
@@ -271,7 +271,7 @@ make_fast_ap_fxarmor(krb5_context context,
 	if (ret)
 	    goto out;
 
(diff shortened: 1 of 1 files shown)


jsutton24:
Nice catch!

jaltman (approved):

jaltman:
This reversal of parameters has been present since the original import from Apple.  47282cae349af91704217b8d3b40c6a4beb624a0.   As far as I can see the bug is still present as of https://github.com/apple-oss-distributions/Heimdal/releases/tag/Heimdal-689.140.1

PR #1262, opened 2024-10-13
Include db.h for nbdb compat mode

booxter:
Otherwise, on nixpkgs@darwin, the build fails as follows:

```
In file included from otp_db.c:42:
/nix/store/...-libSystem-11.0.0/include/ndbm.h:89:3: error: typedef redefinition with different types ('struct datum' vs 'struct datum') } datum;
  ^
/nix/store/...-db-5.3.28-dev/include/db.h:2749:3: note: previous definition is here
} datum;
  ^
```

Spotted in https://github.com/NixOS/nixpkgs/issues/347616

### DIFF
--- a/lib/otp/otp_db.c
+++ b/lib/otp/otp_db.c
@@ -39,7 +39,7 @@ RCSID("$Id$");
 #include "otp_locl.h"
 
 #if defined(HAVE_DB_NDBM)
-# include <ndbm.h>
+# include <db.h>
 #elif !defined(HAVE_NDBM)
 # include "ndbm_wrap.h"
 #endif


lib/otp/otp_db.c
```diff
@@ -39,7 +39,7 @@ RCSID("$Id$");
 #include "otp_locl.h"
 
 #if defined(HAVE_DB_NDBM)
-# include <ndbm.h>
+# include <db.h>
```
jaltman, line 42: Please test the following proposal:

Instead of this change, remove the entire block from line 41 to 45.  Then modify lib/roken/xdbm.h to add 
```

diff --git a/lib/roken/xdbm.h b/lib/roken/xdbm.h
index c2c6c28b4..1648be7da 100644
--- a/lib/roken/xdbm.h
+++ b/lib/roken/xdbm.h
@@ -47,6 +47,8 @@
 #elif defined(HAVE_NDBM_H)
 #include <ndbm.h>
 #endif
+#else /* !HAVE_NDBM */
+#include "ndbm_wrap.h"
 #endif /* HAVE_NDBM */
 
 #endif /* __XDBM_H__ */
```

otp_locl.h includes xdbm.h.

booxter: This worked. Thanks.

PR #1268, opened 2024-11-20
Small improvements in GSS mechglue logs

elric1:


lib/gssapi/mech/gss_accept_sec_context.c
```diff
@@ -264,7 +264,9 @@ gss_accept_sec_context(OM_uint32 *minor_status,
 	    *delegated_cred_handle = GSS_C_NO_CREDENTIAL;
 	_mg_buffer_zero(output_token);
 
+        _gss_mg_log(10, "gss-asc: enter");
```
nicowilliams, line 267: Do we need a log level to avoid logs getting too noisy?

lib/gssapi/mech/gss_accept_sec_context.c
```diff
@@ -264,7 +264,9 @@ gss_accept_sec_context(OM_uint32 *minor_status,
...
         if (!*context_handle) {
+                _gss_mg_log(10, "gss-asc: allocating new context");
```
nicowilliams, line 269: Do we want to log the pointer value for the context handle so as to be able to stitch together related calls?

lib/gssapi/mech/gss_import_name.c
```diff
@@ -230,6 +232,8 @@ gss_import_name(OM_uint32 *minor_status,
                                                name_type, output_name);
 	}
 
+	_gss_mg_log(10, "Importing name \"%*.s\"", (int)input_name_buffer->length,
```
nicowilliams, line 235: Watch out!  A name can be binary-like.

nicowilliams: (Names are allowed to have embedded NULs.

lib/gssapi/mech/gss_import_name.c
```diff
@@ -262,6 +266,8 @@ gss_import_name(OM_uint32 *minor_status,
 	HEIM_TAILQ_FOREACH(m, &_gss_mechs, gm_link) {
 		int present = 0;
 
+                _gss_mg_log(10, "attempting mech \"%s\"", m->gm_mech.gm_name);
```
nicowilliams, line 269: I like the `gdn` prefix you used in `gss_display_name()`.  Can you add a `gin` prefix here?

lib/gssapi/mech/gss_mech_switch.c
```diff
@@ -167,7 +167,7 @@ do {									\
 	m->gm_mech.gm_ ## name = (_gss_##name##_t *)dlsym(so, "gss_" #name); \
 	if (!m->gm_mech.gm_ ## name ||					\
 	    m->gm_mech.gm_ ##name == gss_ ## name) {			\
-		_gss_mg_log(1, "can't find symbol gss_" #name "\n");	\
+		_gss_mg_log(1, "can't find symbol gss_" #name);		\
```
nicowilliams, line 170: Do we want to log the name of the DSO we're looking in here?

elric1: Here, I was just eliminating the "\n" because it puts extra blank lines in. It's a separate commit from the other stuff.

lib/gssapi/mech/gss_accept_sec_context.c
elric1: I think that level 10 is supposed to be for tracing only.

lib/gssapi/mech/gss_import_name.c
elric1: Hmmm, what can we do here? base64 encode it? We must have a way to log these things for debugging and troubleshooting.

elric1: Used strvis

lib/gssapi/mech/gss_import_name.c
elric1: done.

lib/gssapi/mech/gss_mech_switch.c
elric1: done.

PR #1269, opened 2024-11-24
macOS: permit krb5_init_context to succeed if current working directory is inaccessible or invalid

jaltman:
CFReadStreamCreateWithFile() will fail due to the current working directory being invalid or inaccessible.  Instead of returning ENOMEM which is a fatal error, return EACCESS which permits the configuration parsing to continue without the plist.

For example, on macOS, if the cwd is in /afs and the user's tokens have expired, then krb5_init_context() fails with ENOMEM (12).

  user@MacBookAir user % ~/src/heimdal/kuser/heimtools klist
  shell-init: error retrieving current directory: \
    getcwd: cannot access parent directories: Permission denied
  chdir: error retrieving current directory: \
    getcwd: cannot access parent directories: Permission denied
  heimtools: krb5_init_context failed: 12

With this change krb5_init_context() succeeds.  It is important to note that the plist

  %{USERCONFIG}/Library/Preferences/com.apple.Kerberos.plist

is not guaranteed to exist.

### DIFF
--- a/lib/base/config_file.c
+++ b/lib/base/config_file.c
@@ -576,6 +576,7 @@ heim_config_parse_file_multi(heim_context context,
 {
     const char *str;
     char *newfname = NULL;
+    char *exp_fname = NULL;
     unsigned lineno = 0;
     heim_error_code ret = 0;
     struct fileptr f;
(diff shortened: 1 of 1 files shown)


jaltman:
This change as-is is not ideal because it returns EACCES regardless of the actual error.  
There are in fact two underlying problems leading to the krb5_init_context() failures when the current working directory is invalid.  

1. parse_plist_config() must not be called with unexpanded tokens such as %{USERCONFIG}.  An unexpanded token will be treated as a relative path and CFReadStreamCreateWithFile() calls getcwd() to obtain the current working directory to produce an absolute path.   The token expansion in heim_config_parse_file_multi() must apply to all files and exclude plist files.
2. If parse_plist_config() is called with a relative path and CFReadStreamCreateWithFile() fails, call getcwd() to obtain the actual error code to return to the caller.   If getcwd() succeeds, then return ENOMEM.

An updated patch will be pushed.

PR #1278, opened 2025-04-28
cf: remove comm_err header check to cross-compile heimdal

zeetim:
Currently, cf/check-compile-et.m4 is checking for a header file with hard coded path which make cross compilation unsafe. This commit allows to skip the test in the script.


nicowilliams:
I'll fix this when I finish the MinGW support.

PR #1279, opened 2025-04-28
Fix AR warning

zeetim:
I have noticed that `ar` utility generate a warning in error stream when building the project.
This commit solve the issue.


nicowilliams:
What is the warning?

PR #1303, opened 2026-01-22
gssmask: Disable Nagle.

riastradh:
This reduces the check-gssmask runtime from around 7min to 1-2sec on my laptop.

It is, perhaps, suboptimal to disable Nagle for a program that writes successive 4-byte units of a protocol message in separate syscalls rather than a single batch, and might be better to instead disable delayed acks, but:

(a) there's no portable API for disabling delayed acks, and
(b) this program appears to be used exclusively for testing anyway.

fix https://github.com/heimdal/heimdal/issues/1139

### DIFF
--- a/appl/gssmask/common.c
+++ b/appl/gssmask/common.c
@@ -95,3 +95,13 @@ permutate_all(struct getarg_strings *strings, size_t *size)
     free(list);
     return all;
 }
+
+void
+tcp_nodelay(rk_socket_t sock)
+{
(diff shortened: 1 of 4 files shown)


riastradh:
Might be worthwhile to revert 366016b1f6ceb760c99231b15033e4dacf5060c9 (tests: Speed up tests/gss/check-gssmask [by disabling one of three clients]) now!  (Not really sure what this test is about but it still runs in 1-2 seconds with three clients once Nagle is disabled.)

PR #1321, opened 2026-02-11
Implement strong and flexible certificate mappings

jsutton24:
Matching a certificate to a principal via an object SID, serial name, etc. where possible rather than the subject name means we are not prone to various spoofing attacks. See https://support.microsoft.com/en-us/topic/kb5014754-certificate-based-authentication-changes-on-windows-domain-controllers-ad2c23b0-15d8-4340-a468-4d4f3b188f16

Includes one patch ([kdc: support pkinit_kdc_revoke for pkinit anchors](https://github.com/heimdal/heimdal/commit/abe9db44d8010c3ef847dce41d304bb0aa0c5328)) which may be removed if needed.

### DIFF
--- a/lib/hx509/cert.c
+++ b/lib/hx509/cert.c
@@ -1062,19 +1062,33 @@ hx509_verify_ctx_f_allow_best_before_signature_algs(hx509_context ctx,
 	ctx->flags |= HX509_VERIFY_CTX_F_NO_BEST_BEFORE_CHECK;
 }
 
-static const Extension *
-find_extension(const Certificate *cert, const heim_oid *oid, size_t *idx)
+static int
+find_extension(const Certificate *cert, const heim_oid *oid, size_t *idx,
+	       const Extension **ep)
 {
     const TBSCertificate *c = &cert->tbsCertificate;
+    size_t i;
+    int found = 0;
+
+    *ep = NULL;
 
     if (c->version == NULL || *c->version < 2 || c->extensions == NULL)
-	return NULL;
+	return HX509_EXTENSION_NOT_FOUND;
 
-    for (;*idx < c->extensions->len; (*idx)++) {
-	if (der_heim_oid_cmp(&c->extensions->val[*idx].extnID, oid) == 0)
-	    return &c->extensions->val[(*idx)++];
+    for (i = *idx; i < c->extensions->len; i++) {
+	if (der_heim_oid_cmp(&c->extensions->val[i].extnID, oid) == 0) {
+	    if (!found) {
+		*idx = i + 1;
+		*ep = &c->extensions->val[i];
+	    }
+	    found++;
+	}
     }
-    return NULL;
+    switch (found) {
+    case 0:	return HX509_EXTENSION_NOT_FOUND;
+    case 1:	return 0;
+    default:	return HX509_CERTIFICATE_MALFORMED;
+    }
 }
 
 static int
@@ -1084,12 +1098,13 @@ find_extension_auth_key_id(const Certificate *subject,
     const Extension *e;
     size_t size;
     size_t i = 0;
+    int ret;
 
     memset(ai, 0, sizeof(*ai));
 
-    e = find_extension(subject, &asn1_oid_id_x509_ce_authorityKeyIdentifier, &i);
-    if (e == NULL)
-	return HX509_EXTENSION_NOT_FOUND;
+    ret = find_extension(subject, &asn1_oid_id_x509_ce_authorityKeyIdentifier, &i, &e);
+    if (ret)
+	return ret;
 
     return decode_AuthorityKeyIdentifier(e->extnValue.data,
 					 e->extnValue.length,
@@ -1103,12 +1118,13 @@ _hx509_find_extension_subject_key_id(const Certificate *issuer,
     const Extension *e;
     size_t size;
     size_t i = 0;
+    int ret;
 
     memset(si, 0, sizeof(*si));
 
-    e = find_extension(issuer, &asn1_oid_id_x509_ce_subjectKeyIdentifier, &i);
-    if (e == NULL)
-	return HX509_EXTENSION_NOT_FOUND;
+    ret = find_extension(issuer, &asn1_oid_id_x509_ce_subjectKeyIdentifier, &i, &e);
+    if (ret)
+	return ret;
 
     return decode_SubjectKeyIdentifier(e->extnValue.data,
 				       e->extnValue.length,
@@ -1122,12 +1138,13 @@ find_extension_name_constraints(const Certificate *subject,
     const Extension *e;
     size_t size;
     size_t i = 0;
+    int ret;
 
     memset(nc, 0, sizeof(*nc));
 
-    e = find_extension(subject, &asn1_oid_id_x509_ce_nameConstraints, &i);
-    if (e == NULL)
-	return HX509_EXTENSION_NOT_FOUND;
+    ret = find_extension(subject, &asn1_oid_id_x509_ce_nameConstraints, &i, &e);
+    if (ret)
+	return ret;
 
     return decode_NameConstraints(e->extnValue.data,
 				  e->extnValue.length,
@@ -1140,12 +1157,13 @@ find_extension_subject_alt_name(const Certificate *cert, size_t *i,
 {
     const Extension *e;
     size_t size;
+    int ret;
 
     memset(sa, 0, sizeof(*sa));
 
-    e = find_extension(cert, &asn1_oid_id_x509_ce_subjectAltName, i);
-    if (e == NULL)
-	return HX509_EXTENSION_NOT_FOUND;
+    ret = find_extension(cert, &asn1_oid_id_x509_ce_subjectAltName, i, &e);
+    if (ret)
+	return ret;
 
     return decode_GeneralNames(e->extnValue.data,
 			       e->extnValue.length,
@@ -1158,12 +1176,13 @@ find_extension_eku(const Certificate *cert, ExtKeyUsage *eku)
     const Extension *e;
     size_t size;
     size_t i = 0;
+    int ret;
 
     memset(eku, 0, sizeof(*eku));
 
-    e = find_extension(cert, &asn1_oid_id_x509_ce_extKeyUsage, &i);
-    if (e == NULL)
-	return HX509_EXTENSION_NOT_FOUND;
+    ret = find_extension(cert, &asn1_oid_id_x509_ce_extKeyUsage, &i, &e);
+    if (ret)
+	return ret;
 
     return decode_ExtKeyUsage(e->extnValue.data,
 			      e->extnValue.length,
@@ -1181,9 +1200,9 @@ find_extension_ntds_security_extension(const Certificate *subject,
 
     memset(sid_extension, 0, sizeof(*sid_extension));
 
-    e = find_extension(subject, &asn1_oid_szOID_NTDS_CA_SECURITY_EXT, &i);
-    if (e == NULL)
-        return HX509_EXTENSION_NOT_FOUND;
+    ret = find_extension(subject, &asn1_oid_szOID_NTDS_CA_SECURITY_EXT, &i, &e);
+    if (ret)
+        return ret;
 
     ret = decode_SidExtension(e->extnValue.data, e->extnValue.length,
                               sid_extension, &size);
@@ -1267,7 +1286,6 @@ hx509_cert_find_subjectAltName_otherName(hx509_context context,
     i = 0;
     while (1) {
 	ret = find_extension_subject_alt_name(_hx509_get_cert(cert), &i, &sa);
-	i++;
 	if (ret == HX509_EXTENSION_NOT_FOUND) {
 	    return 0;
 	} else if (ret != 0) {
@@ -1326,7 +1344,6 @@ hx509_cert_find_subjectAltName_rfc822(hx509_context context,
     i = 0;
     while (1) {
 	ret = find_extension_subject_alt_name(_hx509_get_cert(cert), &i, &sa);
-	i++;
 	if (ret == HX509_EXTENSION_NOT_FOUND) {
 	    return 0;
 	} else if (ret != 0) {
@@ -1387,7 +1404,6 @@ hx509_cert_get_san(hx509_context context,
 
     while (1) {
 	ret = find_extension_subject_alt_name(_hx509_get_cert(cert), &i, &sa);
-	i++;
 	if (ret == HX509_EXTENSION_NOT_FOUND)
 	    return HX509_NO_ITEM;
 	if (ret != 0)
@@ -1427,12 +1443,13 @@ hx509_cert_get_subject_key_identifier(hx509_context context,
     const Extension *e;
     size_t size;
     size_t i = 0;
+    int ret;
 
-    e = find_extension(cert->data,
+    ret = find_extension(cert->data,
 		       &asn1_oid_id_x509_ce_subjectKeyIdentifier,
-		       &i);
-    if (e == NULL)
-	return HX509_EXTENSION_NOT_FOUND;
+		       &i, &e);
+    if (ret)
+	return ret;
 
     return decode_SubjectKeyIdentifier(e->extnValue.data,
 				       e->extnValue.length,
@@ -1454,8 +1471,8 @@ check_key_usage(hx509_context context, const Certificate *cert,
     if (_hx509_cert_get_version(cert) < 3)
 	return 0;
 
-    e = find_extension(cert, &asn1_oid_id_x509_ce_keyUsage, &i);
-    if (e == NULL) {
+    ret = find_extension(cert, &asn1_oid_id_x509_ce_keyUsage, &i, &e);
+    if (ret == HX509_EXTENSION_NOT_FOUND) {
 	if (req_present) {
 	    hx509_set_error_string(context, 0, HX509_KU_CERT_MISSING,
 				   "Required extension key "
@@ -1464,6 +1481,8 @@ check_key_usage(hx509_context context, const Certificate *cert,
 	}
 	return 0;
     }
+    if (ret)
+	return ret;
 
     ret = decode_KeyUsage(e->extnValue.data, e->extnValue.length, &ku, &size);
     if (ret)
@@ -1515,8 +1534,8 @@ check_basic_constraints(hx509_context context, const Certificate *cert,
     if (_hx509_cert_get_version(cert) < 3)
 	return 0;
 
-    e = find_extension(cert, &asn1_oid_id_x509_ce_basicConstraints, &i);
-    if (e == NULL) {
+    ret = find_extension(cert, &asn1_oid_id_x509_ce_basicConstraints, &i, &e);
+    if (ret == HX509_EXTENSION_NOT_FOUND) {
 	switch(type) {
 	case PROXY_CERT:
 	case EE_CERT:
@@ -1533,6 +1552,8 @@ check_basic_constraints(hx509_context context, const Certificate *cert,
 	}
 	}
     }
+    if (ret)
+	return ret;
 
     ret = decode_BasicConstraints(e->extnValue.data,
 				  e->extnValue.length, &bc,
@@ -1711,11 +1732,13 @@ hx509_cert_is_ca(hx509_context context,
     if (_hx509_cert_get_version(c->data) < 3)
         return certificate_is_self_signed(context, c->data, is_ca);
 
-    e = find_extension(c->data, &asn1_oid_id_x509_ce_basicConstraints, &i);
-    if (e == NULL) {
+    ret = find_extension(c->data, &asn1_oid_id_x509_ce_basicConstraints, &i, &e);
+    if (ret == HX509_EXTENSION_NOT_FOUND) {
         *is_ca = 0;
         return 0;
     }
+    if (ret)
+        return ret;
 
     ret = decode_BasicConstraints(e->extnValue.data,
 				  e->extnValue.length, &bc,
@@ -1865,10 +1888,10 @@ is_proxy_cert(hx509_context context,
     if (rinfo)
 	memset(rinfo, 0, sizeof(*rinfo));
 
-    e = find_extension(cert, &asn1_oid_id_pkix_pe_proxyCertInfo, &i);
-    if (e == NULL) {
+    ret = find_extension(cert, &asn1_oid_id_pkix_pe_proxyCertInfo, &i, &e);
+    if (ret) {
 	hx509_clear_error_string(context);
-	return HX509_EXTENSION_NOT_FOUND;
+	return ret;
     }
 
     ret = decode_ProxyCertInfo(e->extnValue.data,
@@ -2916,6 +2939,7 @@ hx509_verify_path(hx509_context context,
 	    ProxyCertInfo info;
 
 	    if (is_proxy_cert(context, c, &info) == 0) {
+		const Extension *e;
 		size_t j;
 
 		if (info.pCPathLenConstraint != NULL &&
@@ -2932,22 +2956,26 @@ hx509_verify_path(hx509_context context,
 		free_ProxyCertInfo(&info);
 
 		j = 0;
-		if (find_extension(c, &asn1_oid_id_x509_ce_subjectAltName, &j)) {
+		ret = find_extension(c, &asn1_oid_id_x509_ce_subjectAltName, &j, &e);
+		if (ret == 0) {
 		    ret = HX509_PROXY_CERT_INVALID;
 		    hx509_set_error_string(context, 0, ret,
 					   "Proxy certificate has explicitly "
 					   "forbidden subjectAltName");
 		    goto out;
-		}
+		} else if (ret != HX509_EXTENSION_NOT_FOUND)
+		    goto out;
 
 		j = 0;
-		if (find_extension(c, &asn1_oid_id_x509_ce_issuerAltName, &j)) {
+		ret = find_extension(c, &asn1_oid_id_x509_ce_issuerAltName, &j, &e);
+		if (ret == 0) {
 		    ret = HX509_PROXY_CERT_INVALID;
 		    hx509_set_error_string(context, 0, ret,
 					   "Proxy certificate has explicitly "
 					   "forbidden issuerAltName");
 		    goto out;
-		}
+		} else if (ret != HX509_EXTENSION_NOT_FOUND)
+		    goto out;
 
 		/*
 		 * The subject name of the proxy certificate should be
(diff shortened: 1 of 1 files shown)


kdc/kerberos5.c
```diff
@@ -536,7 +537,15 @@ pa_pkinit_validate(astgs_request_t r, const PA_DATA *pa)
 
     ret = _kdc_pk_rd_padata(r, pa, &pkp);
     if (ret || pkp == NULL) {
-	ret = KRB5KRB_AP_ERR_BAD_INTEGRITY;
+	if (ret == HX509_CERT_REVOKED ||
+	    ret == KRB5_KDC_ERR_CLIENT_NOT_TRUSTED) {
+
+	    ret = KRB5_KDC_ERR_CLIENT_NOT_TRUSTED;
+	} else if (ret == KRB5_KDC_ERR_CERTIFICATE_MISMATCH) {
+	    ret = KRB5_KDC_ERR_CERTIFICATE_MISMATCH;
```
lhoward: I would prefer the next line to be `else if (ret != KRB5_KDC_ERR_CERTIFICATE_MISMATCH)`

jsutton24:
Fixed the indentation.

nicowilliams:
While I'm not opposed to supporting matching on more things than the UPN and PKINIT SANs, and/or DN, and while I would understand the motivation being to have equivalent functionality to what AD has, I don't quite understand the stated functionality.

IIUC the issue is that matching on _DN_ is susceptible to spoofing where a PKI admin doesn't understand the attack by someone requesting an innocent-looking DN.  But if there's more to the spoofing concern I'd like to understand it.

And to be clear the obvious way to match PKINIT client certificates is by PKINIT SAN -- this should be sufficient.  And _of course_ one should run a separate PKI (root CA) for PKINIT clients for any given realm.

lib/hx509/cert.c
```diff
@@ -2174,6 +2260,87 @@ hx509_cert_get_subject_unique_id(hx509_context context, hx509_cert p, heim_bit_s
...
+    if (ret) {
+        return ret;
+    }
+
+    if (sid_ext.len != 1) {
+        return HX509_CMS_INVALID_DATA;
+    }
+
+    if (sid_ext.val[0].element != choice_GeneralName_otherName) {
+        return HX509_CMS_INVALID_DATA;
```
nicowilliams: Leaks `sig_ext`.

lib/hdb/version-script.map
```diff
@@ -159,6 +159,10 @@ HEIMDAL_HDB_1.0 {
 		free_HDB_extensions;
 		free_HDB_Ext_KeyRotation;
 		free_HDB_Ext_KeySet;
+		free_HDB_Ext_KeyTrust;
+		free_HDB_Ext_KeyTrust;
```
nicowilliams: Duplicate.

kdc/pkinit.c
```diff
@@ -625,6 +626,461 @@ check_dh_param(krb5_context context,
...
+    if (ret != 0) {
+        return FALSE;
+    }
+
+    ts = calloc(target->length + 1, sizeof(char));
+    if (ts == NULL) {
+        goto out;
+    }
+    memcpy(ts, target->data, target->length);
+    if (strncmp(ts, ns, target->length) == 0) {
```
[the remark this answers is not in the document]
jsutton24: Possibly. Would `strcmp()` work as an alternative in the meantime?

kdc/pkinit.c
```diff
@@ -965,7 +1423,62 @@ _kdc_pk_rd_padata(astgs_request_t priv,
 				      &eContentType,
 				      &eContent,
 				      &signer_certs);
-	if (ret) {
+        if (ret == 0) {
```
nicowilliams: Since in the error case we were doing `goto out` and in the interest of not indenting so much, I propose that you check the two error cases you're interested in then after handling that check for all errors.

Like this:

```
@@ -1423,21 +1423,8 @@ _kdc_pk_rd_padata(astgs_request_t priv,
                                      &eContentType,
                                      &eContent,
                                      &signer_certs);
-        if (ret == 0) {
-            if (signer_certs) {
-                ret = hx509_get_one_cert(
-                    context->hx509ctx, signer_certs, &cp->cert);
-                hx509_certs_free(&signer_certs);
-            }
-            if (ret) {
-                goto out;
-            }
-            ret = pk_check_certificate_binding(context, client, &cp->cert);
-            if (ret) {
-                goto out;
-            }
-        } else if (ret == HX509_CMS_NO_RECIPIENT_CERTIFICATE ||
-                   ret == HX509_ISSUER_NOT_FOUND) {
+        if (ret == HX509_CMS_NO_RECIPIENT_CERTIFICATE ||
+            ret == HX509_ISSUER_NOT_FOUND) {
             /*
              * Certificate not in the chain of trust,
              * however it could be a self signed certificate for key trust
@@ -1478,7 +1465,8 @@ _kdc_pk_rd_padata(astgs_request_t priv,
             if (ret) {
                 goto out;
             }
-        } else {
+        }
+        if (ret) {
            char *s = hx509_get_error_string(context->hx509ctx, ret);
            krb5_warnx(context, "PKINIT: failed to verify signature: %s: %d",
                       s, ret);
@@ -1533,6 +1521,19 @@ _kdc_pk_rd_padata(astgs_request_t priv,
             }
            goto out;
        }
+
+        if (signer_certs) {
+            ret = hx509_get_one_cert(
+                                     context->hx509ctx, signer_certs, &cp->cert);
+            hx509_certs_free(&signer_certs);
+        }
+        if (ret) {
+            goto out;
+        }
+        ret = pk_check_certificate_binding(context, client, &cp->cert);
+        if (ret) {
+            goto out;
+        }
     }
 
     /* Signature is correct, now verify the signed message */
```

kdc/pkinit.c
```diff
@@ -965,7 +1423,62 @@ _kdc_pk_rd_padata(astgs_request_t priv,
...
+            }
+        } else if (ret == HX509_CMS_NO_RECIPIENT_CERTIFICATE ||
+                   ret == HX509_ISSUER_NOT_FOUND) {
+            /*
+             * Certificate not in the chain of trust,
+             * however it could be a self signed certificate for key trust
+             * logon.
+            */
+            int f = flags;
+            f |= HX509_CMS_VS_NO_VALIDATE;
```
nicowilliams: I'd much rather introduce a new flag that says "self-signed certs are OK".  I'm not sure yet that there isn't a weird interplay between `HX509_CMS_NO_RECIPIENT_CERTIFICATE` and `HX509_CMS_VS_NO_VALIDATE`.

kdc/pkinit.c
```diff
@@ -625,6 +626,461 @@ check_dh_param(krb5_context context,
...
+    }
+
+    /*
+     * Compute the SHA1 hash of the certificate subject public key
+     */
+    ctx = EVP_MD_CTX_create();
+    if (ctx == NULL) {
+        goto out;
+    }
+    EVP_DigestInit_ex(ctx, EVP_sha1(), NULL);
```
nicowilliams: We have caching of "fetchable" algorithms now in the `krb5_context`'s `ossl_context`.  Please use that.

Also, we should support no-hashing and alternative hash algorithms because SHA-1 is not really safe for this sort of thing.  What does Microsoft use?  Is it really SHA-1?  Do they have no plans to improve this?

lib/hdb/hdb.asn1
```diff
@@ -192,6 +193,35 @@ KeyRotation ::= SEQUENCE {
...
+	serial-number[3]	OCTET STRING OPTIONAL,
+	ski[4]			OCTET STRING OPTIONAL,
+	public-key[5]		OCTET STRING OPTIONAL,
+	rfc822[6]		OCTET STRING OPTIONAL,
+	...
+}
+
+HDB-Ext-EnforcementMode ::= ENUMERATED {
+	hdb_enf_mode_none		(0),
+	hdb_enf_mode_compatibility	(1),
```
nicowilliams: Commentary on this would be nice.

nicowilliams:
In general I'd really like tests.  And I'd really like not just the KDC side of this but the admin tool side of this too so that we can build tests.  Here that means `kadmin modprinc ...` to get these extensions in, but... yeah, all such work on kadmin and `kadm5` is insufferably annoying and time consuming.  Fortunately Claude or some other LLM can help.

kdc/pkinit.c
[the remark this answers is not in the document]
nicowilliams: Yes, that would work.

lib/hdb/hdb.asn1
```diff
@@ -192,6 +193,35 @@ KeyRotation ::= SEQUENCE {
...
+	subject-name[2]		OCTET STRING OPTIONAL,
+	serial-number[3]	OCTET STRING OPTIONAL,
+	ski[4]			OCTET STRING OPTIONAL,
+	public-key[5]		OCTET STRING OPTIONAL,
+	rfc822[6]		OCTET STRING OPTIONAL,
+	...
+}
+
+HDB-Ext-EnforcementMode ::= ENUMERATED {
+	hdb_enf_mode_none		(0),
```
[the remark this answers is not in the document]
jsutton24: It disables all the certificate mapping logic. I guess omitting the `HDB_Ext_CertificateMappings` structure would do the same thing though.

lib/hdb/hdb.asn1
```diff
@@ -192,6 +193,35 @@ KeyRotation ::= SEQUENCE {
 
 HDB-Ext-KeyRotation ::= SEQUENCE SIZE (1..3) OF KeyRotation
 
+HDB-Ext-KeyTrust ::= SEQUENCE OF SEQUENCE {
+	pub_key[0] OCTET STRING
+}
+
+HDB-Ext-CertificateMapping ::= SEQUENCE {
+	strong-mapping[0]	BOOLEAN,
```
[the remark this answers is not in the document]
nicowilliams: Right, but my question is whether should bother storing the strength as a field here given that we can always work out the matching rules' strength by looking at them.

nicowilliams:
Does [this](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-pkca/282ed46a-97c2-4fab-8456-a6bd67b9ba71) imply we must have a bunch of required rules, right?

nicowilliams:
If you don't have the cycles to make this in-tree testable, if you would point me to Samba tree tests, I might be happy with that + in-tree commentary to that effect, and then I might do the rest of the work in Heimdal.  But I really want to get more serious about test coverage -- in the age of LLMs we really can do much better.

jsutton24:
> Is my reading correct that SID compression only allows compression of SIDs under the principal's domain's SID? If so that doesn't really help if a user moves from one domain to another and/or otherwise ends up with lots of group memberships from other domains.

Was this comment meant for https://github.com/krb5/krb5/pull/1476?

jsutton24:
> The Windows build fails because you need to export more symbols now in `lib/hdb/libhdb-exports.def`:
> 
> ```
> asn1_hdb_asn1.obj : error LNK2019: unresolved external symbol decode_ObjectSid referenced in function decode_HDB_extension
> asn1_hdb_asn1.obj : error LNK2019: unresolved external symbol encode_ObjectSid referenced in function encode_HDB_extension
> asn1_hdb_asn1.obj : error LNK2019: unresolved external symbol length_ObjectSid referenced in function length_HDB_extension
> asn1_hdb_asn1.obj : error LNK2019: unresolved external symbol copy_ObjectSid referenced in function copy_HDB_extension
> asn1_hdb_asn1.obj : error LNK2019: unresolved external symbol free_ObjectSid referenced in function free_HDB_extension
> ```

Do I need to do something more than adding those symbols to lib/hdb/libhdb-exports.def? Because that hasn’t seemed to work.

nicowilliams:
Ok, to recap my questions:

- Does AD allow one to specify a set of more than one matching rules for a principal/user, where all the rules are `AND`ed?  Does one have to specify _every_ possible matching, or "one or more"?
  
  Is there a screenshot of what this looks like in their UI?  I'm not asking what their docs say about the _protocol_, just the UI.
- If the answer to the above is "yes" then `pk_check_certificate_binding()` should count matches and non-matches in the loop, and after the loop the success condition would then be `(matched_count && !nonmatch_count)`.
- Does AD have a per-user `enforcement-mode`?  Or is it a global domain configuration parameter?  Should we do the same?
  
  If we make it a global config parameter then we don't need it in the HDB entry.
- `strong-mapping` can be derived from inspecting the rules, so I think we don't want it as a member of `HDB-Ext-CertificateMappings`
  
  (Otherwise if `strong-mapping` is set incorrectly in the HDB entry extension, then what??)
- Should we not move handling of `HDB-Ext-PKINIT-acl`, `HDB-Ext-PKINIT-hash`, and `HDB-Ext-PKINIT-cert` into the same function that implements `HDB-Ext-CertificateMappings`?  The former are handled in `_kdc_pk_check_client()`.  Unless there is a good reason why these two ways of matching certificates are very different, I think they should be unified.
  
  The new thing is N>=1 rules for one "PKINIT ACL" while the old thing is N==1 rule for one "PKINIT ACL".  But apart from this it seems like the same thing to me.

Once this is ready to merge I might add kadmin/kadm5 bindings for this and a test.

nicowilliams:
Here's part of a Claude response to a prompt about this PR:

```
  You're right to be skeptical. Here's how the existing and new mechanisms relate:                                                                            
                                                                                                                                                              
  Redundancy with existing HDB-Ext-PKINIT-* types                                                                                                             
                                                                                                                                                              
  The existing PKINIT authorization model (in _kdc_pk_check_client(), line 2570) already has a comprehensive pipeline:                                        
                                                                                                                                                              
  1. HDB-Ext-PKINIT-cert (choice 8) -- exact certificate match (full DER comparison)                                                                          
  2. pkinit_princ_in_cert -- match PKINIT SAN or MS-UPN SAN in the certificate against the client principal                                                   
  3. HDB-Ext-PKINIT-acl (choice 0) -- subject DN string match (with optional issuer/anchor, though those are currently continued over)                        
  4. principal_mappings -- file-based subject-to-principal mapping                                                                                            
                                       
  The new HDB-Ext-CertificateMapping (choice 17) duplicates several of these:
  ┌─────────────────────────┬───────────────────────────────────────────────────────────────────────────────┐
  │        New field        │                              Existing equivalent                              │
  ├─────────────────────────┼───────────────────────────────────────────────────────────────────────────────┤                                             
  │ subject-name            │ HDB-Ext-PKINIT-acl.subject                                                    │                                           
  ├─────────────────────────┼───────────────────────────────────────────────────────────────────────────────┤
  │ issuer-name             │ HDB-Ext-PKINIT-acl.issuer (unused but present)                                │
  ├─────────────────────────┼───────────────────────────────────────────────────────────────────────────────┤
  │ public-key (SHA-1 hash) │ HDB-Ext-PKINIT-hash (choice 1, which already supports arbitrary digest OIDs!) │
  ├─────────────────────────┼───────────────────────────────────────────────────────────────────────────────┤
  │ serial-number           │ no direct equivalent                                                          │
  ├─────────────────────────┼───────────────────────────────────────────────────────────────────────────────┤
  │ ski                     │ no direct equivalent                                                          │
  ├─────────────────────────┼───────────────────────────────────────────────────────────────────────────────┤
  │ rfc822                  │ no direct equivalent                                                          │
  └─────────────────────────┴───────────────────────────────────────────────────────────────────────────────┘
  The HDB-Ext-PKINIT-hash type is particularly notable -- it already stores (digest-type OID, digest) pairs, which is strictly more general than the
  SHA-1-only public-key field in the new mapping type. If you wanted SHA-256 public key hashes, you could use HDB-Ext-PKINIT-hash today.

  And the HDB-Ext-KeyTrust (choice 16) is essentially a stripped-down version of HDB-Ext-PKINIT-cert (choice 8) -- it stores public keys where PKINIT-cert
  stores whole certificates.
  The architectural problem

  The new code runs at two different points in the flow, creating a confusing dual-authorization model:

  1. pk_check_certificate_binding() runs inside _kdc_pk_rd_padata() (line 1435), before _kdc_pk_check_client() is called -- it's a gate on CMS signature
  verification
  2. _kdc_pk_check_client() then runs separately as principal authorization

  So now there are two independent cert-to-principal binding checks at different layers, using different HDB extensions, with overlapping semantics. A
  deployment would need to populate both the existing PKINIT-acl/cert/hash extensions AND the new CertificateMappings extension to get correct behavior.
  That's confusing and error-prone.

  On SHA-1 collision feasibility

  Your instinct is right to worry. The match_public_key() function hashes the raw subjectPublicKey BIT STRING value from the SPKI. An attacker who can get
  their own certificate issued (which is the threat model for KB5014754 -- compromised CA) controls their public key bits entirely. With a chosen-prefix
  collision attack on SHA-1 (practical since ~2020, see SHA-1 is a Shambles), the attacker could:

  1. Observe the victim's SPKI SHA-1 hash (stored in the HDB mapping)
  2. Generate a key pair whose SPKI produces the same SHA-1
  3. Get a certificate for that key from the compromised CA

  This is a second preimage attack, which is harder than a collision. But: if the attacker can influence the provisioning of both the mapping and their own
  certificate (e.g., during initial enrollment in a compromised environment), then it's a plain collision attack, which is feasible for SHA-1. The arbitrary
  extensions point is about whole-certificate collisions; here the collision target is just the SPKI, which is more constrained but still attacker-controlled
  in the key bits. Bottom line: SHA-1 is marginal here. If the spec mandates it, fine, but it should at minimum be documented as a known limitation.

  On N>1 sub-mappings

  I agree this is suspect. The Microsoft documentation for KB5014754 describes mapping a single certificate to a single user account, using one of several
  mapping types (Strong: SKI+serial, SHA-1 of public key, explicit 1:1; Weak: subject, issuer, SAN). It doesn't describe a single mapping entry with multiple
  fields ANDed together. The schema in AD uses altSecurityIdentities which is a multi-valued string attribute where each value is one mapping of one type
  (e.g., X509:<SKI>abcd1234<SR>issuer<S>serial or X509:<SHA1-PUKEY>hash). So each AD value is a single-criterion match, not a conjunction.

  The HDB design here -- a SEQUENCE OF mappings where each mapping has multiple optional fields that are (intended to be) ANDed -- doesn't match the AD model.
   It would be simpler and more correct to model this as Microsoft does: each mapping is a single type with a single value, and you OR across the list.
```

The SHA-1 concern was a concern that Claude had brought up earlier in a code review it did.  I think SHA-1 is definitely too weak, so I'm not impressed that Microsoft uses SHA-1.  Here's an opportunity for Samba to do better, no?

Anyways, maybe Claude is hallucinating, but I'm following up and near as I can see `altSecurityIdentities` is indeed a list of values each of which is a single matching rule, and only one needs match.  So...

 - How about we just add new `HDB-Ext-PKINIT-*` types, one for each matching rule type that Heimdal does not already support, not including the SHA-1 case because that's already covered by `HDB-Ext-PKINIT-hash` (which Samba would have to synthesize from the `altSecurityIdentities`).

This approach would make your PR much simpler, and it would make it much easier to finish the `kadmin`/`libkadm5` side of things and write a test.

nicowilliams:
So I propose that we mostly start over and:

- Add `HDB-Ext-PKINIT-*` HDB entry extensions for all the mapping types we don't already have:
  - `HDB-Ext-PKINIT-issuer-serial`
  - `HDB-Ext-PKINIT-object-sid`
  - `HDB-Ext-PKINIT-ski` (but given that there are CAs that blindly copy the SKI from CSRs I think it's best to leave this one out since that makes this kind "weak")
  - `HDB-Ext-PKINIT-rfc822` (considered "weak" though, so maybe we should leave this out)
- Add `[kdc]` `pkinit_strong_cert_cutoff` datetime config parameter
- Add `[kdc]` `pkinit_strong_cert_binding` boolean which, if `true` then the KDC will ignore weak binding types except for certs before the `pkinit_strong_cert_cutoff` (if set) or always (if `pkinit_strong_cert_cutoff` not set)

This should be much simpler to implement.  Mind if I let Claude try it?

jsutton24:
Sure, go ahead.

jsutton24:
> > > The Windows build fails because you need to export more symbols now in `lib/hdb/libhdb-exports.def`:
> > > ```
> > > asn1_hdb_asn1.obj : error LNK2019: unresolved external symbol decode_ObjectSid referenced in function decode_HDB_extension
> > > asn1_hdb_asn1.obj : error LNK2019: unresolved external symbol encode_ObjectSid referenced in function encode_HDB_extension
> > > asn1_hdb_asn1.obj : error LNK2019: unresolved external symbol length_ObjectSid referenced in function length_HDB_extension
> > > asn1_hdb_asn1.obj : error LNK2019: unresolved external symbol copy_ObjectSid referenced in function copy_HDB_extension
> > > asn1_hdb_asn1.obj : error LNK2019: unresolved external symbol free_ObjectSid referenced in function free_HDB_extension
> > > ```
> > 
> > 
> > Do I need to do something more than adding those symbols to lib/hdb/libhdb-exports.def? Because that hasn’t seemed to work.
> 
> Oh, you also have to edit `lib/hdb/NTMakefile` to build the new files.

Do you mean `lib/hdb/Makefile.am`? Only because I don’t see anywhere in `lib/hdb/NTMakefile` where the new files would go.

nicowilliams:
> Do you mean `lib/hdb/Makefile.am`? Only because I don’t see anywhere in `lib/hdb/NTMakefile` where the new files would go.

Oh right, for the Windows build we use `asn1_compile --one-code-file`.  Hmm.

nicowilliams:
> Sure, go ahead.

That a look [here](https://github.com/nicowilliams/heimdal/tree/pr1321-Claude-version).  I've not reviewed this, and I'm not even sure that it covers all your needs, but it's done by extending the existing certificate matching.  Also, I don't like this idea that Microsoft's experience as to what is a "strong" or "weak" matching will always be so and representative of every site's experience, so in the head commit I asked Claude to make the strength of each matching type configurable by issuer.

One thing I've noticed is that some of the code here resembles yours, so I'll have to fix the attribution in the commits.  And since it's all Claude-coded I should make the commits say `Authored-By: ...`, not `Co-Authored-By: ...`.

jsutton24:
The code seems good. I would run it through Samba’s test suite but we haven’t done the migration over to Heimdal 8.0 yet. But if the in‐tree tests are passing that’s good news.

nicowilliams:
Alright, I'll create a new PR, get it passing all builds, and we can review it.  Thanks for your initial review!

nicowilliams:
@jsutton24 @garyWL I've done a bunch more work on this with Claude and force-pushed.  Can you review this?

~I've accidentally dropped `valid_certificate_start`, but I'll re-add it as a configuration parameter, settable both, in `[kdc]` and on a per root CA basis.~  (fixed)

lib/hdb/hdb.asn1
```diff
@@ -192,6 +193,35 @@ KeyRotation ::= SEQUENCE {
...
+HDB-Ext-EnforcementMode ::= ENUMERATED {
+	hdb_enf_mode_none		(0),
+	hdb_enf_mode_compatibility	(1),
+	hdb_enf_mode_full		(2),
+	...
+}
+
+HDB-Ext-CertificateMappings ::= SEQUENCE {
+	enforcement_mode[0]	    HDB-Ext-EnforcementMode,
+	valid_certificate_start[1]  KerberosTime,
```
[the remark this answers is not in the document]
nicowilliams: I ~think I might have~ dropped this feature.  I'll re-add it.

nicowilliams:
(I'll fix the Windows build in the next couple of days.)

nicowilliams:
Most PKINIT "ACLs" -matching rules- in HDB entries should have an option to match a specific trust anchor or issuer.  Or at least they should be extensible so we can add that option later.  EDIT: Added `anchor` where sensible, though it's not implemented at this time.

lib/hx509/cert.c
```diff
@@ -1353,6 +1353,58 @@ hx509_cert_find_subjectAltName_rfc822(hx509_context context,
...
+{
+    GeneralNames sa;
+    size_t i = 0, k = 0, j;
+    int ret;
+
+    *out = NULL;
+
+    while (1) {
+	ret = find_extension_subject_alt_name(_hx509_get_cert(cert), &i, &sa);
+	i++;
```
jsutton24: There are a few of these `i++`s that don’t look right.

jsutton24:
This looks good!

`valid_certificate_start` seems to work differently now. It used to be per‐principal (so that each principal has its own certificate validation period) but now it’s global or per‐CA.

kdc/pkinit.c
```diff
@@ -2148,26 +2328,166 @@ _kdc_pk_check_client(astgs_request_t r,
...
+                const unsigned char *pkey_data;
+                size_t pkey_len;
+
+                pkey_data = spki.subjectPublicKey.data;
+                pkey_len = spki.subjectPublicKey.length / 8;
+
+                EVP_Digest(pkey_data, pkey_len, sha1_buf, &sha1_len,
+                           r->context->ossl->sha1, NULL);
+                EVP_Digest(pkey_data, pkey_len, sha256_buf, &sha256_len,
+                           r->context->ossl->sha256, NULL);
```
jsutton24, line 2456: We found that we should be matching on the certificate digest, rather than on the hash of the public key. So we probably want to call `X509_digest_sig(cp->cert, r->context->ossl->sha1 /* or sha256 */, NULL)` instead.

jsutton24:
I will look more tomorrow.

lib/hx509/cert.c
nicowilliams: I looked at it and it's not wrong.  Normally we'd write a `for` loop here, but this works.

nicowilliams:
> `valid_certificate_start` seems to work differently now. It used to be per‐principal (so that each principal has its own certificate validation period) but now it’s global or per‐CA.

Yes, but I think that's what we'd want.  What does Microsoft do in AD?

kdc/pkinit.c
nicowilliams: I see.  Maybe we should have a separate extension for matching on hash of SPKI and matching on hash of certificate?  We could also check both digests in the KDC, and in `kadmin` use a digest of the certificate.

lib/hx509/cert.c
jsutton24: Doesn’t `i++` happen inside `find_extension_subject_alt_name()`? Meaning `i` is incremented twice?

jsutton24:
> > `valid_certificate_start` seems to work differently now. It used to be per‐principal (so that each principal has its own certificate validation period) but now it’s global or per‐CA.
> 
> Yes, but I think that's what we'd want. What does Microsoft do in AD?

In AD the weakly‐mapped certificate is valid from when the AD account for the principal was created, minus an administrator‐configurable backdating compensation offset.

jsutton24:
Apart from the few points I raised, this looks good.

lib/hx509/cert.c
nicowilliams: Oh, I see now.

nicowilliams:
> In AD the weakly‐mapped certificate is valid from when the AD account for the principal was created, minus an administrator‐configurable backdating compensation offset.

That's very strange.  We can always add a field to these mappings if it turns out you need them, since they are now extensible.

nicowilliams:
The `i++` thing was wrong, but also kinda harmless since there should not be more than one SAN extension.  Which brings me to: should we support more than one SAN extension??  We could get rid of that `while` loop if we don't.

PR #1324, opened 2026-02-17
Add plugin support for hardware authentication policy

jsutton24:
This allows the plugin to indicate what is to happen when a client authenticates using a method other than hardware authentication.


nicowilliams:
This looks good.  Question: PKINIT does not imply HW auth _really_, and there is no HW indicator in certificates, so should we have the `PA_HARDWARE_AUTH` flag be more "dynamic" and have it configured per PKINIT client CA trust anchor or intermediate?

jsutton24:
That sounds reasonable. I’m not sure whether Active Directory implements this sort of configuration or not.

nicowilliams:
Thanks!

PR #1329, opened 2026-03-13
lib:roken: allow large fragmented UDP replies from glibc resolver

toreonify:
When a DNS server contains a lot of service records for Kerberos (in _kerberos._tcp for example), Heimdal client library fails to proceed with authentication as it cannot correctly receive a fragmented UDP reply. This results in an error that KDC cannot be found.

This issue is more noticeable when the reply contains additional section with resolved IP addresses for DNS names, as it almost doubles the length of the reply.

glibc resolver returns same answer length as buffer size when buffer is too small for the answer. By allowing roken to allocate a larger buffer when this happens, it can receive and parse the reply and find a KDC.

See https://github.com/krb5/krb5/commit/2cc718a86d92abef7363384d7eac48190271f676 for similar fix in dnsglue.c.

### DIFF
--- a/lib/roken/resolve.c
+++ b/lib/roken/resolve.c
@@ -568,16 +568,18 @@ dns_lookup_int(const char *domain, int rr_class, int rr_type)
 	    fprintf(stderr, "dns_lookup(%s, %d, %s) --> %d\n",
 		    domain, rr_class, rk_dns_type_to_string(rr_type), size);
 	}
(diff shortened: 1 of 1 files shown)


nicowilliams:
~~What I'm seeing in the linked MIT code is that if the packet is larger than 64KB then that's an error (as it should be).~~

nicowilliams:
Thanks!

PR #1333, opened 2026-03-24
gsskrb5: Correctly check EVP_Cipher() return value for RC4

jsutton24:
If the flag EVP_CIPH_FLAG_CUSTOM_CIPHER is set for the cipher, EVP_Cipher() returns the number of bytes written to ‘out’.


nicowilliams:
Thanks!

PR #1335, opened 2026-03-24
krb5: Fix non‐RC4 build

jsutton24:
The function _krb5_usage2arcfour() is only declared if Heimdal is built with RC4. Put calls to this function behind #ifdef guards so that the build can succeed.


nicowilliams:
Thanks!

PR #1337, opened 2026-03-24
hx509: Use correct algorithm to create/verify signatures

jsutton24:
Passing the digest in results in the wrong algorithm being used for signature creation/verification. ~~This is a regression introduced by commit cbe156d9279effd6780fc36b620321e373217863.~~

### DIFF
--- a/lib/hx509/cms.c
+++ b/lib/hx509/cms.c
@@ -1434,7 +1434,7 @@ sig_process(hx509_context context, void *ctx, hx509_cert cert)
 	ret = _hx509_create_signature(context,
 				      _hx509_cert_private_key(cert),
 				      &sigalg,
-                                      &digest,
+				      NULL,
 				      &sigdata,
 				      &signer_info->signatureAlgorithm,
 				      &signer_info->signature);


jsutton24:
The RFC is rather unclear on what is supposed to happen: https://www.rfc-editor.org/errata/eid3867

“It is unclear for what purpose the message digest algorithm is used.”

“There is no requirement that digest used on the body is the same as the digest used in the signature operation.”

“There are implementations that use the message digest algorithm specified in the messageDigest field instead of the message digest algorithm specified in the signatureAlgorithm. Is the purpose of the messageDigest field to nest the hashing algorithm used in the signing process?”

nicowilliams:
I ended up changing this in cbe156d for a reason, but I don't recall that reason at the moment.

nicowilliams:
Are you finding that this fails to interop with Active Directory?

jsutton24:
No, it just causes Samba’s testsuite to fail. I haven’t tested interoperability with Active Directory yet.

nicowilliams:
> No, it just causes Samba’s testsuite to fail. I haven’t tested interoperability with Active Directory yet.

Let us know when you have done that testing.  Maybe Samba's test suite is wrong on this.

PR #1340, opened 2026-03-25
kdc: Adjust error code mapping

jsutton24:
Commit cbe156d9279effd6780fc36b620321e373217863 mapped error code `HX509_CERT_REVOKED` to `KRB5_KDC_ERR_REVOKED_CERTIFICATE`, so check for the latter instead.


nicowilliams:
Thanks!

Issue #1342, opened 2026-03-29
Static builds fail


heitbaum:
**Describe the bug**
Builds with —-disable-shared —-enable-static fil with the following

**To Reproduce**
1. Compile with —-disable-shared —-enable-static

**Expected behavior**
Successful compile.

**Desktop (please complete the following information):**
 - OS: [e.g. iOS] Ubuntu 26.04

**Additional context**
```c

Making all in dbutils
make[2]: Entering directory '/var/media/DATA/home-rudi/LibreELEC.tv/build.LibreELEC-Generic.x86_64-13.0-devel/build/heimdal-abd35b246a7ab3b304c3ca918e52240154b87008/.x86_64-linux-gnu/appl/dbutils'
  CC       bsearch.o
  CCLD     bsearch
/usr/bin/x86_64-linux-gnu-ld.bfd: ../../lib/base/.libs/libheimbase.a(db.o): in function `open_file':
db.c:(.text+0x54d): undefined reference to `rk_flock'
collect2: error: ld returned 1 exit status
make[2]: *** [Makefile:613: bsearch] Error 1

Making all in test
make[2]: Entering directory '/var/media/DATA/home-rudi/LibreELEC.tv/build.LibreELEC-Generic.x86_64-13.0-devel/build/heimdal-abd35b246a7ab3b304c3ca918e52240154b87008/.x86_64-linux-gnu/appl/test'
  CCLD     kinit_auditdns
/usr/bin/x86_64-linux-gnu-ld.bfd: /var/media/DATA/home-rudi/LibreELEC.tv/build.LibreELEC-Generic.x86_64-13.0-devel/build/heimdal-abd35b246a7ab3b304c3ca918e52240154b87008/.x86_64-linux-gnu/lib/roken/.libs/libroken.a(libroken_la-resolve.o): in function `rk_dns_lookup':
resolve.c:(.text+0xdb0): multiple definition of `rk_dns_lookup'; kinit_auditdns-auditdns.o:auditdns.c:(.text+0x0): first defined here
collect2: error: ld returned 1 exit status

``` 

Following diff fixes the first of the errors.

```diff

--- a/appl/dbutils/Makefile.am   2026-02-11 03:14:53.000000000 +0000
+++ b/appl/dbutils/Makefile.am   2026-03-29 11:52:40.066039244 +0000
@@ -10,4 +10,4 @@
 
 EXTRA_DIST = NTMakefile $(man_MANS)
 
[...]

PR #1346, opened 2026-03-30
appl/test: fix rk_dns_lookup symbol collision in kinit_auditdns

heitbaum:
auditdns.c defines its own rk_dns_lookup to intercept DNS lookups for auditing purposes, causing a multiple definition error at link time when libroken also provides the same symbol. Newer GNU ld treats this as a hard error rather than a warning.

Use -Wl,--wrap,rk_dns_lookup and rename the local interceptor to __wrap_rk_dns_lookup so the override works correctly against both static and shared builds of libroken.

Fixes: multiple definition of `rk_dns_lookup'

- second part of the fix to #1342
- First part is #1343

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -33,6 +33,19 @@ AC_CANONICAL_HOST
 CANONICAL_HOST=$host
 AC_SUBST(CANONICAL_HOST)
 
+dnl
+dnl this is needed to not wrap calls when building on OS X Builds
+dnl
+case $host_os in
+  darwin*)
+    WRAP_LDFLAGS=""
+    ;;
+  *)
+    WRAP_LDFLAGS="-Wl,--wrap=rk_dns_lookup"
+    ;;
+esac
+AC_SUBST([WRAP_LDFLAGS])
+
 rk_SYS_LARGEFILE
 
 rk_AIX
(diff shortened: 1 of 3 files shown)


configure.ac
```diff
@@ -33,6 +33,19 @@ AC_CANONICAL_HOST
...
 AC_SUBST(CANONICAL_HOST)
 
+dnl
+dnl this is needed to not wrap calls when building on OS X Builds
+dnl
+case $host_os in
+  darwin*)
+    WRAP_LDFLAGS=""
+    ;;
+  *)
```
nicowilliams: Do we need to be more careful that `ld` really is GNU and not, say, Solaris/Illumos, or something else?

elric1: I just added a small commit that should use autoconf to see if --wrap is supported and use it if it is. Hopefully this will make more platforms build without breaking any. We shall see if the test suite works...

elric1:
One more addition, in the shared case we `-D__wrap_...=...` to ensure that shared builds are actually unaffected by this change. Hmmm, maybe I'm getting the logic here backwards...

elric1:
Okay, I made one more change. I changed the sense of our workaround so that it is only done when we are not building shared object libraries. To me this makes it a little easier to reason about because the non-static case should compile (after macro expansion) the _exact_ same lines of code with the _exact_ same compiler and linker options.

PR #1347, opened 2026-03-30
pkinit: Include serverDHNonce and dhKeyExpiration in reply when DH key reuse is requested

### LINKED ISSUE #1292: PKINIT logon fails under Windows 11 with Credential Guard
**Describe the bug**
When a Windows 11 client attempts a PKINIT (smart card) logon to a Samba Active Directory domain, that uses Heimdal, the logon fails immediately. The Windows logon screen displays the error: "A null reference pointer was passed to the stub". This occurs even though the smart card is correctly configured and works when Credential Guard is disabled.

**To Reproduce**
1. Configure Samba AD for smart card (PKINIT) logon
2. Attempt to log on to a Windows 11 client using a smart card with Credential Guard enabled

**Expected behavior**
Windows should authenticate the user successfully using the smart card.

**Desktop (please complete the following information):**
 - OS: Windows 11 24H2

**Additional context**
According to Microsoft dochelp, Windows always includes `dhKeyExpiration` and `serverDHNonce` in PKINIT AS-REP, even though it does not reuse DH keys. I have been told that an updated MS-PKCA specification will explain this behavior. These fields are required for logon to succeed when using Credential Guard.

jsutton24:
Windows clients expect these fields to be present in the reply.

Intended to fix #1292.

### DIFF
--- a/kdc/pkinit.c
+++ b/kdc/pkinit.c
@@ -62,6 +62,7 @@ struct pk_client_params {
     const heim_oid *kdf;
     unsigned char *raw_shared_secret;
     size_t raw_shared_secret_len;
+    heim_octet_string *client_dh_nonce;
 };
 
 struct pk_principal_mapping {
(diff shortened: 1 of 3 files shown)


jsutton24:
I set the DH key expiration time to two hours because that’s what Windows does, but I can change it to be shorter.

wokis:
@jsutton24 Not sure if this is ready for review yet, but I’d be happy to help test it if needed.

jsutton24:
I believe it should be ready for review?

PR #1350, opened 2026-04-07
Install heim_threads.h

elric1:


elric1:
My assistant added some nonsense to this because I'm using the branch for another purpose as well. I've created a new PR for this.

PR #1352, opened 2026-04-27
Remove duplicated libedit, find/use installed version

elric1:


dariaphoebe (commented):
I agree there's no reason to be shipping our own libedit. This patch looks reasonable

nicowilliams:
@elric1 can you make one of the GitHub Actions builds use `--with-readline=yes` and another use `--with-libeditline=yes` so we have a test of the `configure.ac` and m4 changes?

elric1:
@nicowilliams, I added the build options as requested. Tests are almost complete and the ones that I updated have already passed.

PR #1354, opened 2026-05-19
krb5: add more KRB5_TRACE coverage

elric1:


lib/krb5/get_cred.c
```diff
@@ -972,9 +1124,10 @@ get_cred_kdc_capath_worker(krb5_context context,
 	 * If we have krbtgt/server_realm@try_realm cached, use it and we're
 	 * done.
 	 */
-	ret = find_cred(context, ccache, tmp_creds.server,
+	ret = find_cred(context, ccache, tmp_creds.client, tmp_creds.server,
 			*ret_tgts, &tgts);
 	if (ret == 0) {
+	    trace_cred_pair(context, "Starting with TGT for client realm", &tgts);
```
nicowilliams: This message is misleading because `find_cred()` uses `KRB5_TC_DONT_MATCH_REALM`.

lib/krb5/get_cred.c
```diff
@@ -1208,12 +1361,13 @@ get_cred_kdc_referral(krb5_context context,
...
         }
 
-	ret = find_cred(context, ccache, tgtname, NULL, &tgt);
+	ret = find_cred(context, ccache, in_creds->client, tgtname, NULL, &tgt);
 	krb5_free_principal(context, tgtname);
 	if (ret) {
             free(start_realm);
 	    return ret;
         }
+	trace_cred_pair(context, "Starting with TGT for client realm", &tgt);
```
nicowilliams: Ditto.

nicowilliams:
BTW, we should also trace use of name canon rules.  That's partly where referrals get turned on/off.  Also, we should change the default rule to use referrals!

PR #1357, opened 2026-05-19
gss/krb5: retry stale service tickets on bad kvno

elric1:
Teach the acceptor to return KRB_AP_ERR_BADKEYVER when the AP-REQ key version is stale and a newer service key exists, and have the initiator refresh the service credential and retry.  The message also includes the current time of day just in case we also have a clock skew problem.

NOTE: like clock skew recovery, this only works with mutual auth.

### DIFF
--- a/lib/gssapi/krb5/accept_sec_context.c
+++ b/lib/gssapi/krb5/accept_sec_context.c
@@ -310,7 +310,12 @@ send_error_token(OM_uint32 *minor_status,
     /* this e_data value encodes KERB_AP_ERR_TYPE_SKEW_RECOVERY which
        tells windows to try again with the corrected timestamp. See
        [MS-KILE] 2.2.1 KERB-ERROR-DATA */
-    krb5_data e_data = { 7, rk_UNCONST("\x30\x05\xa1\x03\x02\x01\x02") };
+    krb5_data err_skew = { 7, rk_UNCONST("\x30\x05\xa1\x03\x02\x01\x02") };
+    krb5_data err_badkeyver = {
+	sizeof(GSS_KRB5_BADKEYVER_RETRY_E_DATA) - 1,
+	rk_UNCONST(GSS_KRB5_BADKEYVER_RETRY_E_DATA)
+    };
+    krb5_data *e_data = NULL;
 
     /* build server from request if the acceptor had not selected one */
     if (server == NULL) {
@@ -333,7 +338,12 @@ send_error_token(OM_uint32 *minor_status,
 	server = ap_req_server;
     }
 
-    ret = krb5_mk_error(context, kret, NULL, &e_data, NULL,
+    if (kret == KRB5KRB_AP_ERR_SKEW || kret == KRB5KRB_AP_ERR_TKT_NYV)
+        e_data = &err_skew;
+    else if (kret == KRB5KRB_AP_ERR_BADKEYVER)
+        e_data = &err_badkeyver;
+
+    ret = krb5_mk_error(context, kret, NULL, e_data, NULL,
 			server, NULL, NULL, &outbuf);
     if (ap_req_server)
 	krb5_free_principal(context, ap_req_server);
@@ -456,12 +466,13 @@ gsskrb5_acceptor_start(OM_uint32 * minor_status,
 	krb5_rd_req_in_ctx_free(context, in);
 	if (close_kt)
 	    krb5_kt_close(context, keytab);
-	if (kret == KRB5KRB_AP_ERR_SKEW || kret == KRB5KRB_AP_ERR_TKT_NYV) {
+	if (kret == KRB5KRB_AP_ERR_SKEW ||
+	    kret == KRB5KRB_AP_ERR_TKT_NYV ||
+	    kret == KRB5KRB_AP_ERR_BADKEYVER) {
 	    /*
 	     * No reply in non-MUTUAL mode, but we don't know that its
 	     * non-MUTUAL mode yet, thats inside the 8003 checksum, so
-	     * lets only send the error token on clock skew, that
-	     * limit when send error token for non-MUTUAL.
+	     * lets only send the error token for recoverable errors.
 	     */
             krb5_auth_con_free(context, ctx->auth_context);
             krb5_auth_con_free(context, ctx->deleg_auth_context);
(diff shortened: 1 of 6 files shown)


nicowilliams:
How does this work?  Do we return `GSS_S_CONTINUE_NEEDED` in this case?  That might need to be configurable behavior in case we break something.

lib/gssapi/krb5/init_sec_context.c
```diff
@@ -283,6 +283,94 @@ gsskrb5_get_creds(
...
+	krb5_timeofday(context, &ts);
+	this_cred.times.endtime = ts + time_req;
+    }
+    this_cred.session.keytype = KEYTYPE_NULL;
+
+    ret = krb5_cc_new_unique(context, krb5_cc_type_memory, NULL, &tmpcc);
+    if (ret)
+	return ret;
+
+    ret = krb5_cc_copy_cache(context, ctx->ccache, tmpcc);
```
nicowilliams: You don't need to do this.  Just remove the credential that didn't work and try again.

lib/gssapi/krb5/init_sec_context.c
```diff
@@ -693,38 +781,56 @@ init_auth_restart
...
+	p[3] = (t >> 0)  & 0xFF;
 
-	    krb5_cc_set_config(context, ctx->ccache, ctx->target,
-			       "time-offset", &timedata);
+	timedata.data = p;
+	timedata.length = sizeof(p);
 
-	    if ((ctx->more_flags & RETRIED) == 0)
-		 ctx->state = INITIATOR_RESTART;
+	krb5_cc_set_config(context, ctx->ccache, ctx->target,
```
nicowilliams: I like to write `(void) func(...);` when I want to ignore a return value -- this is a clue to others that that is not an accident but purposeful.

lib/gssapi/krb5/accept_sec_context.c
nicowilliams, line 365: Oh, TIL that we handled skew/ticket-not-yet-valid with `GSS_C_CONTINUE_NEEDED`!

lib/gssapi/krb5/init_sec_context.c
elric1: Yeah, I may be going overboard here. The idea is to only replace it if I successfully get a new one. I was thinking that because the KRB-ERROR is not authenticated, I should be a little careful.

lib/gssapi/krb5/accept_sec_context.c
```diff
@@ -456,12 +460,13 @@ gsskrb5_acceptor_start(OM_uint32 * minor_status,
 	krb5_rd_req_in_ctx_free(context, in);
 	if (close_kt)
 	    krb5_kt_close(context, keytab);
-	if (kret == KRB5KRB_AP_ERR_SKEW || kret == KRB5KRB_AP_ERR_TKT_NYV) {
+	if (kret == KRB5KRB_AP_ERR_SKEW ||
+	    kret == KRB5KRB_AP_ERR_TKT_NYV ||
+	    kret == KRB5KRB_AP_ERR_BADKEYVER) {
```
nicowilliams, line 471: Got it, this causes `GSS_S_CONTINUE_NEEDED` to be sent.  The fact that we were already doing that for `SKEW` and `TKT_NYV` makes me feel better about the chances that this breaks anything (they must be low).

nicowilliams:
I still need to review more but it looks pretty good to me so far.  I would like a config knob that defaults to this behavior being enabled because I worry that there will be server apps that send back the error token and wait for a retry that won't be forthcoming from clients whose GSS implementation doesn't support this.

elric1:
> How does this work? Do we return `GSS_S_CONTINUE_NEEDED` in this case? That might need to be configurable behavior in case we break something.

Yeah. This is only turned on if MUTUAL_AUTH is requested _and_ interestingly it was specified by MSFT for clock skew _and_ we already do it for clock skew. I'm just adding the mismatched kvno case with the same logic as already exists.
I was thinking of doing it for all cases where we fail to find a keytab entry. We already test every key in the keytab just in case... it might make sense to try to get another key if that fails, but it might be too broad. For my first patch, though, I decided to be a little conservative and only retry if there is a key in the kt with the correct name and a higher kvno.

lib/gssapi/krb5/init_sec_context.c
nicowilliams: Just remove the old credential, or... if you're scared of the `KRB-ERROR` being unauthenticated then first try to acquire a new one (with the flag that skips looking for it in the ccache) and IFF you get one with a diff kvno _then_ remote the old one and store the new one -- this way you don't disrupt other clients/initiators wanting the same cred if the `KRB-ERROR` were spoofed.

nicowilliams: But IMO KISS and just remove the old cred and retry.

nicowilliams:
So let's just simplify the credential refresh code and done.

lib/gssapi/krb5/accept_sec_context.c
elric1: Also, keep in mind that this only occurs in quite rare cases. Misconfigured clocks and deleted old kvnos. And they are failure cases already, so the risk should be relatively low.

lib/gssapi/krb5/gsskrb5_locl.h
```diff
@@ -52,6 +52,15 @@
  *
  */
 
+/*
+ * DER SEQUENCE { OBJECT IDENTIFIER 1.2.752.43.13.32 }
```
nicowilliams, line 56: What is this OID's symbolic arc name form?

nicowilliams: Ah, `{ iso(1) member-body(2) se(752) su(43) ...}` and `1.2.752.43.13 Namn Heimdal GSS-API extensions`, but we should really have an ASN.1 module with all the Heimdal, PADL, and Cryptonector allocations used in-tree.  I could swear I've a branch where I added that.

nicowilliams: Anyways, for now `lib/gssapi/oid.txt` is sort of the registry we have, so please update that file so we don't miss the allocation of `.32`.

nicowilliams:
It's ready except for editing `lib/gssapi/oid.txt` to record the new OID allocation.  Feel free to rebase and merge when you've made that change.

PR #1370, opened 2026-05-27
kadmin: generate stronger random passwords

### LINKED ISSUE #1145: kadmin(8) random passwords are very short and low-entropy
`kadmin --random-password` chooses 10-character passwords with a procedure that has a little under 56 bits of entropy.

This may have been appropriate for the days when the 56-bit key space of DES was considered sufficient (which would have been [no later than 1979](https://ieeexplore.ieee.org/document/6368157)), but perhaps it is time to reconsider that and provide 128 bits of entropy.

My usual approaches are either:

1. 20 independent uniform random choices of the 95 graphic US-ASCII characters
2. 10 independent uniform random choices from a 7776-word diceware dictionary

Of course, I can always just use `tr -cd '[:graph:]' </dev/urandom | head -c 20; echo` or https://mumble.net/~campbell/js/pwgen to generate a password myself instead, and that's what I do, but perhaps the built-in functionality of kadmin(8) should be closer to modern standards of cryptographic security.

elric1:
Generate 12 random bytes and encode them with Heimdal base64url, producing 16-character random passwords with 96 bits of entropy.

This avoids punctuation and quoting issues while still raising generated password entropy substantially from the old roughly 55-bit generator.

Fixes #1145

### DIFF
--- a/kadmin/Makefile.am
+++ b/kadmin/Makefile.am
@@ -54,7 +54,7 @@ kadmind_SOURCES =				\
 
 add_random_users_SOURCES = add-random-users.c
 
-test_util_SOURCES = test_util.c util.c
+test_util_SOURCES = test_util.c util.c random_password.c
 
 TESTS = test_util
 
--- a/kadmin/random_password.c
+++ b/kadmin/random_password.c
@@ -32,131 +32,30 @@
  */
 
 #include "kadmin_locl.h"
+#include <base64.h>
 
-/* This file defines some a function that generates a random password,
-   that can be used when creating a large amount of principals (such
-   as for a batch of students). Since this is a political matter, you
-   should think about how secure generated passwords has to be.
-
-   Both methods defined here will give you at least 55 bits of
-   entropy.
-   */
-
-/* If you want OTP-style passwords, define OTP_STYLE */
-
-#ifdef OTP_STYLE
-#include <otp.h>
-#else
-static void generate_password(char **pw, int num_classes, ...);
-#endif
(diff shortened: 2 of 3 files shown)


nicowilliams:
Does this bypass dumb password quality checks?

elric1:
Do we need to? We are setting the passwd in our db, so we should control the horizontal and the vertical of this.

nicowilliams:
I'm asking if the generated password can fail password policies between generation and finishing setting it.

nicowilliams:
Hmm, yeah, so `random_password()` is called from `kadmin(1)` commands, which will call `kadm5_chpass_principal()`/`kadm5_chpass_principal_3()`/`kadm5_create_principal()`, which will call the `kadm5_s_*` variants directly (`kadmin -l`) or indirectly (via `kadmind`) and then there... ahhh:

```
    if (krb5_principal_compare(context->context, princ, context->caller) ||
        _kadm5_enforce_pwqual_on_admin_set_p(context)) {
        krb5_data pwd_data;
        const char *pwd_reason;

        pwd_data.data = rk_UNCONST(password);
        pwd_data.length = strlen(password);

        pwd_reason = kadm5_check_password_quality(context->context,
                                                  princ, &pwd_data);
        if (pwd_reason != NULL) {
            krb5_set_error_message(context->context, KADM5_PASS_Q_GENERIC, "%s", pwd_reason);
            return KADM5_PASS_Q_GENERIC;
        }
    }
```

So, if it's not the same user (it won't be) then the quality controls are bypassed.

Ok, LGTM!

PR #1376, opened 2026-05-27
gssapi: avoid putenv use-after-free in store_cred

### LINKED ISSUE #1163: _gsskrb5_store_cred_into2 abuses putenv, likely leading to use-after-free
**Describe the bug**
_gsskrb5_store_cred_into2 sometimes allocates an environment variable assignment with malloc:
https://github.com/heimdal/heimdal/blob/1b954faeebb833c5e07c192a76fbff5f3dadeb4e/lib/gssapi/krb5/store_cred.c#L335
https://github.com/heimdal/heimdal/blob/1b954faeebb833c5e07c192a76fbff5f3dadeb4e/lib/gssapi/krb5/store_cred.c#L71
https://github.com/heimdal/heimdal/blob/1b954faeebb833c5e07c192a76fbff5f3dadeb4e/lib/gssapi/mech/gss_buffer_set.c#L85

Then it passes this assignment to putenv:

https://github.com/heimdal/heimdal/blob/1b954faeebb833c5e07c192a76fbff5f3dadeb4e/lib/gssapi/krb5/store_cred.c#L345
https://github.com/heimdal/heimdal/blob/1b954faeebb833c5e07c192a76fbff5f3dadeb4e/lib/gssapi/krb5/store_cred.c#L87

Then it frees the assignment:

https://github.com/heimdal/heimdal/blob/1b954faeebb833c5e07c192a76fbff5f3dadeb4e/lib/gssapi/krb5/store_cred.c#L347
https://github.com/heimdal/heimdal/blob/1b954faeebb833c5e07c192a76fbff5f3dadeb4e/lib/gssapi/mech/gss_buffer_set.c#L112
https://github.com/heimdal/heimdal/blob/1b954faeebb833c5e07c192a76fbff5f3dadeb4e/lib/gssapi/mech/gss_release_buffer.c#L38

At this point, there is still a pointer to the freed environment variable assignment string in environ.

Subsequent calls by the same process to exec or posix_spawn will pass this pointer along.  If the memory has been recycled, this may leak secrets into other processes.

**To Reproduce**
code inspection

**Expected behavior**
no putenv abuse

elric1:
Fixes #1163

### DIFF
--- a/lib/gssapi/krb5/store_cred.c
+++ b/lib/gssapi/krb5/store_cred.c
@@ -83,8 +83,25 @@ set_proc(OM_uint32 *minor, gss_buffer_set_t env)
 #ifndef WIN32
     size_t i;
 
-    for (i = 0; i < env->count; i++)
-        putenv(env->elements[i].value);
+    for (i = 0; i < env->count; i++) {
+        char *var = (char *)env->elements[i].value;
+        char *eq  = strchr(var, '=');
+        char *val;
+        int e;
+
+        if (eq == NULL) {
+            *minor = EINVAL;
+            return GSS_S_FAILURE;
+        }
+        val = eq + 1;
+        *eq = '\0';
+        e = setenv(var, val, 1);
+        *eq = '=';
+        if (e != 0) {
+            *minor = errno ? errno : EINVAL;
+            return GSS_S_FAILURE;
+        }
+    }
 #endif
     return GSS_S_COMPLETE;
 }


lib/gssapi/krb5/store_cred.c
```diff
@@ -83,8 +83,25 @@ set_proc(OM_uint32 *minor, gss_buffer_set_t env)
...
+        char *eq  = strchr(var, '=');
+        char *val;
+        int e;
+
+        if (eq == NULL) {
+            *minor = EINVAL;
+            return GSS_S_FAILURE;
+        }
+        val = eq + 1;
+        *eq = '\0';
```
nicowilliams, line 97: `putenv()` is much easier to call than `setenv()` here.  What do we gain when we could instead either `strdup()` or not `free()` these strings?  But anyways, LGTM.

PR #1379, opened 2026-05-28
kdc: reject kx509 requests when disabled

### LINKED ISSUE #1378: kx509 isn't disabled when enable_kx509 is false

elric1:
Fixes #1378.


nicowilliams:
~~I agree that this is a good change, but I do believe that `kdc/connect.c` will not plumb the service if it's disabled, so perhaps we shouldn't even check if kx509 is enabled in `kdc/kx509.c`.~~

PR #1381, opened 2026-05-30
roken: Don't define any global vars in the pidfile namespace.

rsmarples:
Newer util.h from NetBSD defines pidfile_path() which conflicts here.


rsmarples:
https://mail-index.netbsd.org/source-changes/2026/05/30/msg162215.html

nicowilliams:
Thanks!

PR #1382, opened 2026-06-05
krb5: avoid repeated HMAC digest setup

elric1:


lib/krb5/crypto-evp.c
```diff
@@ -197,12 +198,16 @@ _krb5_evp_hmac_iov(krb5_context context,
         goto out;
     }
 
+    if (crypto == NULL || crypto->hmacctx_md != md)
```
nicowilliams: I don't understand this.  This is the only place where you're using `crypto->hmacctx_md`, and only to decide whether to pass in `NULL` for the params to `EVP_MAC_init()`.  It seems quite incorrect to ever pass in `NULL` for params to `EVP_MAC_init()`.

lib/krb5/crypto.h
```diff
@@ -232,6 +232,7 @@ struct krb5_crypto_data {
     EVP_MAC_CTX *hmacctx;
     EVP_MD_CTX *mdctx;
     EVP_MAC *mac;
+    const EVP_MD *hmacctx_md;
```
nicowilliams: I just don't think we need this.  We already memoize `EVP_MD` values elsewhere.

lib/krb5/crypto-evp.c
elric1: I think that you can pass in a NULL in which case, it will reset the MAC to the beginning but use the same params as last time. It's slightly better than just memoising it, because you avoid the expense of setting up the params each time.

elric1: There may be a more obvious way to do this optimisation.

lib/krb5/crypto-evp.c
```diff
@@ -181,24 +181,37 @@ _krb5_evp_hmac_iov(krb5_context context,
...
-	ctx = crypto->hmacctx;
         mac = crypto->mac;
-    } else {
-        mac = EVP_MAC_fetch(context->ossl->libctx, "HMAC", context->ossl->propq);
-	ctx = EVP_MAC_CTX_new(mac);
+        ctx = crypto->hmacctx;
+    }
+
+    if (mac == NULL) {
+        mac = EVP_MAC_fetch(context->ossl->libctx, "HMAC",
```
nicowilliams, line 189: We already fetch this into `context->ossl->hmac`!

elric1: So, we are caching it in multiple places... I'm just streamlining the existing cache in krb5_crypto.

nicowilliams: We should use `context->ossl->hmac` -- that's the intended cache.  If we cache it in `krb5_crypto` it has to be because we intend to reuse the key schedule because we intend to reuse the key, and even then we should `EVP_MAC_CTX_dup()` the one from `context->ossl->hmac` so we only do the fetch of `"HMAC"` once.

PR #1383, opened 2026-06-06
asn1: Include <errno.h> before using/checking bits from it

pinotree:
Commit 232c936ea3eabfd702aaaa106ddf9474f231d776 added a fallback `ENOTSUP` definition. The check itself and the fallback definition rely on `<errno.h>` to be included already, so in case it was not (depending on the platform) then there will be a wrong definition of `ENOTSUP`.

Instead of rely on `<errno.h>` to be included before including a generated header from this tool, include it manually. This way both the `ENOTSUP` check and its fallback definition (if ever) will always work.


nicowilliams:
Thanks!

PR #1388, opened 2026-06-09
add i386 debian to exercise 32bits

elric1:


elric1:
I'm breaking this into smaller pull requests.

PR #1389, opened 2026-06-10
kadm5: fix use-after-free in principal creation

elric1:


lib/kadm5/create_s.c
```diff
@@ -346,9 +346,8 @@ kadm5_s_create_principal(void *server_handle,
     if (ret)
         goto out;
 
-    free_Keys(&ent.keys);
```
nicowilliams, line 349: How was this a use-after-free?  The ASN.1 `free_<Type>()` functions zeroize the pointed-to value.

nicowilliams: Though I agree with this change -- it's just not a use-after-free.

nicowilliams: Watch out though!  Before this change we'd be creating principals with _no_ keys unless creating them with a password.  But if we look at `kadmin/ank.c` I think this is fine.

nicowilliams:
Apart from my comment about this not being a use-after-free, LGTM.

nicowilliams:
Since this is a not a use-after-free, what is this bug?

lib/kadm5/create_s.c
elric1: I think that this isn't actually a bug and let's table this until after the release.

PR #1391, opened 2026-06-10
asn1: clamp far-future times on 32-bit time_t

elric1:
Compute DER generalized time conversion in uint64_t and clamp to the local time_t range before returning.  This avoids wrapping far-future ASN.1 times on 32-bit time_t platforms.  We do not address time_t being a non-integral type.


lib/asn1/timegm.c
```diff
@@ -46,6 +46,21 @@ static const unsigned ndays[2][12] ={
     {31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31},
     {31, 29, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31}};
 
+static time_t
+_clamp_time_t(uint64_t t)
+{
+#if SIZEOF_TIME_T < 8
+#ifdef TIME_T_UNSIGNED
+#define MAX_TIME ((UINT64_C(1) << (SIZEOF_TIME_T * 8)) - 1)
```
nicowilliams, line 54: If sizeof time_t is less than 8 then it's 4, so just use `INT32_MAX`/`UINT32_MAX` as appropriate, no?

nicowilliams: You would need to subtract one from the shift count, otherwise you'll be using too-big a max value!

lib/asn1/timegm.c
```diff
@@ -55,7 +70,7 @@ static const unsigned ndays[2][12] ={
 time_t
 _der_timegm (struct tm *tm)
```
nicowilliams, line 71: We might want a change to `_der_gmtime()` as well, to return `NULL` when the given `time_t t` value is negative on platforms where `time_t` is signed 32-bit -- times before 1970-01-01T00:00:00Z are symptomatic of overflow or other errors.

elric1: For our use case, this does sound compelling as we will never have certificates which expire before certificates were invented, however, I worry that this is a generic ASN.1 compiler and library and people might eventually use it for something where expressing times in the distant past is perfectly valid.

lib/asn1/timegm.c
elric1: I'm adding a test that ensures that the clamping is to the exact value that we expect.

lib/asn1/timegm.c
nicowilliams: So maybe we should have a `gmtime()`/`timegm()` implementation in `lib/roken/` for use on 32-bit platforms with signed 32-bit `time_t`?  That way `lib/asn1/` can define its own type for time that isn't necessarily `time_t`.  That would force us to do type conversions in `lib/hx509/` and `lib/krb5/` and `kdc/`...

elric1: Yeah, but maybe not for Heimdal 8? I think that the best path forward is, right after the release we should break asn1_compile out into a separate project and maybe move this clamping logic back into libroken at that point? Or should we do it now?

nicowilliams: Let's not do any of this now, yeah.

PR #1395, opened 2026-06-11
Test ilp32

elric1:


cf/largefile.m4
```diff
@@ -19,4 +19,10 @@ if test "$enable_largefile" != no; then
 			[CPPFLAGS="$CPPFLAGS $ac_cv_sys_largefile_opts"])
 	fi
 fi
+if test "$enable_year2038" != no; then
```
nicowilliams, line 22: Don't we need to invoke `AC_ARG_ENABLE()` to get this set??

nicowilliams: ~Also, maybe this belongs in `configure.ac`?  The connection to "large files" is real but tenuous and mainly historical.~

EDIT: Ah, this `--enable-year2038` thing comes from invoking `AC_SYS_LARGEFILE`, apparently.

nicowilliams: Ahhh, this is new and default in autoconf 2.72 and up.  TIL!

nicowilliams: Should we add `AC_SYS_YEAR2038`?

nicowilliams: Or even `AC_SYS_YEAR2038_RECOMMENDED`?

elric1: I think that `AC_SYS_YEAR2038` does this. We _could_ set `AC_SYS_YEAR2038_RECOMMENDED` which would make 64 bit `time_t` the default and disable it with `--disable-year2038`.

lib/hx509/cert.c
```diff
@@ -78,6 +78,7 @@ struct hx509_cert_data {
     unsigned int ref;
     char *friendlyname;
     Certificate *data;
+    heim_octet_string encoding;
```
nicowilliams, line 81: We don't need this.  `TBSCertificate` already has a `_save` field, and if need be we can `--preserve=Certificate` too if you want to preserve its encoding w/o having to make this sort of change here.

nicowilliams: If you make the `--preserve=Certificate` change then all of the changes in this file become unnecessary except a change to `hx509_cert_binary()` to make it output a copy of `c->data->_save`!

nicowilliams: Never mind, we expose `Certificate` in the ABI, as you noted.

PR #1397, opened 2026-06-18
lib/asn1: fix USE_AFTER_FREE in der_print_hex_heim_integer

mikhailofff:
Set `*p` to `NULL` on `asprintf` failure to prevent a dangling pointer and subsequent `USE_AFTER_FREE` when the caller ignores `ENOMEM`.


nicowilliams:
Thanks!

PR #1399, opened 2026-06-19
kdc: Fix possible double free of ac pointer

Dan2206:
This fixes possible double free of ac pointer. In the function 'krb5_auth_con_free' pointer ac (auth_context) is freed. In case 'ac != NULL', 'ac != tgs_ac', 'ac->remote_subkey == NULL' this function is called twice (when logging "FAST AP-REQ remote subkey missing" and at label 'out').

Signed-off-by: Daniil Sarafannikov <sarafannikovda@sgu.ru>

### DIFF
--- a/kdc/fast.c
+++ b/kdc/fast.c
@@ -622,6 +622,7 @@ fast_unwrap_request(astgs_request_t r,
 
     if (ac->remote_subkey == NULL) {
 	krb5_auth_con_free(r->context, ac);
+	ac = NULL;
 	kdc_log(r->context, r->config, 2,
 		"FAST AP-REQ remote subkey missing");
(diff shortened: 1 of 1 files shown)


elric1:
I think that a better fix would be to not free it in the conditional. We are going to jump to out in any case which will free it. And any caller should free it on failure. This would be more idiomatic, I think.

nicowilliams:
Thanks!

nicowilliams:
> I think that a better fix would be to not free it in the conditional. We are going to jump to out in any case which will free it. And any caller should free it on failure. This would be more idiomatic, I think.

Sorry I didn't see this.  See #1400.

PR #1401, opened 2026-06-21
gss-token: add local user authorization check

elric1:


nicowilliams:
`gsstool(1)`  should get this functionality too, and userok probably, then `gss-token(1)` wouldn't strictly need it, and it would be easier to write tests for `gss_localname()` and `gss_userok()`.

elric1:
Yeah, that might be a better idea. Also: we've never released gss-token, it _might_ be an idea to turn it into a verb in `gsstool` as well, i.e.
```
:; gsstool token ...
```

nicowilliams:
> `:; gsstool token ...`

That's fine for the `gss-token` functionality, but we could also have init/accept w/ a file in which to put the exported security context.  And we could have a way to iterate available credentials, maybe, and have `gsstool` be a client for your GSS agent?

nicowilliams:
To be clear: I'm not opposed to this change.

nicowilliams:
LGTM

elric1:
I'm still thinking about this. It might be better to have a slightly better UI and because we are about to release Heimdal 8 and `gss-token` has never been part of Heimdal, this is the time to make sure that we get it right. Honestly, I'd prefer to call `gsstool` `gss`...

nicowilliams:
So let's rename `gsstool` to `gss`!

PR #1402, opened 2026-06-21
gss: Fix NULL pointer dereference in _gss_copy_buffer()

shumikhinaka:
If minor_status is NULL, the if (minor_status) check is skipped and the pointer remains NULL. It is then passed to _gss_copy_buffer, where it is dereferenced without being checked.

A pointer check has been added to the _gss_copy_buffer function before each dereferencing operation.

Signed-off-by: Shumikhina Ksenia <shumikhinaka@sgu.ru>

### DIFF
--- a/lib/gssapi/mech/gss_utils.c
+++ b/lib/gssapi/mech/gss_utils.c
@@ -135,10 +135,14 @@ _gss_copy_buffer(OM_uint32 *minor_status,
 {
 	size_t len = from_buf->length;
 
+    if (minor_status) {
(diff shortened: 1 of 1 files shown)


nicowilliams:
It is always an error to call a GSS-API function with a `NULL` `minor_status` value.  Crashing is better than returning an error in that case.  And if we were going to make it not a crash, there would be many other functions to which to add checks like this.  I'm strongly inclined to reject this, but I appreciate the contribution, and I want you not to be discouraged.

PR #1404, opened 2026-06-25
roken:strsep_copy: Fix out-of-bounds write in strsep_copy

Dan2206:
In case len < (size_t)(*stringp - save) and len > 0
access to buf[l] leads to out-of-bounds write as
l = len and len means the length of buffer.

Change evaluation of l: only len - 1 elements of
buffer can be used for memcpy and terminator
should be written to buf[len - 1] (in case l = len - 1). 
Also, this value of l is not used when len == 0,
so we calculate it only when len > 0.

Signed-off-by: Daniil Sarafannikov <sarafannikovda@sgu.ru>


elric1:
Thanks!

PR #1406, opened 2026-06-29
lib:hx509: fix NULL dereferences in hx509_Name_to_string

shishkovna:
If NULL value passed as 4th parameter in call wind_ucs2utf8 or wind_ucs4utf8, it will be dereferenced inside of conditions.

The length used to allocate space via malloc can be used as the 4th parameter for the call.

Pair-Programmed-With: Dmitriy Mikhalchenko <tascad@altlinux.org>
Signed-off-by: Shishkov Nikita <shishkovna@sgu.ru>


nicowilliams:
image

nicowilliams:
Thanks!

PR #1408, opened 2026-07-22
Systemd integration with start/stop notifs and socket activation

h7x4:
This PR implements better integration with systemd in two ways:

First, we make use of [`sd_notify`](https://man7.org/linux/man-pages/man3/sd_notify.3.html) to give systemd more accurate information about when the service has fully started and when it is about to stop. In addition we post `STATUS` notifs, providing a quick insight into how the service is doing when a sysadmin runs `systemctl status <service>`.

The second, and maybe more interesting change, is socket activation via fd passing. This might enable quicker startup on systems with services that depend on either of the services (`kdc` in particular maybe), because they now only need to wait for the socket to start. If a dependent service connects to the socket before kdc/kpasswd/kadmind is running, systemd will hold the connection until the service has started. It also enables the use of the [`PrivateNetwork` directive](https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#PrivateNetwork=), sandboxing all services into private network namespaces.

Testing this through the existing test framework seemed quite complicated, so I unfortunately have not figured out how to write any. I have a QEMU-based integration test for NixOS specifically, where I have verified all of `Type=notify`, socket activation and `PrivateNetwork` works as expected [here](https://github.com/h7x4/heimdal/blob/sdnotify-nix-testing/nix/nixosTest.nix) ([NixOS module](https://github.com/h7x4/heimdal/tree/sdnotify-nix-testing/nix/module)).

I'd be happy to provide documentation about how to use this functionality, as well as write suggested systemd units, but I'm a bit unsure about where to add them. Would it be best to add them alongside the source/manpages and install them with `make`, write docs in the `doc` directory, write docs in the github wiki, or somewhere else? Do you have something akin to a `contrib` directory somewhere where I could put example units?

### DIFF
--- a/kadmin/kadm_conn.c
+++ b/kadmin/kadm_conn.c
@@ -247,68 +247,93 @@ start_server(krb5_context contextp, const char *port_str)
     krb5_socket_t *socks = NULL, *tmp;
     unsigned int num_socks = 0;
     int i;
+    char **names = NULL;
+    int sd_n, j;
 
-    if (port_str == NULL)
-	port_str = "+";
+    sd_n = rk_sd_listen_fds_with_names(0, &names);
+    if (sd_n < 0)
+        krb5_err(contextp, 1, -sd_n, "rk_sd_listen_fds_with_names");
+    if (sd_n > 0) {
+        socks = malloc(sd_n * sizeof(*socks));
+        if (socks == NULL)
+            krb5_err(contextp, 1, errno, "malloc");
+        for (j = 0; j < sd_n; j++) {
+            if (names != NULL && names[j] != NULL &&
+                strcmp(names[j], "kadmind") == 0) {
+                krb5_socket_t s = SD_LISTEN_FDS_START + j;
 
-    parse_ports(contextp, port_str);
+                socket_set_keepalive(s, 1);
+                socks[num_socks++] = s;
+            }
+        }
+    }
+    for (j = 0; names != NULL && names[j] != NULL; j++)
+        free(names[j]);
+    free(names);
 
-    for(p = kadm_ports; p; p = p->next) {
-	struct addrinfo hints, *ai, *ap;
-	char portstr[32];
-	memset (&hints, 0, sizeof(hints));
-	hints.ai_flags    = AI_PASSIVE;
-	hints.ai_socktype = SOCK_STREAM;
+    if (num_socks == 0) {
+	if (port_str == NULL)
+	    port_str = "+";
 
-	if (krb5_config_get_bool(context, NULL, "libdefaults", "block_dns",
-		NULL)) {
-	    hints.ai_flags &= ~AI_CANONNAME;
-	    hints.ai_flags |= AI_NUMERICHOST|AI_NUMERICSERV;
-	}
-	e = getaddrinfo(NULL, p->port, &hints, &ai);
-	if(e) {
-	    snprintf(portstr, sizeof(portstr), "%u", p->def_port);
-	    e = getaddrinfo(NULL, portstr, &hints, &ai);
-	}
+	parse_ports(contextp, port_str);
 
-	if(e) {
-	    krb5_warn(contextp, krb5_eai_to_heim_errno(e, errno),
-		      "%s", portstr);
-	    continue;
-	}
-	i = 0;
-	for(ap = ai; ap; ap = ap->ai_next)
-	    i++;
-	tmp = realloc(socks, (num_socks + i) * sizeof(*socks));
-	if(tmp == NULL)
-	    krb5_err(contextp, 1, errno, "failed to reallocate %lu bytes",
-		     (unsigned long)(num_socks + i) * sizeof(*socks));
-	socks = tmp;
-	for(ap = ai; ap; ap = ap->ai_next) {
-	    krb5_socket_t s = socket(ap->ai_family, ap->ai_socktype, ap->ai_protocol);
-	    if(rk_IS_BAD_SOCKET(s)) {
-		krb5_warn(contextp, rk_SOCK_ERRNO, "socket");
-		continue;
+	for(p = kadm_ports; p; p = p->next) {
+	    struct addrinfo hints, *ai, *ap;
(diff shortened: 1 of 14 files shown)


kdc/connect.c
```diff
@@ -299,6 +299,69 @@ init_socket(krb5_context context,
...
+	    continue;
+	if (getsockopt(fd, SOL_SOCKET, SO_TYPE, &type, &tlen) < 0) {
+	    krb5_warn(context, errno,
+		      "getsockopt(SO_TYPE) on socket-activated fd %d", fd);
+	    continue;
+	}
+	init_descr(&d[num]);
+	d[num].s = fd;
+	d[num].type = type;
+	socket_set_nonblocking(fd, 1);
```
elric1: You also need:
```
rk_cloexec(d->s);
socket_set_keepalive(d->s, 1);
```
It probably would make sense to split `init_socket()` into a top and bottom where the bottom is everything that will be done to an arbitrary listening socket, that is everything that _can_ be deferred until after the `bind(2)` and then share that between both use cases.

kadmin/kadm_conn.c
```diff
@@ -35,6 +35,11 @@
 #ifdef HAVE_SYS_WAIT_H
 #include <sys/wait.h>
 #endif
+#ifdef HAVE_SYSTEMD
+#include <systemd/sd-daemon.h>
```
nicowilliams: Afteer xz/OpenSSH disaster, can we not implement `sd_notify()` here instead of importing a library from systemd?

nicowilliams: My preference would be that we implement a `rk_sd_notify()` in `lib/roken/` and use that here.

nicowilliams:
FYI, all the Heimdal services daemonize properly in that when you request that they daemonize then the process you started doesn't exit until it's child/grandchild is ready.  I wish that were the standard way to do readiness signalling in systemd units!

nicowilliams:
Because Heimdal services daemonize properly (meaning the process you spawn does not exit until its progeny are a. running, b. ready), we should just not need this `sd_notify()` stuff -- just add `Type=forking` to your systemd units.

I.e., we should not need this PR, except for:

- `WATCHDOG=...`, signaling continued readiness (i.e., the service isn't stuck) -- but see below
- `RELOAD=...`, signaling [config] reload completion, but our services do not support reloading config configuration!)

But I'm not opposed to `lib/roken` gaining an `rk_sd_notify()` that does the trivial thing without linking with systemd.  After the OpenSSH/xz disaster we should be wary of importing all of that systemd library just for a trivial `sd_notify()`.

I am opposed to using `sd_notify()` to signal continued readiness, as opposed to initial readiness unless we add an option to disable that.  Periodic signaling prevents deep C-states and costs $$.

So right now I'm inclined to:

- if need be, better-document Heimdal daemons' daemonization behavior to clarify that `Type=forking` works for systemd integration
- accept a PR that implements `sd_notify()` w/o linking systemd libraries
- accept a PR that implements a command-line option to Heimdal services to enable (or disable) `WATCHDOG=...` signaling

nicowilliams:
Here's a chat with Claude about this: https://claude.ai/share/72ee0b89-8fa8-4d7f-960b-6663cad046da

kdc/connect.c
h7x4: Just making sure I understand you correctly, you're suggesting that I inline most of the contents of `init_sockets_sd` into `init_socket` and then move deferred parts to the bottom? Or do you want me to create a sort of `finalize_socket` postlude to call at the bottom of both functions?

h7x4:
I think the larger part of this PR is the socket activation part, which also uses various pieces of libsystemd. It's not terribly complicated IIRC, I'd be happy to provide an implementation of this as well, but I thought I would clarify that it's not *just* `sd_notify`.

The PR does not currently add any support for the watchdog mechanism. There is a `STATUS=` update put into the loop, but if we only ever update the status message when there a connection is opened or closed it shouldn't cause any unnecessary wakeups (I'll change the current implementation to remove the existing timer).

I imagine watchdog support could be added in combination with the socket activation mechanism by shutting down the services when they seem not to be in active use, and let systemd wake them up again whenever they are needed again. If the daemon is frequently enough used that it doesn't shut itself down, then it can afford to continuously send watchdog pings. But this should be opt in as well (you probably don't want this if you instance is frequently used), and I considered it out of the scope of the PR - especially with regards to the "collect heartbeats from other threads" correctness part, which was also mentioned in the claude thread.

If `sd_notify` and socket activation is added to `libroken`, would you still like to gate it behind a compile time flag or is it okay if it unconditionally tries to detect if there is a notify socket and/or fd-passing sockets available?

nicowilliams:
> If `sd_notify` and socket activation is added to `libroken`, would you still like to gate it behind a compile time flag or is it okay if it unconditionally tries to detect if there is a notify socket and/or fd-passing sockets available?

Unconditionally using it seems fine.  Only `WATCHDOG=` message need to be conditional, and... right, you didn't include any.

h7x4:
Added libroken implementations of the systemd functions, enabled keepalive for kdc and kadmind, changed the STATUS messaging to only happen whenever the loops are active for other reasons.

nicowilliams:
```  CC       libroken_la-rk_sd_listen_fds_with_names.lo
  CC       libroken_la-rk_sd_notify.lo
/tmp/heimdal-src/lib/roken/rk_sd_notify.c: In function 'rk_sd_notify':
/tmp/heimdal-src/lib/roken/rk_sd_notify.c:22:9: error: variable 'ret' set but not used [-Werror=unused-but-set-variable]
   22 |     int ret = 0;
      |         ^~~
/tmp/heimdal-src/lib/roken/rk_sd_notify.c:77:1: error: control reaches end of non-void function [-Werror=return-type]
   77 | }
      | ^
```

h7x4:
Added the functions to `NTMakefile` as well, should hopefully fix the CI failure. Also noticed that `unsetenv` was not included in that list, so I added it in a separate commit.

PR #1410, opened 2026-07-28
MacOS X upgrade requires small changes to pass CI

elric1:
Our macosx image was upgraded and uses a new version of clang which cause our CI to fail.  Here are three small changes which allow the build to proceed.

### DIFF
--- a/lib/gssapi/test_context.c
+++ b/lib/gssapi/test_context.c
@@ -653,10 +653,11 @@ wrapunwrap_iov(gss_ctx_id_t cctx, gss_ctx_id_t sctx, int flags, gss_OID mechoid)
     gss_iov_buffer_desc iov[6];
     unsigned char *p;
     int iov_len;
-    char header_data[9] = "ABCheader";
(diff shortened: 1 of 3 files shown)


lib/gssapi/test_context.c
```diff
@@ -653,10 +653,11 @@ wrapunwrap_iov(gss_ctx_id_t cctx, gss_ctx_id_t sctx, int flags, gss_OID mechoid)
     gss_iov_buffer_desc iov[6];
     unsigned char *p;
     int iov_len;
-    char header_data[9] = "ABCheader";
-    char trailer_data[10] = "trailerXYZ";
+    char header_data[9] = { 'A', 'B', 'C', 'h', 'e', 'a', 'd', 'e', 'r' };
```
nicowilliams, line 656: I would really prefer either `s/9/10/` or `s/9//` as the fix.

nicowilliams: Accepted as-is, but in the future I'd rather not resort to this.  It's ugly and not customary here.

PR #1411, opened 2026-07-29
Issue 1252

### LINKED ISSUE #1252: includedir should read files in predictable order
**Describe the bug**
The implementation of the krb5.conf `includedir` directive currently processes included files in `readdir()` order, which is not determined. MIT kerberos first implementation of `includdir` was like that, but later changed to an alphanumeric order.[1]

1. https://github.com/krb5/krb5/commit/638a161713ff915ff692d91f5249371f7986ddc4

It would probably be best to mimic MIT's behavior here, given /etc/krb5.conf is usually shared between heimdal and MIT, and it also makes sense to have a predictable order.

**To Reproduce**
- create /etc/krb5.conf.d
- create /etc/krb5.conf.d/1config.conf with content:
```
[libdefaults]
default_realm = ONE
```
- create /etc/krb5.conf.d/0config.conf with content:
```
[libdefaults]
default_realm = ZERO
```
- add to the top of /etc/krb5.conf if not present already: `includedir /etc/krb5.conf.d`
- run kinit

**Expected behavior**
kinit should ask for a password for the user in the ZERO realm, as that's defined in 0config.conf, but instead it will ask for the ONE realm which is in 1config.conf which was created first. If these files are created in the reverse order, the default realm will be the other one.

**Desktop (please complete the following information):**
 - OS: Ubuntu/Debian with heimdal patched to support include/includedir from heimdal master (straight cherry-picks)

elric1:
Fixes #1252

### DIFF
--- a/lib/roken/dirent.c
+++ b/lib/roken/dirent.c
@@ -31,6 +31,7 @@
 
 #include<config.h>
 
+#include <limits.h>
 #include <stdlib.h>
 #include <io.h>
 #include <string.h>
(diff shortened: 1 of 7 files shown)


lib/roken/dirent.c
```diff
@@ -214,6 +215,95 @@ readdir(DIR * dp)
...
+
+        len = strlen(entry->d_name);
+        if ((copy = malloc(sizeof(*copy) + len)) == NULL) {
+            errno = ENOMEM;
+            goto fail;
+        }
+        copy->d_ino = entry->d_ino;
+        memcpy(copy->d_name, entry->d_name, len + 1);
+
+        tmp = realloc(list, (n + 1) * sizeof(list[0]));
```
nicowilliams: Watch out!  O(N^2) behavior here.

nicowilliams:
I could not force-push to your branch, so I pushed a variant with a slight fix to `issue-1252` on my clone.  It basically grows the scandir output array by ~.5x each time it needs to, thus avoiding an O(N^2) problem.

elric1:
Thanks! I've rebased it and incorporated your changes.

PR #1416, opened 2026-08-08
krb5: preserve ENC-CHAL state after PKINIT

elric1:


lib/krb5/init_creds_pw.c
```diff
@@ -1770,27 +1771,46 @@ enc_chal_step(krb5_context context,
...
     if (ppaid == NULL) {
-	_krb5_debug(context, 5, "no ppaid found");
-	return HEIM_ERR_PA_CONTINUE_NEEDED;
+	if (!rep) {
+	    _krb5_debug(context, 5, "no ppaid found");
+	    return HEIM_ERR_PA_CONTINUE_NEEDED;
+	}
+	if (ctx->paid.etype == KRB5_ENCTYPE_NULL) {
+	    _krb5_debug(context, 0,
+			"no paid when building ENC_CHAL reply key");
```
nicowilliams, line 1788: "paid" is not going to help anyone who isn't a Heimdal dev or willing to look in the sources.  "preauthentication info data" might be better.

nicowilliams: It's really etype, salt, and s2k params.

lib/krb5/init_creds_pw.c
```diff
@@ -1770,27 +1771,46 @@ enc_chal_step(krb5_context context,
     else
 	paid.etype = rep->enc_part.etype;
     ppaid = process_pa_info(context, ctx->cred.client, a, &paid, in_md);
+    local_paid = (ppaid == &paid);
```
nicowilliams, line 1774: `ppaid` will always be either `NULL` or `&paid`.

nicowilliams: ~So really, `have_paid` might be a better name for this boolean variable.~

nicowilliams:
If you want you can fix up the debug message to be more useful, but this is great.

PR #1417, opened 2026-08-08
iprop: add one-shot slave mode

elric1:


nicowilliams:
Ah, what a nice idea.  This would simplify `tests/kdc/check-iprop.in` a great deal.

PR #1418, opened 2026-08-21
Udp tcp fallback

elric1:
We don't fall back from UDP to TCP unless the KDC tells us to. This only hits us if we have unadorned KDC entries in `/etc/krb5.conf`, so in this case we add tcp/hostname if we are doing UDP and hostname doesn't have a slash in it.
We also recreate the situation in some tests.

### DIFF
--- a/lib/krb5/krbhst.c
+++ b/lib/krb5/krbhst.c
@@ -373,6 +373,23 @@ append_host_string(krb5_context context, struct krb5_krbhst_data *kd,
 	return krb5_enomem(context);
 
     append_host_hostinfo(kd, hi);
+
+    /*
+     * If we are a bare host, and are using UDP, add TCP as well so that
+     * we still try TCP in the absence of KRB_ERR_RESPONSE_TOO_BIG.
+     */
+
+    if (strchr(host, '/') != NULL)
+        return 0;
+
+    if (krbhst_get_default_proto(context, kd) != KRB5_KRBHST_UDP)
+        return 0;
+
+    hi = parse_hostspec(context, kd, host, def_port, port);
+    if (hi == NULL)
+        return krb5_enomem(context);
+    hi->proto = KRB5_KRBHST_TCP;
+    append_host_hostinfo(kd, hi);
     return 0;
 }
 
--- a/lib/krb5/Makefile.am
+++ b/lib/krb5/Makefile.am
@@ -27,6 +27,7 @@ noinst_LTLIBRARIES =				\
 TESTS =						\
 	aes-test				\
 	derived-key-test			\
+	krbhst-test				\
 	n-fold-test				\
 	parse-name-test				\
 	pseudo-random-test			\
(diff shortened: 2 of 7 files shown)


nicowilliams:
> We don't fall back from UDP to TCP unless the KDC tells us to.

Maybe we should just have an option to try TCP if UDP is port-unreachable or even try TCP first.

elric1:
I'm solving a bug here which is that if you specify a KDC in `/etc/krb5.conf`, current Heimdal will _not_ try TCP unless it is told by the KDC that the response is too big. In this case, it doesn't fall back to TCP, it just fails.
The smallest fix is just to interpret `kdc = hostname` as `kdc = udp/hostname` and `kdc = tcp/hostname` which is the documented behaviour: that config implies that the KDC is serving both. Its order is different from MIT's because it's `udp/kdc1`, `tcp/kdc1`, `udp/kdc2`, ..., so the second try will be TCP.
You can force TCP first by using `kdc = tcp/kdc.realm` in `/etc/krb5.conf`. You can also disable UDP with `max_message_size=1`.

nicowilliams:
Does MIT also support that?  It's not in their `krb5.conf` manpage.

elric1:
No, they don't support `tcp/` and `udp/`. But nothing we do to provide this kind of configurability will be compatible with MIT. I'm mainly interesting in succeeding in every case that we should succeed with the KDCs specified in `/etc/krb5.conf` rather than providing ultimate configurability as, I think, that in prod people should usually use DNS SRV RRs.
There's an interesting question, though: should TCP and UDP SRV RRs be sorted together with weights and priorities? Not sure if we do that or not or what would be most intuitive there.
But for `/etc/krb5.conf` specified KDCs, I don't think that we should overthink it.

nicowilliams:
```                                 kdc = [service/]host[:port]
                                      Specifies a list of kdcs for this realm.
                                      If the optional port is absent, the de‐
                                      fault value for the “kerberos/udp”
                                      “kerberos/tcp”, and “http/tcp” port (de‐
                                      pending on service) will be used.  The
                                      kdcs will be used in the order that they
                                      are specified.

                                      The optional service specifies over what
                                      medium the kdc should be contacted.
                                      Possible services are “udp”, “tcp”, and
                                      “http”.  Http can also be written as
                                      “http://”.  Default service is “udp” and
                                      “tcp”.
```

That's from Heimdal's own `lib/krb5/krb5.conf.5`.

nicowilliams:
Oh, you're not adding this functionality; you're using it.

lib/krb5/krbhst.c
```diff
@@ -373,6 +373,23 @@ append_host_string(krb5_context context, struct krb5_krbhst_data *kd,
...
 
     append_host_hostinfo(kd, hi);
+
+    /*
+     * If we are a bare host, and are using UDP, add TCP as well so that
+     * we still try TCP in the absence of KRB_ERR_RESPONSE_TOO_BIG.
+     */
+
+    if (strchr(host, '/') != NULL)
+	return 0;
```
nicowilliams: Please no tabs :)

elric1:
Tabs removed.

PR #1419, opened 2026-08-24
krb5: interoperate with signed KDC nonces

elric1:
MIT and older Apple implementations encode Kerberos nonces as signed integers.  Restore the signed ASN.1 nonce types and constrain generated KDC nonces to the non-negative signed range.

### DIFF
--- a/lib/asn1/krb5.asn1
+++ b/lib/asn1/krb5.asn1
@@ -572,7 +572,7 @@ KDC-REQ-BODY ::= SEQUENCE {
 	from[4]			KerberosTime OPTIONAL,
 	till[5]			KerberosTime OPTIONAL,
 	rtime[6]		KerberosTime OPTIONAL,
-	nonce[7]		Krb5UInt32,
+	nonce[7]		Krb5Int32,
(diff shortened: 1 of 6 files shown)


vdukhovni:
This appears to be the root cause of `kinit -R` on MacOS not interoperating with the KDC.

nicowilliams:
Weird.  I changed them to unsigned because RFC 4556 (EDIT: also RFC 4120) says it should be unsigned 32-bit.  Maybe they should be treated as signed 64-bit and then checked to be signed 33 bit?

elric1:
>    Weird. I changed them to unsigned because RFC 4556 says it should be
>    unsigned 32-bit. Maybe they should be treated as signed 64-bit and then
>    checked to be signed 33 bit?

I know, I saw that in the RFC.  But it turns out that the ecodings are
quite different because they are variable length.  So, -1 encoded as a
signed integer will decode as 0xff as unsigned and 0xffffffff as signed.

What I'm doing here is the prior art from both MIT and Windows: they
also mask off the top bit which means that all numbers are positive and
the encoding is the same.

--
    Roland C. Dowdeswell                          https://Imrryr.ORG/

PR #1420, opened 2026-08-27
Eliminate krb5_principal_data from API

elric1:


kdc/pkinit.c
```diff
@@ -1967,12 +1966,17 @@ match_rfc_san(krb5_context context,
 	    return KRB5_KDC_ERR_CLIENT_NAME_MISMATCH;
 	}
 
-	memset(&principal, 0, sizeof (principal));
-	principal.name = kn.principalName;
-	principal.realm = kn.realm;
-
-	if (krb5_principal_compare(context, &principal, match) == TRUE)
-	    found = 1;
+	if (strcmp(kn.realm, krb5_principal_get_realm(context, match)) == 0 &&
```
nicowilliams: I'd rather construct a `krb5_principal` and compare it as before because we could eventually do something like adopt AD's realm case-insensitivity, in which case we'd have to find and change this code again, but if we continue to compare with `krb5_principal_compare()` then we wouldn't have to.

elric1: Agreed, I've pushed a new version.

nicowilliams:
LGTM.  Merge when checks pass.

nicowilliams:
```Creating database
Check parser
Doing database check
Extracting enctypes
deleting all but sha1 enctypes on kt-aes in keytab
checking globbing keys rules
Testing quick TCP fallback after UDP port unreachable
timed-exec: command took 1000ms; expected less than 250ms
FAIL check-kdc-weak (exit status: 1)
``` _sigh_

elric1:
> ```
> Check parser
> Doing database check
> Extracting enctypes
> deleting all but sha1 enctypes on kt-aes in keytab
> checking globbing keys rules
> Testing quick TCP fallback after UDP port unreachable
> timed-exec: command took 1000ms; expected less than 250ms
> FAIL check-kdc-weak (exit status: 1)
> ``` _sigh_
> ```

This is me. I am trying to write a test that demonstrates that UDP fallback is instant if you get an `ICMP_PORT_UNREACH`, but unfortunately, even on localhost this appears to be a bit flakey. 1000ms is exactly what you'd expect if the port unreach is somehow dropped.

PR #1423, opened 2026-09-16
hx509: accept id-ecPublicKey as a CMS signature algorithm (verify-only)

### LINKED ISSUE #1421: PKINIT: sig_algs[] has no entry for id-ecPublicKey, which is what Windows sends in SignerInfo.signatureAlgorithm
Hi — thanks for Heimdal. This is an interop issue between Heimdal's PKINIT verifier and
Windows CNG that makes **Windows smart-card logon with an EC client certificate**
not work against a Heimdal KDC. Found via Samba's embedded Heimdal; `sig_algs[]` on
master looks to have the same gap, though I have not tested master directly — see the
verification note below.

Happy to send a PR; wanted to check the preferred approach first, since our lab fix
takes a shortcut described below.

## What Windows actually sends

`tshark` dissection of a real Windows 11 AS-REQ with an **ECDSA P-384** logon
certificate:

```
SignerInfo.signatureAlgorithm = id-ecPublicKey  (1.2.840.10045.2.1)
SignerInfo.digestAlgorithm    = sha384          (2.16.840.1.101.3.4.2.2)
```

Windows/CNG places the **key** algorithm OID in `SignerInfo.signatureAlgorithm`, where
RFC 5652 §5.3 expects the **signature** algorithm OID (i.e. `ecdsa-with-SHA384`,
`1.2.840.10045.4.3.3`).

Whether that's strictly conformant is arguable, but it's what every Windows client emits,
so a KDC that wants Windows smart-card logon has to accept it.

## Why it fails

`lib/hx509/crypto.c`'s `sig_algs[]` contains only
`ecdsa_with_sha{1,256,384,512}_alg`. There's no entry keyed on `id-ecPublicKey`, so
`_hx509_find_sig_alg()` returns NULL and verification stops with
`HX509_SIG_ALG_NO_SUPPORTED`:

```
PKINIT: failed to verify signature: Failed to verify signature in CMS SignedData
```

or, depending on build:

```
PKINIT: Signature algorithm not supported
```

This is independent of whether ECDSA support is compiled in — with the ECDSA algorithms
present and available, they still never match, because the lookup key never appears in
the table.

## Suggested fix

Accept `id-ecPublicKey` as a signature algorithm and take the digest from the
`SignerInfo.digestAlgorithm` field, which Windows populates correctly.

Our lab patch adds a `sig_algs[]` entry keyed on `id-ecPublicKey` that dispatches to
`ecdsa_verify_signature` — but with the digest **hardcoded to SHA-384**, which is only
correct for P-384 clients. That was enough to prove the diagnosis; a correct fix needs to
be digest-agnostic, and I'd rather follow your preference on where that logic belongs
than guess. If a `sig_algs[]` entry that consults `digestAlgorithm` is acceptable, I'm
glad to write it up properly.

## Impact

This alone makes Windows EC smart-card logon against Heimdal impossible, which is
probably why it appears never to have worked for anyone. RSA client certificates are
unaffected.

## Checked against master before filing

`lib/hx509/crypto.c` on **master** still has no `id-ecPublicKey` entry — `sig_algs[]`
lists `ed25519_alg`, `ed448_alg`, `ecdsa_with_sha{512,384,256,1}_alg`, the RSA/DSA
entries and the bare digests. In fact `crypto.c` on master does not contain the string
`ecPublicKey` at all (the bundled tree we tested at least had `_hx509_signature_ecPublicKey`
as a key OID), so the OpenSSL 3.x rework appears to have removed the only place it
appeared rather than adding handling for it.

The verification path that consumes it also looks unchanged. Reading master:

```c
/* lib/hx509/cms.c — hx509_cms_verify_signed_ext() */
ret = _hx509_verify_signature(context, cert,
                              &signer_info->signatureAlgorithm,   /* <- id-ecPublicKey */
                              &signer_info->digestAlgorithm,
                              &signed_data, &signer_info->signature);
```

```c
/* lib/hx509/crypto.c — _hx509_verify_signature() */
sa = _hx509_find_sig_alg(&alg->algorithm);      /* alg == signatureAlgorithm */
if (sa == NULL) {
    hx509_clear_error_string(context);
    return HX509_SIG_ALG_NO_SUPPORTED;
}
```

So a Windows AS-REQ whose `SignerInfo.signatureAlgorithm` is `id-ecPublicKey` should still
hit `HX509_SIG_ALG_NO_SUPPORTED` on master, for the same reason it does on the tree we
tested.

**What I have NOT done, and where I could be wrong:** I did not build or test master, only
read it. #1289 changed a great deal here, and if RFC 8636 algorithm agility causes Windows
to construct the CMS differently in some negotiated mode, or if some caller normalises
`signatureAlgorithm` before the code above, then this may already be handled and I have
missed it. I would rather be corrected than have you spend time on a non-issue — say the
word and I will build master and retest against a real Windows 11 client.

No open issue appears to cover this: searched `id-ecPublicKey`, `sig_algs`,
`pkinit ecdsa`. The closest are #1038 (hx509 ECDSA support) and #1289, both merged, and
neither adds a signature algorithm keyed on the *key* OID.

## Environment

- Found on Heimdal as bundled in Samba **4.22.10** (Debian 13); the same `sig_algs[]`
  table is present in Heimdal master
- Client: Windows 11 Pro (10.0.26200), PIV card, ECDSA **P-384** logon certificate
- Also reproduces with Samba 4.17.12 (Debian 12) — longstanding, not a regression

Related: `kdc/pkinit-ec.c` hardcodes secp256r1, which blocks the same flow one step
later — filed separately, since they're independent fixes.

Also reported downstream at Samba, where this ships in the embedded Heimdal:
**https://bugzilla.samba.org/show_bug.cgi?id=16231** (that bug covers all three
Samba-visible causes, including a Samba-side build issue that hides this one).

With both addressed, Windows 11 smart-card logon on a P-384 credential works online and
offline. Glad to provide the pcap, full KDC logs, or test patches.

## Our lab patch, for reference

Applies to Heimdal as bundled in `samba-4.22.10+dfsg`. **Not offered as a merge
candidate** — the digest is hardcoded, as noted above.

```diff
--- a/third_party/heimdal/lib/hx509/crypto-ec.c	2025-01-30 13:57:45.000000000 +0000
+++ b/third_party/heimdal/lib/hx509/crypto-ec.c	2026-09-01 23:21:32.000000000 +0000
@@ -785,6 +785,30 @@
     48
 };
 
+/*
+ * Windows/CNG places the KEY algorithm OID (id-ecPublicKey) in the CMS
+ * SignerInfo.signatureAlgorithm field, where RFC 5652 expects the SIGNATURE
+ * algorithm OID (ecdsa-with-SHA384). Accept it and verify as ECDSA/SHA-384 so
+ * that Windows PKINIT smart-card logon interoperates.
+ *
+ * NOTE: the digest is fixed at SHA-384 here, which is only correct for P-384
+ * clients. A general fix should derive the digest from SignerInfo.digestAlgorithm.
+ */
+const struct signature_alg ecdsa_ecpubkey_sha384_alg = {
+    "ecdsa-ecpublickey-sha384",
+    ASN1_OID_ID_ECPUBLICKEY,
+    &_hx509_signature_ecPublicKey,
+    ASN1_OID_ID_ECPUBLICKEY,
+    &_hx509_signature_sha384_data,
+    PROVIDE_CONF|REQUIRE_SIGNER|RA_RSA_USES_DIGEST_INFO|
+        SIG_PUBLIC_SIG|SELF_SIGNED_OK,
+    0,
+    NULL,
+    ecdsa_verify_signature,
+    ecdsa_create_signature,
+    48
+};
+
 const struct signature_alg ecdsa_with_sha256_alg = {
     "ecdsa-with-sha256",
     ASN1_OID_ID_ECDSA_WITH_SHA256,
--- a/third_party/heimdal/lib/hx509/crypto.c	2025-01-30 13:57:45.000000000 +0000
+++ b/third_party/heimdal/lib/hx509/crypto.c	2026-09-01 23:21:32.000000000 +0000
@@ -760,6 +760,7 @@
 #ifdef HAVE_HCRYPTO_W_OPENSSL
 extern const struct signature_alg ecdsa_with_sha512_alg;
 extern const struct signature_alg ecdsa_with_sha384_alg;
+extern const struct signature_alg ecdsa_ecpubkey_sha384_alg;
 extern const struct signature_alg ecdsa_with_sha256_alg;
 extern const struct signature_alg ecdsa_with_sha1_alg;
 #endif
@@ -968,6 +969,7 @@
 static const struct signature_alg *sig_algs[] = {
 #ifdef HAVE_HCRYPTO_W_OPENSSL
     &ecdsa_with_sha512_alg,
+    &ecdsa_ecpubkey_sha384_alg,
     &ecdsa_with_sha384_alg,
     &ecdsa_with_sha256_alg,
     &ecdsa_with_sha1_alg,
```

## Confirmed end state

With this and the `pkinit-ec.c` curve issue patched (plus a Samba-side build fix so the
ECDSA code is compiled at all), Windows 11 smart-card logon on an **ECDSA P-384**
credential works against a Samba AD DC **both online and offline** (cached credentials,
KDC unreachable):

```
Kerberos: PKINIT pre-authentication succeeded -- alice@example.com using CN=alice
Kerberos: PKINIT using ecdh
AS-REQ SUCCESS  pkinit_client_cert=CN=alice  pa=PK-INIT(ietf)  etype=18/18
```

As far as I can tell that combination has not worked before, which would explain why these
gaps went unnoticed. Happy to supply the AS-REQ pcap, KDC logs at level 10, or to rework
the patch along whatever lines you prefer.

spurglegooge:
Fixes #1421, as discussed there.

Windows CNG fills CMS `SignerInfo.signatureAlgorithm` with the **key** algorithm
(`id-ecPublicKey`) instead of an `ecdsa-with-SHAx` OID when signing with an EC smart-card
key. Observed in PKINIT AS-REQs from Windows 10 and 11, where it makes EC smart-card logon
fail against a Heimdal KDC with `HX509_SIG_ALG_NO_SUPPORTED` before the key type is ever
considered (capture in #1421).

That Windows behaviour is **non-conformant**: RFC 5753 §2.1.1 (via RFC 5754 §3.3) requires
`signatureAlgorithm` to name one of the `ecdsa-with-SHAx` OIDs. For RSA there is a genuine
normative license to use the bare key OID — RFC 3370 §3.2 makes `rsaEncryption` a valid
signature-value identifier "regardless of the message digest algorithm employed", which is
what the existing `rsa-pkcs1-x509` entry implements — but ECDSA has no analogue. So this is
a deliberate interop concession, tolerated on **receive only**: an `ecdsa-x509` entry keyed
on `id-ecPublicKey` with no digest of its own, taking the digest from
`SignerInfo.digestAlgorithm` as delivered by `_hx509_verify_signature()`.

The absence of any normative license shapes the hardening:

- **The entry is strictly verify-only.** `create_signature` is NULL;
  `_hx509_create_signature()` now returns `HX509_ALG_NOT_SUPP` ("signature algorithm is
  verify-only") for such entries instead of calling the pointer unconditionally; and
  `hx509_crypto_available()` and `hx509_crypto_select()` skip verify-only entries — keyed
  off the entry's own `SIG_PUBLIC_SIG` flag rather than the caller's requested selection, so
  that a `HX509_SELECT_ALL` query does not also drop digest-only entries — and
  `alg_for_privatekey()` skips any entry that cannot sign. The OID is therefore neither
  advertised in `supportedCMSTypes` nor selectable from a peer's list for signing.
  Without those guards, a KDC holding an EC key could be crashed by any client listing
  `id-ecPublicKey` first in `supportedCMSTypes` — and since the premise of this patch is
  that Windows treats `id-ecPublicKey` as a signature algorithm, that is a likely interop
  crash, not just an attack. The `_hx509_create_signature()` guard also closes a
  **pre-existing** NULL call: `dsa_sha1_alg` has `SIG_PUBLIC_SIG` with
  `create_signature == NULL`, and the peer-selection loop matches on `sig_oid` alone, so a
  peer listing `id-dsa-with-sha1` against a DSA key reached `(*sa->create_signature)()`
  with a NULL pointer before this change.
- **Digests weaker than SHA-256 are refused on the inferred path.** It exists for one
  broken producer and must not make MD5/SHA-1+ECDSA newly verifiable. (This also excludes
  SHA-224, which RFC 5753 permits for conformant ECDSA — but a conformant producer never
  takes this path, and the observed producer uses SHA-256/384.)
- **Digest precedence is unchanged for existing entries.** The entry's own `digest_alg`
  stays authoritative when it names one (in CMS the content digest may legitimately differ
  from the signature's digest); the caller-supplied digest is used only when the entry
  names none, and both `ecdsa_verify_signature()` and `ecdsa_create_signature()` now fail
  cleanly with `HX509_ALG_NOT_SUPP` instead of dereferencing NULL when neither exists.
- **Known side effect:** `find_keytype()` resolves a private key's `signature_alg` via
  `_hx509_find_sig_alg()`. EC private keys loaded from PEM carry `id-ecPublicKey` there and
  previously resolved to NULL, disabling the keytype filter in select/available; they now
  resolve and the filter applies. A latent-bug fix, but it changes negotiation for keys
  unrelated to the Windows case, so it is called out here.
- Certificate and CRL/OCSP verification paths pass `digest_alg = NULL` and are unaffected:
  they get `HX509_ALG_NOT_SUPP` rather than newly accepting `id-ecPublicKey`-"signed"
  certificates.

**Testing, stated precisely.** The inference itself — accept `id-ecPublicKey`, take the
digest from `SignerInfo.digestAlgorithm` — is what we run in production (in equivalent form,
against Samba's bundled Heimdal): it is the change that makes Windows 10/11 EC smart-card
PKINIT succeed against a Samba AD KDC. The verify-only enforcement, digest floor, and
negotiation guards are new in this patch and are not exercised by the existing suite —
`make -C lib/hx509 check` on Debian trixie is identical before and after (11 pass / 2 skip /
4 pre-existing environment failures, same on clean master in the same container). I am
preparing a checked-in regression vector — a stored SignedData with
`signatureAlgorithm = id-ecPublicKey` verified via `hxtool cms-verify-sd` in `test_cms`,
validated to verify on this branch and fail on master — and will push it to this PR as a
follow-up commit; it has to be a stored blob, since after this patch Heimdal cannot generate
one. Happy to adjust naming or placement of any of it.


spurglegooge:
Added the regression test as a follow-up commit (`659955f`): a stored `SignedData` whose
`SignerInfo.signatureAlgorithm` is `id-ecPublicKey`, verified in `test_cms` against the
in-tree `secp256r1` test CA with the content round-tripped.

It has to be a checked-in blob, since with this fix the `ecdsa-x509` entry is verify-only and
Heimdal won't generate one. I made it by taking a normal `hxtool cms-create-sd` output and
rewriting only the `SignerInfo.signatureAlgorithm` OID (ecdsa-with-SHA384 → id-ecPublicKey);
confirmed it's **accepted on this branch and rejected by master** (`HX509_SIG_ALG_NO_SUPPORTED`),
so the case exercises the new path rather than passing trivially. `make -C lib/hx509 check`
is otherwise unchanged.

Happy to move or rename the vector if you'd prefer it elsewhere.

nicowilliams:
@spurglegooge do we need something like this when Heimdal is a PKINIT _client_ of an AD KDC?

nicowilliams:
This LGTM.  I may rewrite the commit messages and some comments for brevity/pithiness.

spurglegooge:
Yes — good call -- upon review, it looks like the patch covers it.

The client verifies the KDC's signed PKINIT reply through the same path the patch touches:
`pk_verify_sign()` (`lib/krb5/pkinit.c:1150`) → `hx509_cms_verify_signed_ext()` (`cms.c:851`)
→ the per-`SignerInfo` loop → `_hx509_verify_signature()`. So accepting `id-ecPublicKey` in
`SignerInfo.signatureAlgorithm` on verify applies whether Heimdal is the KDC verifying the
client's AuthPack or the client verifying the KDC's reply.

Whether a client actually hits it depends on the AD KDC's own signing key: the client sees
`id-ecPublicKey` only if the KDC signs its reply with an EC key and CNG mislabels the KDC's
`SignerInfo` the way it does on the client-signing side. AD KDC certificates are frequently
RSA, in which case the client never encounters it — but where the KDC cert is EC, a stock
Heimdal client would fail verifying the reply with `HX509_SIG_ALG_NO_SUPPORTED`, and this fix
resolves it.

No create-side change is wanted for the client direction: when Heimdal signs its own AuthPack
(smart card via PKCS#11) it emits conformant `ecdsa-with-SHAx`, and the entry is deliberately
verify-only so Heimdal never produces the malformation itself.

tbh : I've only reproduced the receive side directly (Windows client → Samba/Heimdal
KDC, plus the standalone CMS repro in this PR), but I have not captured an AD KDC emitting
`id-ecPublicKey` in its reply — that leg is inferred from the shared code path and CNG's
behavior. If you have an AD deployment with an EC KDC certificate I'd be glad to confirm it.

spurglegooge:
> This LGTM. I may rewrite the commit messages and some comments for brevity/pithiness.

Good call ... thanks for cleaning that up

nicowilliams:
Thanks!

